PR #13, opened 2015-03-26
[script.module.subtitles.helper] - v 0.0.1

amet:
initial commit of most commonly used subtitle functions, helps to keep it all in one module and everyone can reuse, more can be added if there is interest.


amet:
hm, I changed my mind... I'll keep it in individual addons

PR #23, opened 2015-12-01
[script.xbmc.unpausejumpback] 2.3.4

Memphiz:
as requested by @razze


MartijnKaijser:
@Memphiz maybe also change the the name to "Unpause Jumpback" (drop the xbmc part) ?  Should give no issues as long as you keep addon id the same as it is now. Up to you of course :)

Memphiz:
@MartijnKaijser good call - done

PR #29, opened 2015-12-21
[script.xbmc.boblight] 2.0.14

Memphiz:
add support for ne API


MartijnKaijser:
Maybe for next time "Kodi Bobliight" as description ;)

PR #30, opened 2015-12-21
[script.lifx] 1.0.0 Lifx smartbulb

mhtsbt:
This addon will change the color of your Lifx-bulbs to match the currently playing video.

You can find more information about this addon on my blog: http://www.blinkingled.be/lifx-plugin-for-kodi/


MartijnKaijser:
please make sure you only send the script as one commit. So squash all changes into one

PR #32, opened 2015-12-22
[script.xbmc.boblight] 2.0.15

Memphiz:
This time with rebrand :) - This also adds tvos support.


Memphiz:
@MartijnKaijser i realised that you seem to work on the addons backend - are we sure that scripts in the gotham branch are updated in jarvis accordingly? Just asking because the former boblight bump to 2.0.14 is not visible yet (and also wasn't mentioned in the digest mails yet).

PR #38, opened 2016-01-26
[Xbmc-addons] [Git Pull] script.featherence.service

finalmakerr:
First time trying :)


MartijnKaijser:
It will never be accepted in current state. It contains a lot of reference to stuff which is banned. Start cleaning it up and reading the repo and forum rules

MartijnKaijser:
Too add there absolutely no reason to include keymaps, samba configs, sources.xml files and blatend copies of already existing addons and whatever more.

finalmakerr:
Will work it out np.
Thanks and sorry for your troubles.

finalmakerr:
Martijn,
I removed the entire special folder and the relative modules and settings.xml
Now i am removing the reference for some addons.
Please if you could be so kind to direct me to some list of banned addons just to be sure.

P.S. I have readed the forum rules , will do so again soon.

PR #40, opened 2016-02-02
[service.library.data.provider] 0.1.5 - Fix circular reference warnings on shutdown

anaconda:
Same as #35


MartijnKaijser:
https://github.com/BigNoid/service.library.data.provider for fixes

PR #42, opened 2016-02-13
Fixed plugin loading on Windows

kkroflin:
Fixed: Plugin fails to load on Windows when the username contains UTF-8 characters (e.g. "š")


MartijnKaijser:
pushed in with other updated. thx

PR #55, opened 2016-03-06
[helix] script.extendedinfo 3.1.4

phil65:
Remove autocompletion for older Kodi versions (since it is not possible to detect whether HIDDEN_INPUT is set.)


phil65:
merged as changes are trivial.

PR #56, opened 2016-03-07
[jarvis] - script.maps.browser 1.0.0 + ExtendedInfo Update

phil65:


ronie:
if Maps Browser is supposed to be a script as well as a python module,
you should split them into separate addons.

for python modules (though i'm not an expert here) i think all *.py files need to be in a subfolder in /lib.

also the name of that folder needs to be unique so it doesn't clash with names of other python modules.

so /lib/googlemaps/*.py should work.

for example, have a look at: https://github.com/xbmc/repo-scripts/tree/gotham/script.module.chardet

phil65:
Sorry, that extension point was an old leftover and not used (yet). I removed it now, thx.

script.maps.browser/resources/language/English/strings.po
```diff
@@ -0,0 +1,651 @@
...
+"Plural-Forms: nplurals=2; plural=(n != 1)\n"
+
+msgctxt "Addon Summary"
+msgid "Google Maps Browser (including Street View)"
+msgstr ""
+
+msgctxt "Addon Description"
+msgid ""
+"This Maps Browser Add-on allows you to explore the map with a remote. It can also show Places (Source: FourSquare), "
+"events (source: Eventful), geotagged images and much more on the map. Also offers skin integration."
```
ronie: not sure if it's valid like this, but perhaps stick with using a single line:
see: https://github.com/xbmc/xbmc/commit/5e7507e663461e023aec4a958a6d6dadaea493b9

ronie:
as mention in the inline comments, no idea if this will work:
script.maps.browser/resources/skins/Default/color/defaults.xml

these shouldn't be need afaik:
script.maps.browser/__init__.py
script.maps.browser/resources/__init__.py

icon.png needs to be 256x256px

phil65:
the __ init __ in resources is needed i think because i dont use sys.path.append(), but import from the subfolders directly. The other one is superfluous, I removed it.
The multiline string in .po file is valid, I might ask alan why his system translates those to a single-line string.
The rest of your comments was addressed and fixed.

phil65:
I think we should also allow up to 512x512 for add-ons nowadays. Agreed?

kib:
Makes sense to me

ronie:
not my call, i received some slack when i (not that long ago) told people it was ok to use a 1920x1080 fanart.jpg instead of 1280x720.

the only downside i can think of is it's going to take ~4 times as long to fetch all the thumbs the first time you enter the addon browser.
if skins use a 3x5 panel layout, kodi will need to download 15 512x512px images at once.

phil65:
isnt that the same (or even worse) with movie posters?
I am also not sure if png filesize correlates with pixel amount that much. ;)

phil65:
good to go now?

PR #61, opened 2016-03-19
[script.pinsentry] 1.3.0

robwebset:
Changelog:

v1.3.0 (2016-03-19)
- Add French classifications
- Add screensaver option that triggers a prompt for the pin
- Allow the pin to be prompted for on startup

For Releases Helix and beyond


tamland:
This goes for all you PRs: using 'provides' in all extension points is not valid. If you are unsure, please verify your xmls agains the xsd files found in https://github.com/xbmc/xbmc/tree/master/addons/xbmc.python

robwebset:
I'm not exactly sure what you mean, these sort of addons have been in the repo for a while (via the mailing list).  Looking at the schema's you linked to it seems only the pluginsource can have a 'provides' which does not match what the wiki says:

http://kodi.wiki/view/Addon.xml

As it has it for scrips as well. (Look at the example at the bottom).

I don't think this inconsistency should stop these updates going in.

When both sources match - then I'll align to them and include it in the next pull request.

Thanks, Rob

tamland:
The schemas are the reference documentation. It is right (unless there's a bug). Consider it a friendly warning that it isn't valid and supported.

robwebset:
FYI: Updated all my addons so that they only have "provides" for the pluginsource - so should be picked up when the next pull request happens for each addon.

Tanks, Rob

PR #62, opened 2016-03-19
[script.ebooks] 1.2.0

robwebset:
Changelog:

v1.2.0 (2016-03-12)
- Protect against character encoding issues in PDF files
- Support PDF over OPDS interface
- List more than 30 books when using OPDS
- OPDS support for preference of multiple book formats

For Releases Gotham and beyond


robwebset:
Just wondering if this one had accidentally slipped under the radar? - The other addons updated, PinSentry and Audiobooks that I submitted at the same time have both been done, but not this one.

Thanks, Rob

script.ebooks/resources/lib/kindleunpack/mobi_header.py
```diff
@@ -135,7 +169,7 @@ def dump_contexth(cpage, extheader):
         content = extheader[pos + 8: pos + size]
         if id in id_map_strings:
             name = id_map_strings[id]
-            print('\n    Key: "%s"\n        Value: "%s"' % (name, content.decode(codec)))
```
learningit: Please replace print statements with log()

script.ebooks/plugin.py
```diff
@@ -1,568 +1,600 @@
...
+import os
+import urllib
+import urlparse
+import xbmc
+import xbmcgui
+import xbmcplugin
+import xbmcaddon
+import xbmcvfs
+
+__addon__ = xbmcaddon.Addon(id='script.ebooks')
```
learningit: We are trying not use the label format _\_name_\_ it is used for Python internals and is confusing.

script.ebooks/resources/lib/kindleunpack/mobi_header.py
robwebset: This embedded lib is a free tool used for more than just Kodi, so it would be hard to change all the prints to logs as it would make updating to resolve bug fixes very hard.

script.ebooks/plugin.py
robwebset: All my addons do this sort of thing, I think it was in the original TvTunes, so thought it was the way things were done.  Can you point me at the guidelines for this behavior please? I'll take a look at updating all my addons at the same time if that is OK.

phil65: Not really a kodi-specific thing, but a general recommendation.

https://www.python.org/dev/peps/pep-0008/

> __double_leading_and_trailing_underscore__ : "magic" objects or attributes that live in user-controlled namespaces. E.g. __init__ , __import__ or __file__ . Never invent such names; only use them as documented.

robwebset:
Hi All,

As the items raised are on existing things already in the current repo version, please can we get this update in, and I'll take a look at them for the next pull.

Thanks
Rob

script.ebooks/plugin.py
robwebset: Had a little time, so updated to remove the use of __name__

phil65:
looks fine to me. @learningit 	for last check and green button.

PR #64, opened 2016-03-19
[script.sleep] 1.0.0

robwebset:
A new addon to include please, details on the forum.

http://forum.kodi.tv/showthread.php?tid=261959

For releases Isengard onwards


script.sleep/addon.xml
```diff
@@ -0,0 +1,23 @@
...
+<addon id="script.sleep" name="Sleep" version="1.0.0" provider-name="robwebset">
+	<requires>
+		<import addon="xbmc.python" version="2.20.0"/>
+		<import addon="script.module.simplejson" version="3.3.0"/>
+	</requires>
+	<extension point="xbmc.python.script" library="default.py">
+		<provides>executable</provides>
+	</extension>
+	<extension point="xbmc.service" library="service.py" start="login">
+		<provides>executable</provides>
```
phil65: does the service extension point need the "provides" tag?

script.sleep/default.py
```diff
@@ -0,0 +1,29 @@
...
+import xbmcaddon
+
+
+__addon__ = xbmcaddon.Addon(id='script.sleep')
+__cwd__ = __addon__.getAddonInfo('path').decode("utf-8")
+__resource__ = xbmc.translatePath(os.path.join(__cwd__, 'resources').encode("utf-8")).decode("utf-8")
+__lib__ = xbmc.translatePath(os.path.join(__resource__, 'lib').encode("utf-8")).decode("utf-8")
+
+sys.path.append(__resource__)
+sys.path.append(__lib__)
```
phil65: import resources.lib.whatever is generally a nicer way than polluting sys.path.

phil65:
Looks fine. Some of my comments from other PR also apply here.
Thx.

script.sleep/addon.xml
robwebset: There is some confusion between the wiki and the schema with this one, a bit of a discussion from here:

https://github.com/xbmc/repo-scripts/pull/61#issuecomment-198765162

However will actually quickly update it.

script.sleep/default.py
robwebset: The sys path update was something that I "borrowed" from another addon when I wrote my first addon - so thought it was the way it needed to be done.

Please can we get this one merged with this method, then I'll do some investigation to try and improve this on all my addons in one go.

Thanks, Rob

phil65: Sure.

robwebset: Just did a quick test, the import resources.lib.xxx didn't seem to work

robwebset: OK, got it working now - I'd only added init.py to the lib directory, needed it in the resources one as well

PR #65, opened 2016-03-19
[service.addonsync] 1.0.0

robwebset:
Please can we add this new addon.  Details can be found in the following forum post:

http://forum.kodi.tv/showthread.php?tid=260802

For releases Isengard onwards


Razzeee:
Looks fine for me, but not sure what our rules are for addons changing stuff in other addons.

learningit:
This appears to be something that can be used for both good and evil.  It makes evil things really easy by allowing a dropbox to be master.

ronie:
i discussed this addon some time ago with @MartijnKaijser and imo this needs a bit more thought.

there are some occasions where things can go wrong:
- restoring skin settings will not work, so preferably these should be excluded.
- ideally the addon should check whether the versions of each addon matches between master and slave. overwriting a settings file with one from a newer/older version might cause issues.
- service addons would need to be restarted otherwise the new settings won't apply
- paths stored in a settings file may differ between master/slave. (on master a certain path may be local, while on slave it needs to point to a smb:// share on master)

there might be more occasions where things could go wrong.

it's mainly the reason why we dislike all those wizard/maintenance addons so much..
they often cause more harm than do good.

learningit:
I ran into the issue Ronie outlines in point 4 above when using Windows as master and Android as slave.

robwebset:
Hi all, thanks for your comments, will try and answer them as best I can:

> not sure what our rules are for addons changing stuff in other addons

I believe this is OK, as long as you make it clear what is happening - there is a big warning displayed when you install addonsync that I believe will cover this.

> restoring skin settings will not work, so preferably these should be excluded.

Thanks, I have done an update to exclude all skins (I'm not very familiar with skins, so great advice!)

> ideally the addon should check whether the versions of each addon matches between master and slave. overwriting a settings file with one from a newer/older version might cause issues.

Another good idea - I have update the code to check version numbers before copying

> service addons would need to be restarted otherwise the new settings won't apply

I really didn't want to interfere with the running of services - I was looking at AddonSync to be a sort of "eventually consistent" process - I thought this was much safer.

> paths stored in a settings file may differ between master/slave. (on master a certain path may be local, while on slave it needs to point to a smb:// share on master)

Yes, there may be cases where people have set different paths on different systems - in which case, they should ideally be aware of this and would exclude it from AddonSync (if they are using it) - I was sort of viewing this as a generic addon where each other addon could declare if they supported it - a bit like current skins state they support TvTunes/VideoExtras etc.

I hope this clarifies things enough to get this addon into the repo.

Thanks, Rob

KenV99:
FWIW, just in case you decide to go that way, this code restarts an addon by disabling and then reenabling:
https://github.com/KenV99/script.service.kodi.callbacks/blob/master/restartaddon.py

robwebset:
Thanks for the tip @KenV99 I have now added an option in the settings to allow the service addons to be restarted.

I think I've now covered all of the issues raised in the post by @ronie 

I do believe this fits with all the guidelines for the repo - please would it be possible to get this addon added now?

Thanks, Rob

tamland:
You seem to have missed the point. Those were all example failings this addon has, not a list of things to fix that would grant inclusion.

The main problems with this addon is that:
-It relies internals of the addon system, which can break at any moment, even between bug fix releases.
-Is based on one persons understanding of those internals, which is incorrect and incomplete.

That's why this and similar maintenance addons will be buggy at best; cause irreversible damage to userdata at worst.

Adding that broken restart option @KenV99 proposed just underlines the point.

robwebset:
Hi @tamland 

I have not missed the point. I think these are all bordering on "opinions" - I have checked the rules for addons, and I believe this addon fits within the rules (Regardless of some people liking or disliking it - there are several addons I dislike in the repo, doesn't mean they do not fit within the rules).

After all, if developers develop addons that fit with the rules, I think they should be allowed into the repo - after all, they have taken the time to develop the addon and ensure they fit within the rules. I'm happy to listen to suggestions, but believe the "In/Out" choice should always go down to fitting within the rules already set out on the wiki.

Thanks, Rob

learningit:
These are not just "opinions". This add-on clearly breaks the overriding and unspoken rules of "do no harm" and "thou shalt not screw with someone else's add-on". @ronie was very polite in his comments,  there are a number of issues which are not well thought out here. Beyond the multi-platform issue I point out above, this will never work with any add-on which makes use of a file path which includes the user name (particularly windows). This, in the end, will confuse end-users, trash their add-on configs, cause them to complain to other add-on authors and give the general perception that Kodi doesn't work.  When people spend time to read, analyze, consider and comment on real issues, please do the same and do not dismiss their time spent as "opinions".

robwebset:
> These are not just "opinions".

That's your opinion :)

> This add-on clearly breaks the overriding and unspoken rules of "do no harm" and "thou shalt not screw with someone else's add-on".

No harm intended, I wrote this because there was requests for something like it, please read the forum thread for more information:

http://forum.kodi.tv/showthread.php?tid=260802

It was designed to replace the "cron job" that some people are running - I thought that if there was an addon that did this then it can be refined over time to support what is needed.  Again, before starting, I read the addon rules, and believe this does not break them (Obviously I can only take account of "Written Rules" - things in peoples head are hard to get the source material for!)

> @ronie was very polite in his comments, there are a number of issues which are not well thought out here.

Yes, @ronie always seems to be polite and give good feedback, if you look I answered all his questions, and also incorporated his suggestions.

> Beyond the multi-platform issue I point out above, this will never work with any add-on which makes use of a file path which includes the user name (particularly windows). This, in the end, will confuse end-users, trash their add-on configs, cause them to complain to other add-on authors and give the general perception that Kodi doesn't work.

As I highlighted here as the last comment

https://github.com/xbmc/repo-scripts/pull/65#issuecomment-198876732

I see this as being something that Addon Authors highlight as "Supported by their addon" - I'm happy to put a comment to this effect on the Wiki and Forum post.

> When people spend time to read, analyze, consider and comment on real issues, please do the same and do not dismiss their time spent as "opinions".

Did you read this post:

https://github.com/xbmc/repo-scripts/pull/65#issuecomment-198876732

I took the time to try and address all the questions raised - you seem to be selectively reading through the comments?

As I've already shown, I'm happy to accommodate peoples suggestions - but I'm not getting much specific, other that "I don't like it" from a couple of people - if people can suggest specific things then I can get a better understanding. (Interesting these same comments have not been made on the forum post?)

robwebset:
Hi All,

As I've covered all the suggested changes, please can we get this one in? As already highlighted, I believe it fits within the rules and there are no remaining specific points raised that I have not addressed.

Thanks
Rob

robwebset:
I have pushed an update to cover the two items raised:

https://github.com/xbmc/repo-scripts/pull/64#discussion_r56757247

And here:

https://github.com/xbmc/repo-scripts/pull/61#issuecomment-198765162

These are comments raised that applied to all my addons.

It's gone a bit quite on this one, please can we get this in (Noticed several other addons getting merged in)

Thanks, Rob

tamland:
As several people have explained, there is simply far too many failing and things that will go wrong. The end result is that the add-on does not work as advertised. A small reminder that it is not our job to review the code in detail and raise what you call "specific points", but rather ensure that rules, guidelines and certain standards are met.

The rule in question is:

> add-ons should store all their data in their own subfolder inside the addon_data directory. Access (read/write/delete) to any other files / folders is not allowed by default. 
>Exceptions to this rule may be granted in specific cases only. Please contact Team XBMC's add-on repository maintainers via the add-on mailing list if your add-on needs access to such files. 

At this time I have not seen compelling reason given for why such exception should be granted in this case.

robwebset:
Interesting that you quote an updated wiki!!!

I originally submitted via the mailing list on 23rd Feb - when the wiki read:

> add-ons should store all their data in their own subfolder inside the addon_data directory. Access (read/write/delete) to any other files in settings folders is not allowed. Access (read/write/delete) to other all other files/directories must be opt-in by the user, and be clear for the user to understand what is being accessed. Exceptions to this rule may be granted in specific cases only. Please contact Team XBMC's add-on repository maintainers via the add-on mailing list if your add-on needs access to such files.

So it fitted with the rules: **Access (read/write/delete) to other all other files/directories must be opt-in by the user, and be clear for the user to understand what is being accessed.**

Which were deleted on 15th March.  So the Kodi team "Moved the Goalposts AFTER I submitted"

So I did fit with the rules, you're back to the "Just don't like it and won't give explicit details why" - not helpful!!! And you say "A small reminder that it is not our job to review the code in detail and raise what you call specific points, but rather ensure that rules, guidelines and certain standards are met." - so do that, review based on the original rules when it was submitted, in which case it should be merged!

Rob
P.S. Based on previous actons and God Complex issues, I'm guessing you'll soon block comments on this thread!

phil65:
Seems as if you really lost your mind. It's obvious that you hit a grey area which triggered a discussion aboutt what's allowed and what not. @tamland explained very well why such an add-on is not really a good idea, and I fully agree that at least writing to any other addon directory is an awful idea. Dont act as if this would come out of nowhere.

PR #70, opened 2016-03-22
[script.service.kodi.callbacks] 0.9.9

KenV99:
Please disregard mailing list pull request. Downloading from github removed as requested.
ken.vives@gmail.com


KenV99:
BTW wiki: [http://kodi.wiki/view/Add-on:Kodi_Callbacks](url)

script.service.kodi.callbacks/default.py
```diff
@@ -0,0 +1,316 @@
...
+
+
+def createUserTasks():
+    paths = [translatepath('special://addondata')]
+    paths.append(os.path.join(paths[0], 'lib'))
+    paths.append(os.path.join(paths[1], 'usertasks'))
+    for path in paths:
+        if not os.path.isdir(path):
+            try:
+                os.mkdir(path)
```
phil65: we also have methods for creating / checking folders in our xbmcvfs module, those should be preferred.

KenV99: Ok. Is that required? I can do that but then that requires a lot of mocking when running tests outside of Kodi.

tamland: Not required.
Frankly I disagree with @phil65 here. The xbmcvfs module is buggy, so if you do not need to use the VFS but only the local filesystem, it should be avoided.

phil65: Ok,  didn't know they were that buggy.  Ignore this then.

ronie:
please use a _x.y.z_ addon version.

KenV99:
No problem. Version changed. Pull request updated.

script.service.kodi.callbacks/default.py
```diff
@@ -0,0 +1,316 @@
...
+    dispatcher.q_message(PubSub_Threaded.Message(PubSub_Threaded.Topic('onShutdown'), pid=os.getpid()))
+    xbmc.sleep(500)
+    log(msg=_('Shutdown started'))
+    for p in publishers:
+        try:
+            p.abort()
+        except threading.ThreadError as e:
+            log(msg=_('Error aborting: %s - Error: %s') % (str(p), str(e)))
+    dispatcher.abort()
+    xbmc.sleep(1000)
```
tamland: Please try to find a proper solution for what you are trying to solve with this. It will cause issues. When abort has been requested above you should exit _immediately_ with no sleeping/blocking afterwards.

KenV99: Hopefully I can explain the rationale for this so that it is seen that all I am trying to do is catch situations where the threads don't abort. Except in running tests, I have yet to see that this code has ever run on a 'live' Kodi installation in my own or anyone else's logs. So here it goes:

Problem:
In my own logs I see other addons not exit properly and then the dreaded 'xxx has not responded in 5 seconds, let's kill it'. I think, if it were my addon that doesn't respond, I would be embarrassed. I think to myself well you have tested it and it doesn't do that. But then I realize that my testing consists of my own high-ish powered windows machine, two VM's running LinuxMint and OSX 10.11 and a bluestacks Android Emulator. What if it hangs on someone's Rpi running OpenElec? Just let the embarrassment happen? Or try to plan for it?

All of the publishers are running loops in their own separate threads. The dispatcher is in a loop it's own thread. Each of these runs their own code and then sleeps and then checks if an abort has been set in it's own threading.Event(). The amount of time it takes the code portion to run is going to depend on many things so I don't truly know how often each loop is checking for it's own abort event to be set. In addition each of the tasks runs in its own thread and the end user may have written code that takes an unknown amount of time to run. However, I do know that I have five seconds before big brother Kodi shuts me down. So maybe in that five seconds I can catch situations where things have not gone as planned and either fix them or at least write into the log what happened so that I can change the code. So I set up tests where either the publisher, the dispatcher doesn't respond to the abort or if the task is long running.

So on line 137 I publish the onShutdown message and wait an arbitrary 500ms for it to do it's thing and then I send the abort signals to the publishers and the dispatcher (it's really the dispatcher that I don't want to kill before the onShutdown event completes).
On line 145 I give the publishers and dispatcher an arbitrary 1000ms (because of the unknown duration of the loops they are running) before I start looking to see if _any_ of the threads that I own haven't exited (lines 146-159). The sleep in line 145 is there because during _testing_ I found that occasionally the message doesn't make it into the log if the thread aborts successfully in the remaining code. Perhaps the 525ms in thread.abort() is too long, Note that only works if it is one of my own threads that subclassed threading.Thread. I should also catch the exception where the thread does not implement abort. I then log if I was able to terminate the threads. So at least if my addon doesn't exit as planned the log can guide me back to identify where the bad actor is.

Again, I have yet to see this code run on a live install in anyone's logs. Anyway I hope this explains why the code is there.

KenV99: I initially considered having each thread instantiate xbmc.Monitor and use waitForAbort() with a time out instead of xbmc.sleep(). I decided not to do that because:

- I don't really know the time resolution of that function.
- I wanted to avoid the overhead of each of them instantiating that class.
- I wanted to control the order of the shutdown in shutting down the publishers before the dispatcher.
- It doesn't solve the issue with a user executing a long running task of his/her own.

If in the end, this issue prevents letting the addon into the repo, I would be happy to change it and see what happens during testing.

tamland: What I'm trying to point out here is that if you ever need to use sleep to "fix" a problem, you have a fundamental underlying issue will eventually cause shutdown problems. It doesn't matter how much anecdotal evidence you get that it 'works'. By using sleep you actually only making the problem worse for yourself because this time that's potentially wasted _only_ waiting for time to pass. You don't have to use Monitor or waitForAbort to solve this. You can use any event mechanism you want.

This is not just about this one sleep call, but all of them. I don't know any details of this script to be able to give any more specifics, but I assure you there is no problem that sleep "solves" that cannot be solved correctly with events. It's not a blocker, so I think any further discussion on the topic is best taken in the forums. 

Ps: these 5 seconds scripts gets to shut down is _shared_ cpu time. If you are unlucky you will get a few operations executed and you are out of time; which is why it's so important to shut down _fast_, not just within 5 seconds of wall clock time.

KenV99:
tamland: I took your advice to heart and removed all sleep from shutdown. Any needed _potential_ wait time is done via joins with timeouts. Currently passing test on Windows and emulated/VM on OSX, Android and Linux.

KenV99:
I updated my commit to include some added functionality with http verbs and content-type for http tasks and to emit additional info during playback events as was requested by users on the forum.

KenV99:
Hello,
I was wondering if there was anything I could do to assist in finalizing whether or not this addon may be added to the official repo? More code documentation? A few UML diagrams to help understand the core functionality?

I also noted only recently that perhaps I should have submitted this to the Isengard branch since the wiki states that helix is not accepting new addons ([http://kodi.wiki/view/Submitting_Add-ons#Allowed_submissions](url))? Github seems to offer different information and perhaps the PR needs to go to Jarvis ([https://github.com/xbmc/repo-scripts/tree/isengard](url))?

Thanks.

learningit:
This PR has been discussed among a few us for some time. What I explain below is my personal opinion:
Firstly, your code is good.  I haven't looked through everything, but what I have is good. To fully test this addon will take a long time and even then, the test assumes that the executed scripts are always written correctly.  While it's an interesting concept, it's way beyond a user that doesn't have a fair technical background.  It's almost like a scripting control environment on it's own. We have not approved several wizard/backup/config add-ons because they use external scripts, write to other addon directories, access invalid video sources, etc.  The primary concern that I have is that a user can do a fair amount of damage to both their Kodi install as well as the System OS if a script is rogue.  To the same point this addon could become a vehicle to easily inject code into a Kodi install, as well as the OS which could infect the user's system.  We generally do not include addons in the repo which can be functionally extended externally other than for content access.
An add-on does not have to be included in the kodi.org repo to be widely accepted by users - PTVL is a good example.

KenV99:
@learningit Thank you for your comments.

I understand your point of view. If I may, I would like to offer an alternative way of possibly looking at things. I am not trying to be argumentative or defensive. I know that the Kodi team takes great pride and responsibility in the application. I hope that any replies to the comments below would be in the tone of a healthy, open discussion.

The submitted addon only simplifies things that anyone can do if they understand python and the built-in python addon interface. It allows users who do not wish to go through the process of understanding that interface to be able to run simple tasks in response to Kodi events, similar to the smartphone app IFTTT. I might argue that anyone sophisticated enough to write malicious code to be used via this addon is likely also sophisticated enough to write a full blown addon. Parenthetically, if it is the 'plugin' architecture interface ([http://kodi.wiki/view/Add-on:Kodi_Callbacks#Create_your_own_custom_task](url)) that is included in the addon that is a major cause for concern, that can be easily removed.

As best as I can tell from the forums ([http://forum.kodi.tv/showthread.php?tid=256170](url) and [http://forum.kodi.tv/showthread.php?tid=151011&page=7](url)), the majority of people are using it to automate ambient lighting solutions, run library updates or external scrapers. Users can damage their Kodi installs or OS's with or without the addon and perhaps the addon may make that _less_ likely than if they attempt to do that by writing an addon on their own. Regarding this addon running scripts, the only scripts that are included in the distro are example scripts to demonstrate how the user can have a script run after Kodi exits or run a script that requires sudo. They are not configured to do anything on their own. I can eliminate them and show that example code in the wiki, if desired. The only scripts that the addon does run are ones that the user themselves writes. Perhaps a statement such as 'User assumes all risks and responsibilities associated with running their own tasks' or similar in a popup each time the settings are saved may be appropriate.

If any of the Kodi team members feel that I can 'cripple' the addon by eliminating certain portions of it to make it suitable for the official repo, I would be willing to entertain doing that.

In any case, I will, of course, accept your final decision. My intent is only to expose as many users as I can to the possibilities of what the addon can do for them to make their Kodi experience the best that it can be.

phil65:
Thx for such a balanced response, we have seen much worse here. :)
Moving example scripts to a wiki sounds like a good solution to me. In general I would be fine with including this addon.  @ronie what is your opinion?

ronie:
i haven't reviewed the code, so couldn't give you a balanced opinion ;-)

if the addon doesn't write/overwrite stuff to places it shouldn't go
(outside of it's own addon_data subfolder), it should be acceptable i guess.

i agree script examples would be better fitted in the wiki yeah.

learningit:
Ok. so if there aren't any more comments, I will merge. The PR needs to change to at least Isengard as we are not accepting new add-ons into Helix any longer. The example scripts should be moved to the wiki. Once that's done, I'll merge.

KenV99:
OK. Thank all for your time and consideration. The above changes have been made and a new PR for isengard has been made: [https://github.com/xbmc/repo-scripts/pull/120](url)

Razzeee:
I don't think isendgard accepts new addons? see https://github.com/xbmc/repo-scripts/tree/isengard#status

KenV99:
@Razzeee please see @learnigit 's comment above. The wiki page and the GitHub page have conflicting info about which branches are accepting new PR's.

Razzeee:
Wiki wasn't up to date, I just changed that. Thanks for pointing it out.

learningit:
I'm closing this PR as PR#120 supersedes this one (I hope).

PR #74, opened 2016-03-28
[service.nextup.notification] 1.0.17

im85288:
Fix an issue introduced in the last version


MartijnKaijser:
please make sure the commit message is correct

PR #75, opened 2016-03-30
[service.subtitles.subspedia] 1.0.1

luivit:
general bugfix, thx

### DIFF
--- a/service.subtitles.subspedia/service.py
+++ b/service.subtitles.subspedia/service.py
@@ -35,9 +35,9 @@ def Search(item):
         idserie=checkexp(item['tvshow'])
         linkdownload=""
         eptitolo=""  
-        response = urllib2.urlopen(urlgetid)
-        data = json.loads(response.read())
         if idserie==0:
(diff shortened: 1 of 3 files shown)


service.subtitles.subspedia/service.py
```diff
@@ -182,59 +205,41 @@ def get_params():
...
-            item['3let_language'].append(xbmc.convertLanguage(lang,xbmc.ISO_639_2))
-        Search(item) 
+    item = parseSearchString(params['searchstring'])
+    if item:
+        langstring = urllib.unquote(params['languages']).decode('utf-8')
+        item['3let_language'] = [xbmc.convertLanguage(lang,xbmc.ISO_639_2) for lang in langstring.split(",")]
+        Search(item)
     else:
-        notify(__language__(32002))      
+        notify(__language__(32002))
```
learningit: In the next update please remove all the \__name\__ style labels. We're all guilty of using them, but they are reserved for python internals naming and should not be used. Thanks.

PR #88, opened 2016-04-07
[screensaver.video] 1.4.0

robwebset:

Thanks, Rob


screensaver.video/resources/collections/aquarium.xml
```diff
@@ -0,0 +1,41 @@
...
+		<image>http://kodi.wiki/images/d/df/VideoScreensaver-Aquarium003.PNG</image>
+		<duration>7200</duration>
+		<primary>aHR0cDovL2tvZGkuemlnZ3k3MzcwMS5zZWVkci5pby9WaWRlb1NjcmVlbnNhdmVyL0FxdWFyaXVtMDAzLTcyMHAubXA0</primary>
+	</video>
+	<video>
+		<name>Aquarium 4 - [780MB] - 720p</name>
+		<filename>Aquarium004-720p.mp4</filename>
+		<image>http://kodi.wiki/images/3/39/VideoScreensaver-Aquarium004.JPG</image>
+		<duration>2048</duration>
+		<primary>aHR0cDovL2tvZGkuemlnZ3k3MzcwMS5zZWVkci5pby9WaWRlb1NjcmVlbnNhdmVyL0FxdWFyaXVtMDA0LTcyMHAubXA0</primary>
```
phil65: sorry in case this question has been asked already...
This seem to be base64-encoded URLs.. why do they get hidden? looks suspicious to me.
@ronie

robwebset: It's just a simple way to discourage people from trying to grab the items directly without the addon (not going to stop anyone with any coding knowledge - but does stop novice users) This has been in since the original submission - it's just recently been moved to the xml file.

ronie: please understand one of the goals of team kodi is to promote **open source** software.
code obfuscation doesn't fit that vision i'm afraid.

the addon review process is to make sure we don't distribute illegal content (or links to it) through our addon repository. by obfuscating the url, it's impossible for us to determine what videos you link to.
at least without going through all kind of hoops to decode it.

robwebset: OK, removed the encoding now. The videos are the same as when the screensaver.video was first released and included in the repo - the only new addition is the Apple TV videos which were suggested on the following post:

http://forum.kodi.tv/showthread.php?tid=210943&pid=2146495#pid2146495

learningit:
Why are there 4 commits in this PR?  If you actually want them all please squash into a single commit.
Please see https://github.com/ginatrapani/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit for more info on that.

robwebset:
Sorry - not very good at git - just made a mess of it - will make a new pull request! - Sorry!

PR #95, opened 2016-04-07
[script.sonos] 1.4.3

robwebset:

Thanks, Rob


script.sonos/resources/lib/speech.py
```diff
@@ -1,330 +1,327 @@
...
+
+    # Writes the list of phrases to a file
+    def savePhrases(self, phrases):
+        # Get the location of the speech list file
+        configPath = xbmc.translatePath(ADDON.getAddonInfo('profile'))
+        speechfile = os_path_join(configPath, "speech.txt")
+        log("Speech: Phrases file location = %s" % speechfile)
+
+        try:
+            file_out = open(speechfile, 'w')
```
phil65: please use the "with" statement to make sure that file gets closed properly again.

phil65:
one comment, rest looks fine.

script.sonos/resources/lib/speech.py
robwebset: Done - thanks, Rob

PR #97, opened 2016-04-07
[script.tvtunes] 6.3.0

robwebset:

Thanks, Rob


Razzeee:
Looks good, the only thing that would be nice, but is very much optional is moving the soundcloud lib to  it's own module.

robwebset:
Hi @Razzeee, There was already a stand-alone soundcloud addon, so I didn't think there would be a need for it. It's sort of a cut down version compared to the full addon.
Thanks, Rob

robwebset:
Hi All,

This one has been in the queue for a while, please can it be updated? - had a recent bug report on the forum, when they updated to this version it appeared to be fixed for them

Thanks, Rob

PR #106, opened 2016-04-12
[service.subtitles.legendasdivx] 0.2.7

highlandr:
Addon DEBUG with a bug when it's turned on. My bad...


highlandr:
Hmmm I hope there is no problem with 2 commits ...
Thanks.
Regards.

Razzeee:
Looks good.

Two commits are fine in this case.

learningit:
In the future please use separate PRs for each addon.  It becomes problematic if there is an issue with one of the commits, it will hold up processing the other.

PR #111, opened 2016-04-17
[screensaver.video] 1.4.0

robwebset:
@learningit 
Sorry for messing you about with this one,  this is the corrected pull request from:

https://github.com/xbmc/repo-scripts/pull/88

Thanks, Rob


screensaver.video/addon.xml
```diff
@@ -1,13 +1,14 @@
...
 		<import addon="script.module.simplejson" version="3.3.0"/>
 	</requires>
 	<extension point="xbmc.ui.screensaver" library="default.py"/>
-  	<extension point="xbmc.python.script" library="screensaver.py" />
-	<extension point="xbmc.service" library="service.py" start="login">
-		<provides>executable</provides>
+  	<extension point="xbmc.python.script" library="screensaver.py"/>
+	<extension point="xbmc.service" library="service.py" start="login"/>
+	<extension point="xbmc.python.pluginsource" library="plugin.py">
+		<provides>video</provides>
```
phil65: One thing to consider for the future would be to separate the different extension points, because not everybody who wants the screensaver functionality wants this to appear in the vieos section, too.
This is no blocker , I´m even guilty of this myself (planning to change that though), but would be nice if you would keep this in mind for the future.

robwebset: I have to say, I like that all the related features are together in one addon - I've done this in several addons.  All the features are related, and normally depend on each other.  I think if some-one chooses to use the addon then it's OK to ensure they have all of the addon.

phil65: Just because you have done that on several occasion doesnt mean that it is the correct way to do it ;)
Extension points never "depend on each other". With some clever refactoring you could put shared resources in a separate module add-on and then create separate add-ons for the different extension points. That way people can keep their video add-on list clean when trying out screensavers.

robwebset: I think we'll have to agree to dis-agree on this one. I view the addon as a complete set of features - regarless of the extension points implemented - I think it's valid for me to say "If you want you use VideoScreensaver, you need to take all of it". After all addons are optional - and if used, should be used the way that the author intended :)

phil65: I didnt even mention the fact that our core does not handle all extension point combinations correctly.... Seeems not  to matter though, I understood that you are unconvincable anyways .

PR #112, opened 2016-04-17
[script.colorbox] 1.0.10

BADMS:


learningit:
Sorry there are still two commits in the PR. If you really want both of the commits, you need to squash them down into a single commit. Instructions to do that are here: https://github.com/ginatrapani/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit
When you squash them, please rename the commit to something with the correct version. Right now, your PR has 1.0.10 and your commits (both of them) are named version 1.0.9 . This is extremely confusing.

PR #113, opened 2016-04-18
[script.service.next-episode] 1.0.0

romanvm:
My second attempt to submit this addon :smile: 

This is a program/service addon that allows to synchronize your Kodi video library data with your account on next-episode.net site. Movies and TV episodes are added to your inventory on the site along with their "watched" status. The addon also monitors video playback and updates "watched" status for recently played movies and episodes from media library.


learningit:
Code looks fine. The only issue I see is that for new addons Isengard is the minimum version allowed currently.

romanvm:
Do I need to change the addon re-submit again? Honestly, I don't want to break compatibility with Helix, because this is the maximum version supported on my 3 years old Android tablet which don't receive updates from the manufacturer any more. And I guess, I'm not the only one in such situation.

PR #115, opened 2016-04-18
[script.module.trakt] 2.6.0

Razzeee:


script.module.trakt/addon.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
 <addon id="script.module.trakt"
        name="trakt.py"
-       version="2.5.2"
+       version="2.6.0"
        provider-name="Fuzeman, Razzeee">
   <requires>
     <import addon="xbmc.python" version="2.1.0"/>
```
learningit: Why is xbmc.python set to 2.1.0 for a pull into Jarvis?

script.module.trakt/lib/trakt/interfaces/oauth/device.py
```diff
@@ -0,0 +1,174 @@
...
+            # Poll for token
+            response = self.client['oauth/device'].token(self.device_code, parse=False)
+
+            if response:
+                # Authentication complete
+                self.emit('authenticated', self.get_data(response))
+                break
+
+            # Sleep for defined interval
+            time.sleep(self.interval)
```
learningit: Is there a reason for using time.sleep, rather than xbmc.sleep here? Timing looks like it's in secs so xbmc.Monitor().waitForAbort might be an even better choice.

Razzeee: This module is the 1:1 copy of a python package. See: https://github.com/fuzeman/trakt.py

script.module.trakt/addon.xml
Razzeee: When I did this I think the rule was not really in place and I was much dumber, so this was mostly c&p back then.
Will patch this asap.

script.module.trakt/lib/trakt/interfaces/oauth/device.py
learningit: I also face this problem in places. Don't have an elegant answer, maybe @phil65 has some ideas how to best deal with it?

phil65: I think it can stay this way unless it is known to make problems.

PR #117, opened 2016-04-19
[script.skin.helper.service] 1.0.84

marcelveldt:
Pull request to update script.skin.helper.service to version 1.0.84


script.skin.helper.service/addon.xml
```diff
@@ -10,22 +10,34 @@
...
             <description>Display Series level button in next/recent/in progress episodes listings</description>
-			<visible>Container.Content(episodes) + !StringCompare(Container.ListItem(0).TvShowTitle,Container.ListItem(1).TvShowTitle)</visible>
+			<visible>!IsEmpty(Window(Home).Property(SkinHelper.enablecontextmenu_series)) + Container.Content(episodes) + !StringCompare(Container.ListItem(0).TvShowTitle,Container.ListItem(1).TvShowTitle)</visible>
+		</item>
+	</extension>
+    <extension point="kodi.context.item" library="resources\lib\ContextMenu_pvr.py">
+		<item>
+			<label>32120</label>
+            <description>PVR artwork Configuration</description>
+            <visible>!IsEmpty(Window(Home).Property(SkinHelper.enablecontextmenu_pvr)) + [Window.IsActive(MyPVRChannels.xml) | Window.IsActive(MyPVRGuide.xml) | Window.IsActive(MyPVRRecordings.xml) | Window.IsActive(MyPVRTimers.xml) | Window.IsActive(MyPVRSearch.xml)] + Skin.HasSetting(SkinHelper.EnablePVRThumbs)</visible>
```
phil65: Some XML formatter wouldnt hurt :)
What do these context menu items do? Are they "contextual" at all? Sounds like a shortcut to settings..(and in that case they shouldnt be part of the context menu)

script.skin.helper.service/default.py
```diff
@@ -99,36 +114,50 @@ def __init__(self):
                 if resultAction:
                     if "jsonrpc" in resultAction:
                         xbmc.executeJSONRPC(resultAction)
+                        xbmc.executeJSONRPC(resultAction)
```
phil65: two times same JSON call?

script.skin.helper.service/addon.xml
marcelveldt: The contextmenu pops up a dialog with the specific option for that feature.
Each feature is enabled by the skinner by setting a skin bool. Also, the contextmenu item can be disabled by the user in the addon settings.

For example, the PVR artwork. If enabled by the skinner, the user is able to do some overrides if the auto detection was off or just set the artwork manually.

As we discussed on Github last week I will most likely migrate all code into seperate (module) addons in the near future.

script.skin.helper.service/plugin.py
```diff
@@ -97,27 +125,28 @@ def __init__(self):
...
             utils.logMsg("plugin.py --> Not forfilling request: Kodi is exiting" ,0)
+            xbmcplugin.endOfDirectory(handle=int(sys.argv[1]))
+        elif enableProfiling:
+            import cProfile
+            import pstats
+            import random
+            from time import gmtime, strftime
+            filename = os.path.join( ADDON_DATA_PATH, strftime( "%Y%m%d%H%M%S",gmtime() ) + "-" + str( random.randrange(0,100000) ) + ".log" )
+            cProfile.run( 'Main()', filename )
+            stream = open( filename + ".txt", 'w')
```
phil65: so you open a file... but do you also close it again? :) (preferably use "with" statement)

script.skin.helper.service/default.py
marcelveldt: Well, this may sound silly but I have discovered that the call fails from time to time.
Been playing with it for hours and this was the only "solution".

This will launch a movie/episode from the library if the play button is clicked in the special infodialog. It may be related to the fact that a play action is launched when the dialog is still closing but I couldn't find any documentation about that

script.skin.helper.service/plugin.py
marcelveldt: Good catch, in fact this is just my debug code I use from time to time to check my code for speed. It will never be used by users. Want me to change it right away ?

phil65: Fine if you do it for next version.

script.skin.helper.service/plugin.py
```diff
@@ -46,8 +46,36 @@ def __init__(self):
             elif action == "LAUNCH":
                 xbmcplugin.setResolvedUrl(handle=int(sys.argv[1]), succeeded=False, listitem=xbmcgui.ListItem())
                 path = sys.argv[2].split("&path=")[1]
-                xbmc.executebuiltin("Action(Close)")
                 xbmc.executebuiltin(path)
+            elif action == "FOCUSANDCLICK":
+                #used as workaround to display local media in extendedinfo actorinfo
```
marcelveldt: @phil65 now that you're reviewing the code. This is some ugly workaround I used to get a "in library" container added to extendedinfo.
This is something I'd like to remove from here and add it to extendedinfo. Just another container that will list the local library items for an actor.

Is this something you'd like me to PR to you or you want to implement yourself ?

script.skin.helper.service/plugin.py
marcelveldt: OK, will do

script.skin.helper.service/resources/lib/PluginContent.py
```diff
@@ -1111,7 +1158,7 @@ def getCast(movie=None,tvshow=None,movieset=None,episode=None,downloadThumbs=Fal
             cachedataStr = xbmc.getInfoLabel("ListItem.Title")+xbmc.getInfoLabel("ListItem.FileNameAndPath")+str(downloadThumbs)
     except: pass
     
-    cachedata = WINDOW.getProperty(cachedataStr).decode("utf-8")
+    cachedata = WINDOW.getProperty(cachedataStr+"bla").decode("utf-8")
```
phil65: correct or typo?

script.skin.helper.service/default.py
phil65: Does the response tell you if it was successful?. If yes, then you should probably check that first before triggering a second one.

script.skin.helper.service/resources/lib/PluginContent.py
marcelveldt: oops, and I was wondering why the cache was giving me a hard time. will correct it right away.

script.skin.helper.service/plugin.py
phil65: Could you explain in a bit more detail what you want to achieve?

script.skin.helper.service/default.py
marcelveldt: yes, it does return success indeed, even if the playback doesn't start.
Well, I will just remove that second command for now. If I receive some issues about it I can start looking at it again on some rainy day ;-)

script.skin.helper.service/plugin.py
marcelveldt: On my extendedinfo-actor dialog I have added an extra container which displays the library items for the actor.
This "hack" is the onclick action if a user clicks a movie/show in the list, it will then cycle through the containers (provided by extendedinfo) and select the correct title so it opens up in extendedinfo. This was the only way to get the window stack not joking with me.

I think it makes more sense to just add this properly to the extendedinfo addon and add an extra container with the local library items. That way no weird hacks like this are needed.

If you feel something for this addition to extendedinfo I can subit a PR to you

script.skin.helper.service/default.py
marcelveldt: Oh well, I just remember..it does start but stops again after 1 second, weird huh...
Anyway I removed the extra line for now. We'll just wait untill someone actually complains about it.

script.skin.helper.service/plugin.py
phil65: I might have a look at it (probably included in Titan?), but first impression is that this is too workaround-ish, sorry. I invested quite some energy to get rid of stuff like that. :)

script.skin.helper.service/default.py
phil65: if it is a kodi bug, please create a trac ticket or ping me, perhaps we can fix it the proper way.

script.skin.helper.service/plugin.py
marcelveldt: Yes, that's why I hereby ask you if it can be added to extended info.
In the actorinfo dialog there are some containers: movie roles, tvshow roles, videos etc.
The request is 1 additional container/list: items in library

Yes, it is included in Titan

phil65: ...yeah but that probably requires fetching the complete library including all actorinfo etc via JSON before, right?

script.skin.helper.service/default.py
marcelveldt: Yes, will do. When I have some leftover time, I will invest some time in debugging this thing to do a proper bug report. For now the additional line will be removed.

script.skin.helper.service/plugin.py
marcelveldt: Yes, indeed. query the database for a movie by actor:

PluginContent.py line 1119:

json_result = getJSON('VideoLibrary.GetMovies', '{ "filter": {"operator": "contains", "field": "actor", "value": "%s"}, "properties": [ %s ] }' %(name,fields_movies))

phil65: Any experience how long that may take for large MySQL databases on low-powered platforms?

marcelveldt: just a few ms, the filter works pretty well.

marcelveldt: BTW: I think you already have the query in place because you have the local library count in a window prop in the actor dialog

marcelveldt: image

phil65: It may look similar, but it´s done differently. ;)

marcelveldt: Ah OK, I just made an assumption here, didn't actually look at the code.
Well, just think about it if you want it added (it was actually requested a few times to me on the forums). If you like it I can give it a shot and do the PR. You can always decline the PR if you don't like it haha.

For now I've removed the workaround-ish code from skinhelper.

phil65: Hmm wouldnt that list basically only contain dupes from the other lists? Atm all movies of the filmography which are available locally are positioned at beginning of the lists.

marcelveldt: Yes, I know but users don't understand that ;-) They like to have a small quick list of local library items for an actor.

phil65: hmm will have to think about that bit longer... atm I´m paying great attention to not have any dupes, so my first answer would be "no". (i even cross-compare and filter youtube videos with trailer links returned from tmdb for example to not have duplicates)

marcelveldt: OK, no problem at all. Maybe a very simple solution would be to add a "in library" property to the listitems. The skinner can than display some tag/overlay for items in the library

phil65: Already there, in older versions it´s ListItem.property(dbid), in newer ones it is ListItem.DBID.
See https://github.com/xbmc/xbmc/pull/9627

marcelveldt: You're the best ;-) Also great PR with the ListItem.DBID, I'm indeed also working around that with the ListItem.property(dbid) in skinhelper.

In the meanwhile I've added another commit (didn't squash it) to remove the ugly hack and fix the few typos you pointed me to. I will adjust my skin to check against ListItem.Property(dbid) in the extendedinfo actor dialog to show some overlay to notify it's in the library.

Thanks!

phil65: Be aware that this property will be gone for Krypton. I dont spend too much thaught on backwards compatibility when it comes to removing hacks. ;)

marcelveldt: haha!
Well a simple !IsEmpty(ListItem.Property(dbid) | [!IsEmpty(ListItem.DBID) + !StringCompare(ListItem.DBID,-1) will do the trick than :-)

marcelveldt:
@phil65 are you happy with the changes ? OK to merge ?

PR #119, opened 2016-04-20
[service.nextup.notification] 1.0.20

im85288:
Sorry for the quick additional PR, found a critical bug with the last changes. I put back the sleeps for the ones that are under a second.


learningit:
Yeah it needs to be an int and is in seconds, I should have caught that.

im85288:
No worries, thanks for the quick merge.

PR #121, opened 2016-04-21
Master

russellharrower:


Razzeee:
This repo doesn't accept PRs against it's master branch. You have to choose a version. The lowest version your addon supports
a) If this is a new addon you need to choose Jarvis at minimum
b) If this is an addon already in this repo choose the correct prior versin that still accepts pushes. You can see in the readme of each branch. For e.g. https://github.com/xbmc/repo-scripts/tree/isengard#status

PR #122, opened 2016-04-23
[script.tvtunes] 6.3.1

robwebset:
Please can I have the following small update merged in. (Prevents needless processing)

Thanks,
Rob

### DIFF
--- a/script.tvtunes/addon.xml
+++ b/script.tvtunes/addon.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<addon id="script.tvtunes" name="TvTunes" version="6.3.0" provider-name="robwebset">
+<addon id="script.tvtunes" name="TvTunes" version="6.3.1" provider-name="robwebset">
 	<requires>
 		<import addon="xbmc.python" version="2.20.0"/>
 		<import addon="script.module.simplejson" version="3.3.0"/>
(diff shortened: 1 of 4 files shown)


robwebset:
Thanks @MartijnKaijser

SchapplM:
I noticed, that a few files in your addon have CRLF line endings checked in (instead of LF).
That leads to a conflict on linux systems (conflicted files in git repo prevent ability to merge/stash).
Could you change the endings in the following files?

	               plugin.py
	               resources/lib/idLookup.py
	               resources/lib/library.py
	               resources/lib/screensaver.py
	               resources/lib/themeFinder.py
	               resources/lib/upload.py
	               saver.py

I did not see this in your addon-repo https://github.com/robwebset/script.tvtunes though.
Thanks

tamland:
@SchapplM there is a .gitattributes that should make it impossible to submit crlf but I think the problem is that it was already in in some branches, and it wasn't cleaned properly when added. Feel free to follow https://help.github.com/articles/dealing-with-line-endings/#refreshing-a-repository-after-changing-line-endings and submit a PR.

robwebset:
Hi @SchapplM 

No problem, new pull request: https://github.com/xbmc/repo-scripts/pull/134

I have noticed that some of my other addons also suffer the same issue - should I convert those and do a new pull request for them as well?

Thanks
Rob

PR #123, opened 2016-04-23
[script.colorbox] 1.0.10

BADMS:
Hopefully within guidelines for 1 commit per pull.

### DIFF
--- a/script.colorbox/Readme.md
+++ b/script.colorbox/Readme.md
@@ -9,6 +9,8 @@ script.colorbox
   
   pixelate: will bulk up pixels to size requested and add black border round each pixel block
   
+  distort: will fuzzy spread image using two deltas, x & y.
+  
   blur: will return a guassian blurred image dependant on radius supplied, larger radius means larger blur + color match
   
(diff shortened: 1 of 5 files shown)


learningit:
This looks OK to me, but I'm not sure that all the assumptions that are made about cache files are valid in newer versions of Kodi.  Someone with more knowledge will need to comment on that.

BADMS:
Well they were fine with first release for Jarvis [and have remained the same since].
Please direct me to documentation otherwise, thanks!

script.colorbox/Readme.md
```diff
@@ -75,15 +81,29 @@ script.colorbox
...
+  
+  SetProperty(cfa_daemon_set,Pixelate,home)
+  
+  SetProperty(cfa_daemon_set,Posterize,home)
+  
+  SetProperty(cfa_daemon_set,Distort,home)
+  
+  This MUST be used as of now.
+  NONE will be processed if not set.
+  Next version will default to ALL being processed if not set!
```
phil65: this doesnt sound good. Default = off should be preferred.

BADMS: noted, will default off with and add property to call all. thanks

PR #127, opened 2016-04-26
[service.subtitles.divxplanet] v0.1.1

korayal:
The older version was v0.0.12 and it's in the gotham repository.

I plan to support isengard and above from now on.


service.subtitles.divxplanet/service.py
```diff
@@ -0,0 +1,351 @@
...
+reload(sys)  # Reload does the trick!
+sys.setdefaultencoding('UTF8')
+
+__addon__ = xbmcaddon.Addon()
+__author__ = __addon__.getAddonInfo('author')
+__scriptid__ = __addon__.getAddonInfo('id')
+__scriptname__ = __addon__.getAddonInfo('name')
+__version__ = __addon__.getAddonInfo('version')
+__language__ = __addon__.getLocalizedString
+
```
phil65: Doesnt need to be done for this pull request, but in the future  please get rid of `__varname__` naming style.

phil65:
code looks fine.

service.subtitles.divxplanet/service.py
korayal: got it, thanks! will do on the next release.

PR #131, opened 2016-04-30
[service.nextup.notification] 1.0.22

im85288:


learningit:
Can I ask you to have a look at your implementation of lock in lock.py? You're using time.sleep in the spin for acquiring the lock which is going to affect thread processing. Are you able to use xbmc.sleep instead?

im85288:
Yes sure, can we leave that for the next pull request though as I fear it may take a bit of testing and this version is requested by a few people to have the unwatched setting as false by default.

learningit:
Yeah, I already did the merge. I would actually like you to look at the whole lock mechanism and the reasoning behind it. I'm struggling to understand why you need this - I don't see multiple threads accessing the file (though I haven't looked too closely) and I would think you would use the native file locking (but again I haven;t looked too closely). What I can tell you is that Kodi thread yielding/event processing is getting screwed because of the use of time.sleep.  This seems to be causing a number of display issues.

im85288:
Thanks, yes I don't think I actually need the lock in there at all. I'll take a closer look this weekend hopefully.

PR #136, opened 2016-05-01
Screensaver.tvtunes 1.0.0

robwebset:
This pull request is for the split of TvTunes screensaver. It really needs to be done just after the TvTunes pull https://github.com/xbmc/repo-scripts/pull/135 which removed the screensaver from that Addon so that it can be in it's own addon.

Thanks, Rob


robwebset:
For some reason it pulled in the tvtunes changes as weel - -I'll code and have another go

PR #137, opened 2016-05-02
[screensaver.tvtunes] 1.0.0

robwebset:
This pull request is for the split of TvTunes screensaver. It really needs to be done just after the TvTunes pull https://github.com/xbmc/repo-scripts/pull/135 which removed the screensaver from that Addon so that it can be in it's own addon.

Thanks, Rob


phil65:
This doesnt really solve the "issue" i mentioned though since you list script.tvtunes in your dependencies. The proper way would be to move shared logic out into a separate module add-on.

robwebset:
Hi @phil65 

script.tvtunes has all the code to download themes etc as well - I don't believe that anyone would want to use screensaver.tvtunes without having code tvtunes sunning, so I think it's fine having the entire thing as a dependency. The change does at least prevent users that don't want the screensaver automatically having it when they install script.tvtunes.

Please can this one be merged asap as without it anyone using the screensaver part of TvTunes will be broken when the other update #135 gets pushed to their installations.

Thanks, Rob

phil65:
Since this doesnt make things worse I will consider it fine then for now.

PR #146, opened 2016-05-06
[script.module.t1mlib] 1.0.15

learningit:


script.module.t1mlib/addon.xml
```diff
@@ -13,7 +13,7 @@
         <email></email>
         <website></website>
         <forum>http://forum.kodi.tv/showthread.php?tid=247163</forum>
-        <source>https://github.com/learningit/repository.t1m.v2.addons/tree/master/src/script.module.t1mlib</source>
+        <source>https://github.com/xbmc/repo-scripts/tree/isengard/script.module.t1mlib</source>
```
Razzeee: This should be your own repo not this one.

learningit: Why?

Razzeee: Well if this was valid/wanted we could generate this data for our devs. 
I think the idea behind it was too drive traffic to your repo with possible questions, PRs or reviews. Which now might end up on our repos instead. Our they realize it and will have to search for the correct repo to do a PR against.

learningit: Reasonable. The issue is that I am no longer maintaining the sources in the individual repos as I have way too many add-ons to keep messing with multiple versions. I do all the work against what is in the official repo, so the ones I have are not up to date. I am having serious issues with people loading the add-ons from anywhere and it gets more confusing under Jarvis because users have no idea which repo they should update against. It's a real mess.

Razzeee: I feel your pain :/
Not quite sure how to solve it. We could ask github for an option to disable the download zip button? :D

learningit:
Fixed the source repo in all the PRs

learningit:
Martijn, Thanks for the merge. That problem was a bitch to find.

PR #147, opened 2016-05-07
[script.service.hyperion] 1.0.2

brindosch:
Hello Kodi team
Initial commit/add request

I know that you no longer accept plugins in older repositorys then jarvis, but due to the point this addon is very old and compatible up to the latest Krypton builds. You may consider a exception.
Please leave some feedback.

Thank you


phil65:
If the add-on broke with krypton, then send a pull request with the updated add-on to "krypton" branch.

brindosch:
Thank you for feedback. You mean in case it may broke with later Krypton builds? If i understood it right, with old branch is meant -> visible up to the latest (krypton) version. Am i wrong?
Thank you

phil65:
If you add an addon with the same id to the krypton branch of this repository, then krypton users will get that new one instead of the old one. All previous kodi versions will get the old one.

phil65:
Ok, I misunderstood and thaught because of the version number that this would be an update and not an initial addition. Please add this to "jarvis" branch then, we do not make exceptions unless there are special reasons.

brindosch:
Okay :)
Thank you

PR #148, opened 2016-05-08
[service.nextup.notification] 1.0.24

im85288:


service.nextup.notification/resources/lib/NextUpInfo.py
```diff
@@ -43,6 +44,14 @@ def onInit(self):
...
         self.getControl(3004).setLabel(info)
 
+
+        try:
+            tvShowControl = self.getControl(3007)
+            if tvShowControl != None:
+                tvShowControl.setLabel(tvshowtitle)
+        except:
+            pass
+
```
phil65: just a small remark: python has something called "for-loops". :grin:

PR #151, opened 2016-05-09
[service.audio.profiles] 1.0.1

Regss:


phil65:
Reason?

phil65:
Ah I see, you converted it to a script? https://github.com/xbmc/repo-scripts/pull/153
Some small explanation would help in these cases so we know your intention :)

Regss:
I moved service to script.audio.profiles because it is a integrity part of this addon. When I do PR for other addon i get suggest from kodi team member to use multiple extension points.

PR #152, opened 2016-05-10
[service.subtitles.legendasdivx] 0.2.8

highlandr:
Android\Linux crash solved. _temp variable not decoded to utf-8 by default.

### DIFF
--- a/service.subtitles.legendasdivx/service.py
+++ b/service.subtitles.legendasdivx/service.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Service LegendasDivx.com version 0.2.7
+# Service LegendasDivx.com version 0.2.8
 # Code based on Undertext (FRODO) service
 # Coded by HiGhLaNdR@OLDSCHOOL
 # Ported to Gotham by HiGhLaNdR@OLDSCHOOL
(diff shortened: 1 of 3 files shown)


highlandr:
Any1 alive?:)
Cheers!

service.subtitles.legendasdivx/service.py
```diff
@@ -38,7 +38,7 @@
 _cwd        = xbmc.translatePath(_addon.getAddonInfo('path')).decode("utf-8")
 _profile    = xbmc.translatePath(_addon.getAddonInfo('profile')).decode("utf-8")
 _resource   = xbmc.translatePath(os.path.join(_cwd, 'resources', 'lib' )).decode("utf-8")
-_temp       = xbmc.translatePath(os.path.join(_profile, 'temp')).decode("utf-8")
+_temp       = xbmc.translatePath(os.path.join(_profile, 'temp'))
```
phil65: hmm doesnt really make sense to me to .decode() some paths, and some others not? youre sure that youre not just hiding some other issues with this change?

highlandr:
Me neither. I will re-check the code. The problem is always the same, accents from our latin language. Since  I use the interface language in English I had no issues. I was warned by users. Thanks.

highlandr:
I've been checking the code when I get time and If I don't decode the _temp variable I get no 'ascii' errors in Android. It's the only OS that I've been getting problems. I didn't use it before, It was something I added 2 versions ago cause It made the think why I didn't before. Now I know why.
If you wanna allow this change it would be the best way to get this fixed in my humble POV.
Thanks.

phil65:
I still think that the issue is somewhere else, but if you have tested it I am fine with this change.

PR #155, opened 2016-05-15
[service.nextup.notification] 1.0.25

im85288:
oops tvshowtitle is not available for episodes...use showtitle instead


phil65:
Changelog update?

im85288:
Now updated thanks

PR #156, opened 2016-05-16
[script.xbmc.unpausejumpback] 2.3.5

Memphiz:
This replaces the icon with something public domain.

@NedScott - just fyi - i asked for a new icon 2 times (to independend design guys) - and i never heard back and then just forgot about it.


phil65:
Could you add a non-transparent background to the icon, please? Some skins add a border texture, and that does not look too well with transparent icons :)

Memphiz:
is white as a background color ok then?

Memphiz:
done @phil65

PR #157, opened 2016-05-16
[script.user.rating] 1.0.3

Regss:


learningit:
A couple of general comments to be considered for your next update:

1) Please avoid using \__varname\__ style labels, these should only be used for python internals.

2) I see you using an index to keep count in loops like this:
for item in items:
   item_added += 1

A more pythonic way to do this is:
for item_added, item in list(enumerate(items, start=1)):

I've only reviewed your changes as this is an update. They look ok, with the above comments.

phil65:
`for item_added, item in enumerate(items, start=1)` is enough. ;)

learningit:
Yep, my bad - had trouble enough trying to remember if it was enum() or enumerate().

PR #160, opened 2016-05-17
script.module.kodi65 1.0.0

phil65:


phil65:
That untranslated notification should never occur if I do my work correctly :)

learningit:
That's what I was thinking there.

I've been through all this and it looks OK. I only scanned through because
I'm on a cell phone. Do you want me to merge?

On May 17, 2016 1:19 PM, "Philipp Temminghoff" <notifications@github.com>
wrote:
>
> That untranslated notification should never occur if I do my work
correctly :)

phil65:
yup, i need that workaround because I pass strings via params which may contain an ampersand.  Not nice, but sadly there´s no better alternative.

PR #162, opened 2016-05-17
[script.trakt] 3.1.7

Razzeee:


Razzeee:
Updated to need https://github.com/xbmc/repo-scripts/pull/165 as it fixes an important bug

learningit:
OK, I've never heard of an important bug, but I'll merge the important fix.

PR #171, opened 2016-05-22
[script.recap] 1.0.0

robwebset:
A new addon for inclusion please.

Details available from the following forum post:

http://forum.kodi.tv/showthread.php?tid=275040

Thanks, Rob


script.recap/addon.xml
```diff
@@ -0,0 +1,31 @@
...
+		<import addon="xbmc.python" version="2.20.0"/>
+		<import addon="script.module.beautifulsoup" version="3.2.1"/>
+		<import addon="script.module.requests" version="2.3.0"/>
+	</requires>
+  	<extension point="xbmc.python.script" library="default.py"/>
+	<extension point="xbmc.service" library="service.py" start="login"/>
+	<extension point="kodi.context.item" library="default.py">
+		<item>
+			<label>32001</label>
+			<visible>IsEmpty(Window(videos).Property("RecapHideContextMenu")) + [Container.Content(tvshows) | Container.Content(seasons) | Container.Content(episodes)]</visible>
```
phil65: Use ListItem.DBType instead of Container.Content.

script.recap/confupdate.py
```diff
@@ -0,0 +1,132 @@
...
+\t\t\t\t\t\t<description>Recap</description>
+\t\t\t\t\t\t<include>ButtonInfoDialogsCommonValues</include>
+\t\t\t\t\t\t<label>$ADDON[script.recap 32001]</label>
+\t\t\t\t\t\t<onclick>RunScript(script.recap)</onclick>
+\t\t\t\t\t\t<visible>System.HasAddon(script.recap) + [Container.Content(tvshows) | Container.Content(seasons) | Container.Content(episodes)]</visible>'''
+
+        insertTxt = previousButton + (DIALOG_VIDEO_INFO_BUTTON % idval)
+        dialogXmlStr = dialogXmlStr.replace(previousButton, insertTxt)
+
+        self._saveNewFile(dialogXml, dialogXmlStr)
```
phil65: Overwriting DialogVideoInfo.xml is a no-go.

robwebset: This is exactly the same as VideoExtras and TvTunes - a warning is displayed and the user knows exactly what was going on.  These were discussed and agreed were OK in the forum.

script.recap/addon.xml
phil65: Hmmm or was ListItem.DBType introduced with Jarvis? Would need to get checked first.

script.recap/confupdate.py
phil65: Please point me to the discussion. My vote goes to "not allowed", esp. bc youre modifying an XML without any XML parser. That looks quite frightening to my eyes.

robwebset: Not using an XML parser (which was my first implementation to the original TvTunes) was because all the XML parsers that seem to be available for Kodi reformatted the entire file - which I really didn't want to do!

One discussion:
http://forum.kodi.tv/showthread.php?tid=214289&pid=1890088#pid1890088

Also in Suitability which is discussed in the mailing list, detailed in this thread with Martijn:

https://sourceforge.net/p/xbmc/mailman/message/34774499/

script.recap/addon.xml
robwebset: Wouldn't that require a string comparison, isn't the Content cleaner when reading it?

script.recap/confupdate.py
phil65: Well, @tamland sums up the concerns quite well in that discussion. I brought this up again within our team, and we wont allow this kind of stuff anymore. Please remove it, without makin a big drama out of it.

script.recap/addon.xml
phil65: Might be clearer when reading, but it´s wrong semantically and doesnt work for edge cases. (example: movies can get listed on season level with "link-to-tvshow" feature)

script.recap/confupdate.py
ronie: i reckon the wording in the wiki was poor at best (and have been updated since then),
but it always included this clause:
`Exceptions to this rule may be granted in specific cases only. Please contact Team XBMC's add-on repository maintainers via the add-on mailing list if your add-on needs access to such files.`

from the forum discussion, it's pretty clear both Martijn & tamland strongly opposed to the modification other addon's files.

please note that altering the default skin will fail on all platforms except windows, as it's either installed on a read-only partition (openelec) or requires root privileges to make such changes.

robwebset:
Updated as per requests

script.recap/resources/lib/core.py
```diff
@@ -0,0 +1,233 @@
...
+
+#########################
+# Main
+#########################
+def runRecap(previous=False):
+    log("runRecap: Started")
+
+    # First check to see if we have a TV Show already selected
+    showName = None
+    if xbmc.getCondVisibility("Container.Content(tvshows)") or \
```
phil65: ideally this here should also be done via listitem.dbtype

script.recap/resources/lib/core.py
```diff
@@ -0,0 +1,233 @@
...
+       xbmc.getCondVisibility("Container.Content(Seasons)") or \
+       xbmc.getCondVisibility("Container.Content(Episodes)"):
+        showName = xbmc.getInfoLabel("ListItem.TVShowTitle")
+
+    # If there is no video name available prompt for it
+    if showName in [None, ""]:
```
phil65: nit: `if showname:`

script.recap/resources/lib/core.py
```diff
@@ -0,0 +1,233 @@
...
+        log("Recap: No TV Show detected, prompting user")
+
+        # Prompt the user for the new name
+        keyboard = xbmc.Keyboard('', ADDON.getLocalizedString(32014), False)
+        keyboard.doModal()
+
+        if keyboard.isConfirmed():
+            try:
+                showName = keyboard.getText().decode("utf-8")
+            except:
```
phil65: why should an exception occur here?

script.recap/resources/lib/core.py
```diff
@@ -0,0 +1,233 @@
...
+        # Check which season will be displayed
+        if len(seasons) == 0:
+            log("Recap: No seasons found for %s" % searchMatches[selectedItem]["link"])
+        elif len(seasons) == 1:
+            log("Recap: Only one season exists")
+            selectedSeason = seasons.keys()[0]
+        else:
+            # Check if we already have a season selected
+            seasonNumber = None
+            if xbmc.getCondVisibility("Container.Content(Seasons)") or \
```
phil65: listitem.dbtype

script.recap/resources/lib/viewer.py
```diff
@@ -0,0 +1,165 @@
...
+        if Settings.isAutoSlideshow():
+            previousControl.setVisible(False)
+            nextControl.setVisible(False)
+        # If we are looping then need to always show the next and previous buttons
+        elif Settings.isLoopSlideshow():
+            previousControl.setVisible(True)
+            nextControl.setVisible(True)
+        else:
+            if idx < 1:
+                previousControl.setVisible(False)
```
phil65: previousControl.setVisible(idx >= 1)
;)

script.recap/resources/skins/Default/720p/script-recap-window.xml
```diff
@@ -0,0 +1,197 @@
...
+            <texturenofocus border="5">button-nofocus.png</texturenofocus>
+            <texturefocus border="5">button-focus.png</texturefocus>
+			<textcolor>ffffffff</textcolor>		<!-- white -->
+            <onright>305</onright>
+            <onleft>302</onleft>
+            <onup>203</onup>
+            <ondown>203</ondown>
+        </control>
+		<!-- The following are a list of cached images -->
+		<control type="image" id="401"><posx>40</posx><posy>636</posy><width>12</width><height>5</height><texture background="true"></texture></control>
```
phil65: perhaps better use a container instead of this?

phil65:
some nitpicks (feel free to ignore them).
Apart from those it looks well enough to me. @learningit mind takin another look?

tamland:
addon.xml is invalid (two extensions of the same type)

robwebset:
> addon.xml is invalid (two extensions of the same type)

It does actually seem to work OK. Looking at the schema in isengard there doesn't appear to be a documented way of having multiple call different scripts.

https://github.com/xbmc/xbmc/blob/Isengard/addons/xbmc.python/contextitem.xsd

However this seems to change in later releases.  But it does work well as it is in terms of providing the required behavior. Is it possible to allow it in this instance - or do you know of another way that it can be done in the same addon?

Thanks, Rob

robwebset:
> some nitpicks (feel free to ignore them).

Thanks, will take a look at the m for the next pull request if that's OK.

tamland:
Can't be done in Isengard. All explained in http://kodi.wiki/view/Context_Item_Add-ons It's just a bug that makes it seemingly work, and it's gone in more recent releases. 

That's another thing btw: unless I'm mistaken we don't accept new addons to Isengard so this PR should be moved to Jarvis, dropping backwards compatibility anyway.

robwebset:
OK, will kill this PR and re-work for Jarvis with the new structure

Thanks, Rob

script.recap/resources/lib/core.py
robwebset: OK, will do that in the next pull

script.recap/resources/lib/core.py
robwebset: Had loads of issues with character encoding, this is just a safety thing :)

script.recap/resources/lib/core.py
robwebset: Will change in replacement pull request

script.recap/resources/lib/viewer.py
robwebset: I'm a bit old-school, also, there is a more complex comparison about 5 lines below it, and this way there is a certain amount of symmetry to the code

script.recap/resources/skins/Default/720p/script-recap-window.xml
robwebset: If it switched to a Container, I'm not sure I'd be able to access if from Python via the API would I? I'd need that to be able to set the images.

PR #173, opened 2016-05-23
[service.subtitles.traduttorianonimi] 2.0.4

ShellAddicted:
Hi. The version of my addon (2.0.3) merged 15 hours ago. works only on Windows and Linux but not on android devices (tested on Android 5.1.1 with kodi 16.1 apk downloaded from kodi.tv) because version of logging module is lower than "0.5.1.2", so now i fixed my addon and now is compatibile with lower versions of logging (removing "getChild()"), another thing, python 2.6 requires that "{}".format(something) must have and integer like this => ("{0}".format(something), so if fixed also this.
-Why the SAME version of kodi uses different version of python on android ???

### DIFF
--- a/service.subtitles.traduttorianonimi/resources/lib/TraduttoriAnonimi.py
+++ b/service.subtitles.traduttorianonimi/resources/lib/TraduttoriAnonimi.py
@@ -35,7 +35,7 @@ resource = xbmc.translatePath(os.path.join(cwd, "resources", "lib")).decode("utf
 
 sys.path.append(resource)
 from utils import *
-log=log.getChild("Core")
+#log=log #.getChild("Core") #Disabled for android compatibility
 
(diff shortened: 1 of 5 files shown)


service.subtitles.traduttorianonimi/resources/lib/TraduttoriAnonimi.py
```diff
@@ -35,7 +35,7 @@
 
 sys.path.append(resource)
 from utils import *
-log=log.getChild("Core")
+log=log #.getChild("Core") #Disabled for android compatibility
```
phil65: log=log ?

phil65:
Different python versions is solved with krypton, there we have 2.7 for all platforms.

ShellAddicted:
Log=log should be commented now. I'm not sure because github is not working correctly.

ShellAddicted:
Something else?

PR #174, opened 2016-05-23
[service.subtitles.torec] 1.0.9

slmosl:
* fixed some major changes with the provider (Torec) API 
* major code beautification refactor

### DIFF
--- a/service.subtitles.torec/resources/lib/TorecSubtitlesDownloader.py
+++ b/service.subtitles.torec/resources/lib/TorecSubtitlesDownloader.py
@@ -132,9 +132,11 @@ class TorecSubtitlesDownloader(FirefoxURLHandler):
             "current_datetime": current_time
         }
 
-    def _get_user_auth(self):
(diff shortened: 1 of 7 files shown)


service.subtitles.torec/resources/lib/TorecSubtitlesDownloader.py
```diff
@@ -1,165 +1,244 @@
...
+        params = {
+            "sub_id":     sub_id,
+            "code":       option_id,
+            "sh":         "yes",
+            "guest":      self._request_subtitle(sub_id),
+            "timewaited": "11",
+            "userAuth":   self._get_user_auth()
+        }
+
         for i in xrange(16):
```
phil65: please use xbmc.monitor to check for kodi exit. see http://kodi.wiki/view/Service_addons

slmosl: thanks @phil65, your comment actually made realize that the retry mechanism is no longer needed for this feature, so I removed it altogether.

PR #175, opened 2016-05-25
[artistslideshow] 2.0.0

pkscout:
v.2.0.0
- separated code for downloads into modules to make it easier to add other services
- added preferences to allow selection of services and priority for use
- removed remote MusicBrainz lookup, MBID must now be in Kodi
- changed default to get bio from Kodi and images from htbackdrops
- rewrite of cache mechanism to reduce load on external sites
- changed cache expire range to 12 - 24 weeks (set randomly when cache is refreshed)
- updated preference strings (only French and English included, other languages
  will be added back later via Kodi Transifex project)


ruuk:
Thanks for cleaning up the __x__ style labels :)

PR #177, opened 2016-05-26
[script.recap] 1.0.0

robwebset:
A new addon for inclusion please.

Details available from the following forum post:

http://forum.kodi.tv/showthread.php?tid=275040

This is a replacement pull request for #171 

Thanks, Rob


script.recap/addon.xml
```diff
@@ -0,0 +1,31 @@
...
+		<import addon="script.module.beautifulsoup" version="3.2.1"/>
+		<import addon="script.module.requests" version="2.3.0"/>
+	</requires>
+  	<extension point="xbmc.python.script" library="default.py"/>
+	<extension point="xbmc.service" library="service.py" start="login"/>
+	<extension point="kodi.context.item">
+		<menu id="kodi.core.main">
+			<item library="default.py">
+				<label>32001</label>
+				<visible>IsEmpty(Window(videos).Property("RecapHideContextMenu")) + [StringCompare(ListItem.dbtype,tvshow) | StringCompare(ListItem.dbtype,season) | StringCompare(ListItem.dbtype,episode)]</visible>
```
phil65: @tamland is it fine to use this hack for setting context menu item visibility or should we stick to separate add-ons for context items?
This add-on uses the xbmc.service extension point just for setting these properties.. (see service.py)

robwebset: This is also how Suitability, TvTunes and VideoExtras do it. It allows the user to configure if it is shown, otherwise I'd just always force them to be there - which didn't seem very nice (I believe giving the user the option is much better). Ideally it would read directly from the addon settings - but I could not see a way of doing that.

phil65: Just because some other add-ons do it doesnt mean that it is correct.
I think ideally it would be a separate context item add-on, then the user can choose whether he wants that context menu entry or not, without any workarounds.
Let´s wait for tamland´s opinion.

robwebset:
Hi All,

Any progress on getting this one merged in.  @tamland had previously seen the change in https://github.com/xbmc/repo-scripts/pull/171 where he commented on something else on exactly the same line and did not highlight this as an issue.

Thanks, Rob

script.recap/addon.xml
ruuk: I would say if we **_are_** ok with it, then it should be: ```<visible>!IsEmpty(Window(videos).Property(RecapShowContextMenu))```
so if the service fails for some reason the context menu is not shown.

robwebset: Hmm, I'd actually want it so the default is enabled/showing - that way if "the service fails" (But not sure why it would) - then the context menus are shown - not hidden. After all this is how the vast majority of people will use the addon.

ruuk: It may just be me, but I don't like the idea of add-ons adding context menu items by default, and I was assuming (which I shouldn't have done) this one didn't. I understand the reasoning and have been tempted myself. It doesn't seem bad in the context of a single addon, but eventually we can run into a situation where users have a context menu full of items which they may not want, and are not sure which installed addon provided it. This is especially true with add-ons that require the context menu addon or (as is the case here) include it in the same add-on.

@phil65: Thoughts? I'm fine with it otherwise, as the extension points are not doing anything wrong themselves, and the service is not a long running service. If he breaks the context menu out as an addon, that doesn't solve the issue, because he could just add it as a dependency, and then we're in basically the same position (unless we're not allowing context menu items as a dependency - and if that's the case then the point is moot).

robwebset:
Any more news on this?

RThanks, ob

script.recap/resources/lib/core.py
```diff
@@ -0,0 +1,230 @@
...
+
+#########################
+# Main
+#########################
+def runRecap(previous=False):
+    log("runRecap: Started")
+
+    # First check to see if we have a TV Show already selected
+    showName = None
+    if xbmc.getInfoLabel("ListItem.dbtype") in ['tvshow', 'season', 'episode']:
```
phil65: This way wont work when triggered from any non-media window (home screen for example) via context menu. You need to check sys.listitem in that case, otherwise it will be buggy in krypton.

phil65:
Concerning the context menu entries: In general I also dislike putting all those different extension points into one single add-on, in this case it might be fine since the context menu is the only way to access this script. Is that correct, @robwebset ?

script.recap/resources/lib/core.py
phil65: Also a general note (feel free to ignore): You are doing these "if x in y" checks quite often. Those checks perform a lot better when "y" is a set, and not a list. ( ---> if x in {a, b, c} )

robwebset: Hi @phil65 - I'm afraid I don't follow you for the first point - this should only be called from the tvshow/season/episode windon (not the home window) - do you have an example by any chance? (The context menu should not be shown anywhere else due to the settings in addon.xml)

The the second point - interesting, never really thought about using a set for that sort of lookup - will keep it in mind for future changes.

Thanks, Rob

robwebset:
@phil65 Yes, the context menu is the way all users will access the addon (To check what's happened before watching the next episode).  It is possible to launch it from the Programs Section (from the Home page) as it is a script. (But that will just prompt the user to enter the name of the TV Show - I don't think that approach will be used very much)

Thanks, Rob

script.recap/resources/lib/core.py
phil65: since krypton you also get a context menu for widgets, and for these widgets xbmc.getInfoLabel("Listitem.xxx") wont work. Just try latest nightly and press "c" when focussing an episode / tvshow.

phil65: In case youre interested: https://stackoverflow.com/questions/2831212/python-sets-vs-lists

ruuk: Just a small note on sets (you got me looking them up as I have never used them) the {} notation apparently only works in python 2.7+ which I don't think we can't count on until Krypton. Ignorable opinion follows: In a three element/single iteration like this I would personally use a tuple - based on the assumption that it is a lower resource using object and also that in most cases (I believe) with very small sets of data you don't get any benefit from more complex lookup algorithms.

phil65: @ruuk: yup, correct. It's only really relevant for long lists or codepaths which are often used (thats why i said it doesnt really matter in this specific case, but it's still good to know i think)

ruuk:
@robwebset: Just so that I understand, what is the purpose of the script portion of the addon then?

robwebset:
@ruuk that enabled users to input a TV Show that they don't currently have on their Kodi System. Usecase: The user watches Game Of Thrones on their Satellite TV provides channel - they don't own any of the DVDs/BluRays but want to be reminded what happened. They select Recap from the Programs section - it prompts for a title to  be entered - they enter "Game Of Thrones" and it displays options for that program.

robwebset:
Hi All, I think everything is covered now isn't it?

Are we good to get this merged?

Thanks, Rob

robwebset:
Hi All,

Sorry to bump this again, but is there anything still outstanding - would we be able to get this in now?

Thanks, Rob

phil65:
I would like to have an ACK from @tamland first if it is fine to abuse the service extension point in the way you did.

robwebset:
We have been waiting for almost 3 weeks for @tamland to comment maybe @ronie could help out?

Thanks, Rob

tamland:
Sorry. Didn't know people were waiting for me.. To be clear: this 'feature' of accessing any sort of global state inside the visibility condition doesn't exist and is not supported, and is going to break the moment e.g caching is added or the context menu feature is extended with something. And if it does, I'll certainly file it under wontfix. 
That being said, I don't see how it's worse than many of the current skin hacks being used.. so it's fine with me.

tamland:
need to squash though..

robwebset:
Thanks @tamland - should be down to a single commit now

Thanks, Rob

phil65:
ok, thx @tamland.

robwebset:
Thanks all, good to finally get this one in

Rob

PR #180, opened 2016-05-31
[script.tablo] 1.0.0

ruuk:
Could someone please review this? I don't want to just push an new addon I've authored.

Please let me know if all is well to merge but let me do the merge. I'm trying to time the release with a firmware update.

Thanks :)


script.tablo/addon.xml
```diff
@@ -0,0 +1,25 @@
+﻿<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<addon id="script.tablo"
+       name="Tablo"
+       version="1.0.0"
+       provider-name="Nuvyyo">
+    <requires>
+        <import addon="xbmc.python" version="2.14.0"/>
```
learningit: xbmc.python should be 2.24.0 for Jarvis.

learningit:
747 files? Even commenting on this is pretty slow in github.

script.tablo/addon.xml
ruuk: That would leave it unusable on previous versions. Since I can't put it in the older repos, I was hoping to keep the version as is. Is this a hard and fast rule? I thought this was only needed if you used features that were only available on that version.

learningit: We have been enforcing this for new add-ons. It's easier all around. We don't have to move it to a new repo when support for the old one stops, people should be updating to newer versions and we encourage add-on authors to take advantage of the newer

ruuk:
@learningit: I've changed the xbmc.python to 2.24.0. Is there anything else that should block this?

learningit:
I think it's OK. I really haven't looked through everything because there are so many files. I'm actually curious as to why so many?
I assume the lib stuff is from an OEM  - there's some stuff there that we discourage like using \__varnames\__ but we understand the maintenance issue. 
@phil65 may have some issues.
If nothing is pointed out in a day, I'll merge it.

ruuk:
@learningit: I went looking and one of the libs I had included was in the repo, so I dumped it and added it  as a dependency. Now it's only 150 files :)

learningit:
Well I guess it was worth asking, 747 to 150 is pretty good. :)

ruuk:
Went ahead and removed the remaining included libs after some more searching on the repo.

script.tablo/lib/windows/guide.py
```diff
@@ -0,0 +1,1030 @@
...
+    AIRINGS_LIST_ID = 300
+
+    SCHEDULE_GROUP_ID = 500
+    SCHEDULE_BUTTON_TOP_ID = 501
+    SCHEDULE_BUTTON_BOT_ID = 502
+
+    sectionAction = 'Schedule...'
+
+    EMPTY_MESSAGES = {
+        'SERIES': ('No Episodes', 'There are no upcoming episodes for this show.'),
```
learningit: Do these belong in strings.po?

ruuk: Strings moved to strings.po.

Shouldn't be any sleeps longer than that.

ruuk:
@learningit: Thanks for taking the time to look at this. I've made the last changes you mentioned.

script.tablo/addon.xml
```diff
@@ -0,0 +1,27 @@
...
+    </extension>
+    <extension point="xbmc.addon.metadata">
+        <summary lang="en">Tablo TV - Live TV and DVR</summary>
+        <description lang="en">Tablo is a DVR that lets you watch, pause and record free OTA TV with Kodi. Enjoy live sports, local news and your favorite network broadcast TV shows in HD through the Tablo add-on. No cable or satellite subscription required. Additional hardware purchase required. For more information visit www.TabloTV.com.
+</description>
+        <disclaimer lang="en">Requires a Tablo device</disclaimer>
+        <license>Apache License 2.0</license>
+        <forum>https://community.tablotv.com</forum>
+        <website>https://www.tablotv.com</website>
+        <source></source>
```
learningit: On my final look, realized that <source> needs to be filled out. This is my last comment, once fixed I'll merge.

script.tablo/lib/util.py
```diff
@@ -0,0 +1,508 @@
...
+            w.show()
+            return func(*args, **kwargs)
+        finally:
+            if w:
+                w.close()
+            del w
+    return inner
+
+
+def withLaodingDialog(method, *args, **kwargs):
```
phil65: typo (laoding vs. loading)

script.tablo/lib/windows/livetv.py
```diff
@@ -0,0 +1,936 @@
...
+
+    BASE_XML_PATH = os.path.join(
+        SKIN_PATH, 'resources', 'skins', 'Main', '720p', 'script-tablo-livetv.xml'
+    )
+
+    OUTPUT_PATH = os.path.join(
+        SKIN_PATH, 'resources', 'skins', 'Main', '720p', 'script-tablo-livetv-generated.xml'
+    )
+
+    CHANNEL_BASE_XML = '''
```
phil65: Hmm not really a big fan of creating XML files like this.+
Perhaps better move this stuff out of the .py files into separate xml files and parsing them with a proper XML parser?

phil65:
Looks good in general I would say.

script.tablo/lib/util.py
ruuk: Turns out it's not even used :) I probably was using and dropped it. I'll remove this for the _next_ version if it's all the same.

script.tablo/lib/windows/livetv.py
ruuk: I'm not a "fan" either normally, but for something simple like this where I'm more concerned about speed, I didn't really want all the overhead of a parser.

script.tablo/addon.xml
ruuk: Is the <source> tag mandatory, or can I just remove it?

ruuk:
I removed the empty ```<source>``` tag and went ahead and removed the unused and misspelled function. Is it good to merge?

learningit:
No. You must fill out the \<source\> field.

ruuk:
Ok, thanks. I'll get them to set something up that I can point this at.

ruuk:
Ok. I've added the ```<source>``` tag with the GitHub repo.
Thanks for your help @learningit and @phil65.

learningit:
I don't see the \<source\> tag in addon.xml

ruuk:
Hmmm. Neither do I. One second.

ruuk:
Ok. It's there now.

PR #183, opened 2016-06-01
[script.colorbox] 1.0.11

BADMS:
Actually please commit this version as new version will break skins. So major overhaul will be next version! Sorry, long day!

### DIFF
--- a/script.colorbox/default.py
+++ b/script.colorbox/default.py
@@ -30,6 +30,15 @@ class Main:
             self.image_now = xbmc.getInfoLabel("Player.Art(thumb)")
             self.image_now_fa = xbmc.getInfoLabel("MusicPlayer.Property(Fanart_Image)")
             self.image_now_cfa = xbmc.getInfoLabel("ListItem.Art(fanart)")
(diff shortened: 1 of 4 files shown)


script.colorbox/default.py
```diff
@@ -30,6 +30,15 @@ def __init__(self):
             self.image_now = xbmc.getInfoLabel("Player.Art(thumb)")
             self.image_now_fa = xbmc.getInfoLabel("MusicPlayer.Property(Fanart_Image)")
             self.image_now_cfa = xbmc.getInfoLabel("ListItem.Art(fanart)")
+            self.image_now_cpa = xbmc.getInfoLabel("ListItem.Art(poster)")
+            if not HOME.getProperty("cpa_ignore_set") and self.image_now_cpa != self.image_prev_cpa:
+                try:
```
phil65: so this script creates images for ANY focused item, but doesnt do any garbage collection?

BADMS:
Garbage collection? English or expand..
As for any focused item. NO. Only specified item as per previous submissions.
Is there a problem?

BADMS:
To be honest, this isn't needed for jarvis.
Next update will be for Krypton as that is what the skin this is helper for us using.

PR #184, opened 2016-06-03
[script.pinsentry] 2.0.0

robwebset:
Please can I have the following version of pinsentry merged. It is to get it looking better in Jarvis, the existing version is in Helix (Which is still valid for versions before Jarvis).

The changes (for easy diff viewing) are:

https://github.com/robwebset/script.pinsentry/commit/0a478b4546f8b105f60fb7000fff05d666a9d8c7

Thanks, Rob


robwebset:
Thanks, Rob

PR #188, opened 2016-06-07
[script.service.kodi.callbacks] 1.0.0

KenV99:
This is a request to update with the following changes:
v1.0.0 (2016-05-29)
Better po file formatting
Fixes many bugs with unicode strings
Fixes bug with mp3 files with no tags
Fixes bug in reading windowsids from settings
Improves onShutdown task


script.service.kodi.callbacks/resources/language/English/strings.po
```diff
@@ -17,7 +20,7 @@ msgid "Callbacks for Kodi"
 msgstr ""
 
 msgctxt "Addon Description"
-msgid "Provides user definable actions for specific events within Kodi. Credit to Yesudeep Mangalapilly (gorakhargosh on github) and contributors for watchdog and pathtools modules."
+msgid "Provides user definable actions for specific events within Kodi. Credit to Yesudeep Mangalapilly (gorakhargosh on github) and contributors for watchdog and pathtools modules.    "
```
phil65: why adding spaces here?

script.service.kodi.callbacks/resources/language/English/strings.po
```diff
@@ -123,11 +126,14 @@ msgid "Command for Task %s, Event %s completed succesfully!"
 msgstr ""
 
 msgctxt "#32025"
-msgid "\nThe following message was returned: %s"
+msgid "\n"
+"The following message was returned: %s"
```
phil65: Is this valid gettext? looks weird to me, but I might be wrong.

script.service.kodi.callbacks/resources/lib/dialogtb.py
```diff
@@ -61,22 +61,22 @@ def onFocus(self, controlID):
...
-            nmsg = '\n'.join(omsg) + '\n'
+                l1.append(u'i=%s, len=%s' % (i1, len(i1)))
+            nmsg = u'\n'.join(omsg) + u'\n'
             tmsg.append(nmsg)
-        wmsg = ''.join(tmsg)
+        wmsg = u''.join(tmsg)
     else:
-        wmsg = ''
+        wmsg = u''
     msgbox.set_text(title, wmsg)
```
phil65: might be easier to use
`from __future__ import unicode_literals`

script.service.kodi.callbacks/resources/lib/publishers/log.py
```diff
@@ -39,7 +40,7 @@ def __init__(self, interval=100):
         self.interval = interval
 
     def run(self):
-        f = open(self.logfn, 'r')
+        f = codecs.open(self.logfn, 'r', encoding='utf-8', errors='ignore')
```
phil65: please use "with" statement when opening files.

script.service.kodi.callbacks/resources/lib/utils/poutil.py
```diff
@@ -304,38 +608,115 @@ def getFileList(self):
...
+            found, strid = self.podict.has_msgid(s.decode('unicode_escape'))
             if found is False:
                 self.podict.addentry(strid, s)
         self.podict.write_to_file(self.current_working_English_strings_po)
 
-kodipo = KodiPo()
-_ = kodipo.getLocalizedString
-__ = kodipo.getLocalizedStringId
+
+def getPlatform():
```
phil65: might be easier to check system.platform.x infolabel?
see http://kodi.wiki/view/List_of_boolean_conditions

script.service.kodi.callbacks/resources/lib/utils/poutil.py
```diff
@@ -304,38 +608,115 @@ def getFileList(self):
...
+                    findstr = ur"%s\('(.+?)'\)" % finds[0]
                     find = re.compile(findstr)
-                    finds = []
                     try:
                         finds = find.findall(lines)
                     except re.error:
-                        pass
+                        finds = []
                     lstrings += finds
         return lstrings
```
phil65: Not sure what this is used for, but might be worth havin a look at https://pypi.python.org/pypi/polib

script.service.kodi.callbacks/resources/language/English/strings.po
KenV99: See: [https://www.gnu.org/software/gettext/manual/html_node/PO-Files.html](url) at end of page.
My understanding is that it is easier for translators.

script.service.kodi.callbacks/resources/language/English/strings.po
KenV99: I'll fix on next pull request (if I need to squash before this is 'commitable' it will be on this pull, otherwise next revision).

script.service.kodi.callbacks/resources/lib/dialogtb.py
KenV99: Didn't realize that was included in the kodi python distro on all systems.
Will look into it for next version.

script.service.kodi.callbacks/resources/lib/utils/poutil.py
KenV99: Yes - I run a 'build' routine outside of Kodi that utilizes this module. Still good for me to know about an alternative way to get that info.

script.service.kodi.callbacks/resources/lib/utils/poutil.py
KenV99: The whole module has three purposes:
1) Allow me to write gettext _style_ inline string translation using the original English text instead of the kodi numbers (I couldn't figure out what my own code was doing without looking up the number in another file - painful in a largish project).
2) It scans the files in the directory tree looking for other modules where the proper import to use the module is present. If it is, it scans the file for translatable strings and if they aren't in the pofile, they are added.
3) It also allows me to use routines when I am generating my settings.xml file to also add those strings to the po file.

polib looks interesting for help with reading and writing the file, perhaps I'll look into it for the next version, but the current code seems to be working pretty well from what I can tell.

script.service.kodi.callbacks/resources/lib/publishers/log.py
KenV99: OK, do I need to fix this before this pull can be merged? Or can it wait until the next revision?

script.service.kodi.callbacks/resources/lib/publishers/log.py
```diff
@@ -39,19 +40,19 @@ def __init__(self, interval=100):
         self.interval = interval
 
     def run(self):
-        f = open(self.logfn, 'r')
-        f.seek(0, 2)           # Seek @ EOF
-        fsize_old = f.tell()
-        while not self.__abort_evt.is_set():
+        with codecs.open(self.logfn, 'r', encoding='utf-8', errors='ignore') as f:
```
KenV99: Corrected as requested by @phil65

script.service.kodi.callbacks/resources/language/English/strings.po
KenV99: Corrected as requested by @phil65

script.service.kodi.callbacks/resources/lib/publishers/log.py
KenV99: Corrected as requested by @phil65

KenV99:
Corrections made. Suggestions will be examined for next actual revision. Incorporating 'unicode_literals' will require a major rewrite of many files. Unclear to me if that will be worthwhile.

Please let me know if there are any other needed corrections in order to have this merged. I know a few users from the forum are awaiting the bug fixes. Thanks.

phil65:
unicode_literals was just a quick suggestion bc of that excessive prefixing. If you dont think it's worth it, then just ignore. :)

KenV99:
If I had thought about it from the beginning I definitely would have used it and it would have saved many headaches. I actually will definitely use it in all modules in future projects. 

- Ken

PR #190, opened 2016-06-09
[script.module.lamlib] 1.0.0

lambda81:
helper module for various addons
https://github.com/xbmc/repo-plugins/pull/523
https://github.com/xbmc/repo-plugins/pull/524


script.module.lamlib/lib/lamlib/youtube.py
```diff
@@ -0,0 +1,185 @@
...
+                image = image.encode('utf-8')
+
+                append = {'title': title, 'url': url, 'image': image}
+                if not next == '': append['next'] = next
+                self.list.append(append)
+            except:
+                pass
+
+        try:
+            u = [range(0, len(self.list))[i:i+50] for i in range(len(range(0, len(self.list))))[::50]]
```
phil65: range(len(range(len()))) ? :)

phil65:
That was fast :)
Looks good to me, mind takin another look, @learningit or someone else?

lambda81:
hi phil
i am doing this to split to 50 items for the addons i want to use pagination
there is a greek network, STAR that has uploaded many tv shows to their official youtube channel
in this addon i want to get all videos = episodes
perfect for starting all episdes with "play from here":)

lambda81:
sure phil, no rush to be accepted

lambda81:
i forgot:)
i am also trying to get the duration of videos in parallel
this may doesn't make sense
but i need it because this network, STAR, also uploads trailers of new episodes
i am later filtering these videos based on duration, and keep only full episodes

learningit:
I'm OK with it.

PR #191, opened 2016-06-11
[screensaver.random] 1.2.0

robwebset:
Please can I have the following merged.

Changes:
- Add option to schedule different screensavers at different times

Thanks, Rob


phil65:
looks good to me.
Why are your diffs always so polluted though? Do you change line endings back and forth?

robwebset:
@phil65 It's something I'm changing to LF when each addon changes, it was requested here:

https://github.com/xbmc/repo-scripts/pull/122#issuecomment-216025273

I am just changing them as I update each addon - rather than doing a mass of updates in one go.

phil65:
ok, thx for info.

robwebset:
Thanks, Rob

PR #194, opened 2016-06-18
[script.myepisodes] 1.2.7

maximeh:


script.myepisodes/addon.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<addon id="script.myepisodes"
+       name="MyEpisodes"
+       version="1.2.7"
+       provider-name="Maxime Hadjinlian">
+  <requires>
+    <import addon="xbmc.python" version="2.14.0"/>
```
learningit: xbmc.python should be set to 2.24.0 for Jarvis.

script.myepisodes/default.py
```diff
@@ -0,0 +1,213 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+import os
+import sys
+import threading
+import xbmc
+import xbmcaddon
+
+__addon__         = xbmcaddon.Addon()
```
learningit: in the future please do not use \__varname\__ type labels. These are reserved for python internals.

PR #195, opened 2016-06-18
[service.subtitles.subztv.gr] 1.0.0

lambda81:
subtitle service for
http://subztv.gr/


learningit:
Please change icon and fanart to solid backgrounds.

service.subtitles.subztv.gr/resources/lib/subztvgr.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+    def download(self, path, url):
+        try:
+            result = client.request(url)
+
+            f = os.path.splitext(urlparse.urlparse(url).path)[1][1:]
+            f = os.path.join(path, 'file.%s' % f)
+
+            file = open(f, 'wb') ; file.write(result) ; file.close()
```
learningit: Please use "with" for file control.

service.subtitles.subztv.gr/resources/lib/subztvgr.py
```diff
@@ -0,0 +1,168 @@
...
+                    if len(files) > 1: break
+                    time.sleep(1)
+                except:
+                    pass
+
+            control.deleteFile(f)
+
+            subtitle = [i for i in files if any(i.endswith(x) for x in ['.srt', '.sub'])][0]
+
+            try: subtitle = os.path.join(path, subtitle)
```
learningit: rather than a try-except, couldn't you just do a path.decode('utf-8') ?

service.subtitles.subztv.gr/service.py
```diff
@@ -0,0 +1,153 @@
...
+
+        threads.append(workers.Thread(self.subztvgr))
+
+        [i.start() for i in threads]
+
+        for i in range(0, 10 * 2):
+            try:
+                is_alive = [x.is_alive() for x in threads]
+                if all(x == False for x in is_alive): break
+                time.sleep(0.5)
```
learningit: please use xbmc.sleep() in place of time.sleep().

service.subtitles.subztv.gr/resources/lib/subztvgr.py
```diff
@@ -0,0 +1,168 @@
...
+
+            if len(files) == 0: return
+
+            control.execute('Extract("%s","%s")' % (f, path))
+
+            for i in range(0, 10):
+                try:
+                    dirs, files = control.listDir(path)
+                    if len(files) > 1: break
+                    time.sleep(1)
```
learningit: Please use xbmc.sleep() in lace of time.sleep and do abort test in the loop.

service.subtitles.subztv.gr/resources/lib/subztvgr.py
lambda81: Added it in my new PR
https://github.com/lambda81/service.subtitles.subztv.gr/blob/master/resources/lib/subztvgr.py#L141

service.subtitles.subztv.gr/resources/lib/subztvgr.py
lambda81: There is a greek subtitle team that uses on it's naming acute accents (polytonic greek)
https://en.wikipedia.org/wiki/Greek_diacritics
I tried with decode('utf-8'), but python can't handle it since it's not a part of modern greek
The only easy way i found for python to handle files with these accents and save them was this
utf-8 failed all the time

service.subtitles.subztv.gr/service.py
lambda81: Added it in my new PR
https://github.com/lambda81/service.subtitles.subztv.gr/blob/master/service.py#L77

service.subtitles.subztv.gr/resources/lib/subztvgr.py
lambda81: Added it in my new PR
https://github.com/lambda81/service.subtitles.subztv.gr/blob/master/resources/lib/subztvgr.py#L155

PR #197, opened 2016-06-19
[script.module.kodiswift] 0.0.1

Sinap:


script.module.kodiswift/lib/kodiswift/constants.py
```diff
@@ -0,0 +1,90 @@
...
+        """Returns the sort method specified. sort_method is case insensitive.
+        Will raise an AttributeError if the provided sort_method does not
+        exist.
+
+        >>> SortMethod.from_string('title')
+        """
+        return getattr(cls, sort_method.upper())
+
+# View mode ids pulled from skins
+VIEW_MODES = {
```
phil65: This viewmode stuff is too error-prone, please get rid of it.

script.module.kodiswift/lib/kodiswift/xbmcmixin.py
```diff
@@ -0,0 +1,532 @@
...
+        """
+        # This method is named with an underscore to suggest that callers pass
+        # the subtitles argument to set_resolved_url instead of calling this
+        # method directly. This is to ensure a video is played before calling
+        # this method.
+        player = xbmc.Player()
+        for _ in range(30):
+            if player.isPlaying():
+                break
+            time.sleep(1)
```
phil65: please check if an abort is requested, see http://kodi.wiki/view/Service%20addons

phil65:
Very cool, thx for updating this. :)

Sinap:
Removed the view mode stuff and updated the playlist code to use xbmc.Monitor

script.module.kodiswift/addon.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<addon id="script.module.kodiswift" name="kodiswift" provider-name="Aaron Frase (Sinap)" version="0.0.1">
+  <requires>
+    <import addon="xbmc.python" version="2.24.0" />
+  </requires>
+  <extension library="lib" point="xbmc.python.module" />
+  <extension point="xbmc.addon.metadata">
```
phil65: there might be some metadata tags missing here? @learningit , you know this part much better than me. :)

MartijnKaijser:
How many of the plugin wrappers need there be in repo? Seriously why?

script.module.kodiswift/lib/kodiswift/constants.py
```diff
@@ -0,0 +1,73 @@
...
+    TITLE = 9
+    TITLE_IGNORE_THE = 10
+    TRACKNUM = 7
+    UNSORTED = 37
+    VIDEO_RATING = 18
+    VIDEO_RUNTIME = 29
+    VIDEO_SORT_TITLE = 24
+    VIDEO_SORT_TITLE_IGNORE_THE = 25
+    VIDEO_TITLE = 23
+    VIDEO_YEAR = 17
```
MartijnKaijser: Never ever use the numbered values for sorting. They are NOT meant to be used, ever.

MartijnKaijser: That same holds for every numbered value you find at the bottom of every pydocs for Kodi.

phil65: Sorry martijn, I am not sure if you know what youre talkin about.
KodiSwift was one of the best frameworks around, and a good framework reduces the workload to develop new addons by a lot.
By putting the values for sorting here, you will get them for AutoCompletion / DocStrings in Python IDEs.

MartijnKaijser: Those values are not fixed and can break at whatever kodi version that comes out. Only strings should be used for sorting.

Repo is already filled with this stuff cause each dev creates his own which i totally hate. xbmcswift 1&2 is also still in repo. Also wth is it called kodiswift? Can't they think of something better. Remember KOdi is a trademark

phil65: What strings are you talkin about? There are no strings related to this anywhere. Nowhere.
I rather have some more module add-ons than having to review twice as much code (or even more). I even urge addon devs to create frameworks when they do lot of different add-ons. That also helps to properly separate your code, so you dont run into danger to have circular imports for example when developing more complex add-ons. If you hate something, give a proper reason why. Havin some module add-ons do not hurt at all.
You really put up trademark concerns? Then I might also drop my own framework (kodi65) as well, and some other add-on devs would have to do that as well.
Your second point might be valid. @Sinap, do you have an add-on in the pipeline which makes use of this?
If this is the way we review add-ons, then I wil step away from it. Someone puts up some great work and then has to listen to this....

Sinap: He might be referring to using `xbmcplugin.SORT_METHOD_ALBUM` instead of using the integer value of that variable. I should probably change it to `ALBUM = xbmcplugin.SORT_METHOD_ALBUM` etc. in case the values change.

@phil65 I forked the xbmcswift2 project because the original author hasn't made a commit since 2013 and the project appears to be dead. I was working on this for my own projects but recently 2 other add-on developers asked when I would have it on the official repo so they could use it.

phil65: yup, agreed, that would be better. (see https://github.com/phil65/script.module.kodi65/blob/master/lib/kodi65/itemlist.py#L10 ;) )
As said, I'm totally +1 for this module (I think I even recommended you to update it?)

moedje:
Is there any update on when Kodiswift might be merged into the main REPO I've started building addon's against this framework instead of the old XBMCSWIFT2 and with Sinap's latest Hotfixes to 0.0.3 it would be great to have this in the main Repo to encourage a migration away from XBMCSWIFT2 and easy distribution of Addons now being built against this fantastic update! Thanks

PanderMusubi:
@MartijnKaijser xbmcswift2 will be replaced by kodiswift as the xbmcswift2 is no longer actively maintained. So this falls in the aim of Kodi only to ship maintained code. At the moment, several plugins are waiting until kodiswift becomes available before they will update and say goodbey to the unmaintained xbmcswift2. Much obliged if this PR moves forward.

PR #198, opened 2016-06-21
Krypton updates

phil65:


phil65:
Trivial update, will pull since it is needed bc of changelog core changes.

PR #200, opened 2016-06-22
[script.recap] 1.0.1

robwebset:
Please can I have the following updated. It is a set of changes agreed with the author of recapguide.com to help reduce the load on their web server.


robwebset:
Thanks, Rob

PR #201, opened 2016-06-22
[service.subtitles.subscene] 1.7.1

manacker:
Fix HTTP proxy issue, check download errors
https://github.com/manacker/service.subtitles.subscene/pull/3


learningit:
In your next update, please remove the use of \__varname\__ stype labels. These are reserved for python internals.

thiagofigueiro:
@learningit it would do well to update the wiki to replace the magic names int he examples so newcomers don't do it: https://kodi.wiki/view/GUI_tutorial

```python
import xbmcaddon
import xbmcgui

__addon__ = xbmcaddon.Addon()
__addonname__ = __addon__.getAddonInfo('name')
```

PR #206, opened 2016-07-04
[service.subtitles.subtitles.gr] 1.2.0

lambda81:
Updated version of the subtitles.gr service


service.subtitles.subtitles.gr/resources/lib/subztvgr.py
```diff
@@ -0,0 +1,170 @@
...
+                    if len(files) > 1: break
+                    control.sleep(1000)
+                except:
+                    pass
+
+            control.deleteFile(f)
+
+            subtitle = [i for i in files if any(i.endswith(x) for x in ['.srt', '.sub'])][0]
+
+            try: subtitle = os.path.join(path, subtitle)
```
learningit: I'm guessing that you get an error when there's unicode in the path or file name. You should try .decode('utf-8')

service.subtitles.subtitles.gr/service.py
```diff
@@ -0,0 +1,191 @@
...
+
+        threads.append(workers.Thread(self.subtitlesgr))
+
+        [i.start() for i in threads]
+
+        for i in range(0, 10 * 2):
+            try:
+                is_alive = [x.is_alive() for x in threads]
+                if all(x == False for x in is_alive): break
+                control.sleep(500)
```
learningit: You should check for an abort here.

service.subtitles.subtitles.gr/resources/lib/subztvgr.py
```diff
@@ -0,0 +1,170 @@
...
+                result = re.sub(r'[^\x00-\x7F]+', ' ', result)
+
+                url = client.parseDOM(result, 'div', attrs = {'id': 'movies'})[0]
+                url = re.findall('(/movies/\d+)', url)
+                url = [x for y,x in enumerate(url) if x not in url[:y]]
+                url = [urlparse.urljoin('http://subztv.gr', i) for i in url]
+                url = url[:3]
+
+                for i in url:
+                    try:
```
learningit: What is with all the nested try-except stuff? You're probably not solving a problem with these, but rather masking one.

learningit:
You need to change the background of the icon from transparent to solid.

service.subtitles.subtitles.gr/resources/lib/subztvgr.py
lambda81: I updated the PR
decode('utf-8') was failing with some characters when i was decoding after the join of the path.
I am decoding now only the srt filename and now it works all the time.

service.subtitles.subtitles.gr/service.py
lambda81: Added

service.subtitles.subtitles.gr/resources/lib/subztvgr.py
lambda81: I removed it.

PR #208, opened 2016-07-12
[script.module.pycaption] 0.5.5

emilsvennesson:
I think this module will be useful for a lot of video plugins.
https://github.com/pbs/pycaption


learningit:
Are there any addons using this module? We only include modules which addons make use of.

emilsvennesson:
I'm using it in my plugin (https://github.com/emilsvennesson/kodi-viaplay) which I am planning to submit to the official repo once it's reasonably stable.

learningit:
OK. We'll hold off including until then. Frankly I've found the module to be overkill for what needs to be done on the actual PBS content.

emilsvennesson:
Fair enough. Should I just reference to this pull request once I've submitted my addon?

learningit:
Yes. I would encourage you to think about writing 5 lines of code to eliminate the module, assuming you're doing a .vtt to .srt translate.

emilsvennesson:
Unfortunately, I have to do a .sami to .srt which can be a pain in the ass. :/

learningit:
I thought .smi files were supported by the player, at least they used to be...

emilsvennesson:
And this just reminded me why you should always read the documentation... :see_no_evil:

PR #209, opened 2016-07-12
[script.module.cssutils] 1.0

emilsvennesson:
This is needed for #208.


learningit:
Are there any addons in the repo which make use of this module? We only include modules which are actually used by addons in the repo.

PR #211, opened 2016-07-15
[script.matchcenter] 2.0.0

enen92:
This a request to add script.matchcenter to the Krypton repository. This is a script which the main goal is to bring extended information for football/soccer matches (official thread on the forum: http://forum.kodi.tv/showthread.php?tid=259894)

Requires script.module.twitter which was sent through this PR: https://github.com/xbmc/repo-scripts/pull/210

Ping @zag2me as he owns the website (thesportsdb.com)

Many thanks


script.matchcenter/resources/lib/utilities/keymapeditor.py
```diff
@@ -0,0 +1,168 @@
...
+
+### Key mapp            
+def run():
+    ## load mappings ##
+    try:
+        setup_keymap_folder()
+    except Exception:
+        pass
+
+    defaultkeymap = read_keymap(default)
```
learningit: I'm not sure if it's OK to modify keymaps on the fly from an add-on.  Is this just loading the system keymap and then creating a local copy for the add-on or is it actually modifying the system keymap? (easier to ask than follow all the code).

enen92: It's modifying the system keymap. The code is basically the code used by takoi on the keymap editor addon that is on the official repository with some modifications. It creates a global keymap (modifying the keymaps.xml) that points to the addon. I don't know either if this is allowed so please check with some other team member. If not acceptable I can strip this functionality from the addon and simply add a wiki entry later that explains how to set a keymap using the keymap editor addon.

enen92: Any news @learningit ?

Cheers

learningit: Asked for an opinion from others, I really don't know the answer here. I'll keep an eye on this until we get an answer.

tamland:
````
except Exception:
    pass
````
I would never write such codes. this cant be accurately copied:) 

Seriously though, the problem with writing to the keymap file like this is that permanently overrides a key (including after uninstalling the addon) without users realizing it and with no obvious way to fix it. And they'll come to the forums as why key x doesn't work.

Why not just run the real keymap editor from here? It will be a few more steps for users but it'll at least be clearer what is going on and they'll have the option to go back to default.

learningit:
Sorry it took a while to come back. I've asked around and the long and short of it is that keymaps are meant to be for devices mapped to keys, not for use by an add-on to change mappings.  @tamland 's suggestion above seems to be the correct way to go.

enen92:
I understand perfectly the problem. It seems too logical to not have though it could be a problem while building the addon. I just have been without much time and also in places with really bad internet connection. I have also had a look at the keymap editor addon and it seems there is no way to call it with the method I want it to trigger. So probably the best way to deal with it is just to present a text dialog with instructions and to just run the keymap editor addon if the user feels like adding a shortcut to the addon. At least users will know what they are doing.

enen92:
@learningit  it should be by the book now. I found some exceptions on the code so I'll commit some fixes and squash all the commits later.
Probably it is a good idea to add the "keymap constraint" to http://kodi.wiki/view/Add-on_rules

Cheers

enen92:
@learningit I am happy with the fixes and also squashed all the commits.
Cheers

PR #214, opened 2016-07-25
Script.skin.helper.service 1.0.100

marcelveldt:
One more update of the skinhelper addon. I'm still working on the rewrite (into seperated chunks like agreed). Thanks in advance for reviewing and updating.


marcelveldt:
?? why on earth do I also get all those other commits into my PR ? Must have done something wrong with the branching I guess. I'll close this one and repeat it once I sorted it out

PR #215, opened 2016-07-26
[script.skin.helper.service] 1.0.100

marcelveldt:
@phil65 sorry still working on the rewrite of the addon into several parts so one more update for the repo. Thanks in advance for updating and enjoy the holidays.

BTW: Is it fine if I will submit the new version of this addon (which we discussed) only to Krypton and keep this one as legacy version with minor bugfixing only ?


phil65:
It is fine to push the newer versions to krypton only and let the older versions die a slow death.
Please keep in mind that aim no. 1 is not to create a library module just for the heck of it, but to separate the different extension points so that skinners who for example just want a color selector dont need to automatically also have some service active.

marcelveldt:
@phil65 yeah offcourse all clear. Also the library modules can be used in other addons.

marcelveldt:
Any news on this ?

learningit:
I haven't looked at this as @phil65 was commenting. I'll try to see if he has anything else and/or have a look over the weekend.

PR #216, opened 2016-07-26
[script.skinshortcuts] 1.0.10

marcelveldt:
new stable version of skinshortcuts, thanks in advance for updating!

### DIFF
--- a/script.skinshortcuts/addon.xml
+++ b/script.skinshortcuts/addon.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<addon id="script.skinshortcuts" name="Skin Shortcuts" version="1.0.8" provider-name="BigNoid,marcelveldt">
+<addon id="script.skinshortcuts" name="Skin Shortcuts" version="1.0.10" provider-name="BigNoid,marcelveldt">
 	<requires>
 		<import addon="xbmc.python" version="2.20.0"/>
 		<import addon="script.module.simplejson" version="3.3.0"/>
@@ -7,6 +7,7 @@
 	</requires>
 	<extension point="xbmc.python.library" library="default.py"/>
 	<extension point="xbmc.addon.metadata">
+		<news>New features for skinners including context menu in management dialog, icon override and build improvements. New features for users including ability to disable a menu item and Krypton fixes.</news>
 		<description lang="bg">Добавка за облици, която осигурява просто управление и списък с потребителски преки пътища (изисква поддръжка от облика)</description>
 		<description lang="de">Add-on für Skins für einfache Bearbeitung und Auflistung von benutzerdefinierten Verknüpfungen (erfordert Skin-Unterstützung)</description>
 		<description lang="el">Πρόσθετο για κελύφη, το οποίο παρέχει έναν απλό τρόπο διαχείρισης και κατηγοριοποίησης των συντομεύσεων του χρήστη (πρέπει να υποστηρίζεται από το κέλυφος)</description>
@@ -37,4 +38,8 @@
 		<email></email>
 		<source>https://github.com/BigNoid/script.skinshortcuts</source>
 	</extension>
+	<assets>
+		<icon>icon.png</icon>
+		<fanart>fanart.jpg</fanart>
+	</assets>
 </addon>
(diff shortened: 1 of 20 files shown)


script.skinshortcuts/default.py
```diff
@@ -307,11 +304,22 @@ def _manage_shortcuts( self, group, defaultGroup, nolabels, groupname ):
         if homeWindow.getProperty( "skinshortcuts-loading" ) and int( calendar.timegm( gmtime() ) ) - int( homeWindow.getProperty( "skinshortcuts-loading" ) ) <= 5:
             return
 
+        # Get current Super Favourites context menu status
+        sfMenu = None
+        if xbmc.getCondVisibility( "System.HasAddon(plugin.program.super.favourites)" ):
```
learningit: I'm not sure if it's OK to change another addon's settings, at least not without first alerting the user and asking for permission from him/her.  Doesn't feel right.

Ignoble61: The setting is only changed whilst the scripts dialog is open - it's put back to the users choice afterwards.

It's done - [at Spoysers suggestion](http://forum.kodi.tv/showthread.php?tid=192662&pid=2353547#pid2353547) - because when this setting is enabled in super favourites attempts by the user to trigger the context menu don't trigger the normal id's that can be checked for in onAction. Instead, Super Favourites own context menu shows - which also closes this scripts window.

Ignoble61: From a UX perspective, explaining to the user why I want to disable something that can't be used and that they won't notice isn't enabled and asking for their permission to do so - in the three lines we're given and with the default option being 'No' - isn't something I consider acceptable from a UX perspective.

There are alternatives (such as editing the keymap so that the Skin Shortcuts window isn't affected when the SF global context menu is active) that I can look into, but can I get confirmation that this is considered to be blocking the update on the repo before I do?

marcelveldt:
Any news about this ?
Looks like the repo mirror currently is failing on skinshortcuts as there is no actual zip file available, causing a lot of skin failing to install:

http://mirrors.kodi.tv/addons/jarvis/script.skinshortcuts/

tamland:
> causing a lot of skin failing to install:

You're imagining things;) Works fine: http://mirrors.kodi.tv/addons/jarvis/script.skinshortcuts/script.skinshortcuts-1.0.8.zip It's just a bug in the mirror web page.

marcelveldt:
haha, okay great. Thanks!

learningit:
I've discussed this and it violates the thou shalt not change another add-on's settings or files.  A workaround would be to have the other add-on create a function which would manage the on/off of the setting. You could then call it through xbmc.runplugin or runscript with the correct parameter.  That allows the addon's author to perform whatever checking or cleanup is necessary to ensure a known state.

marcelveldt:
Makes a lot of sense to me. @Ignoble61 I'm not familiar with the use case of this particular piece of code but I do have some time to fix it if you want. My suggestion would be to remove this from the code for now untill a proper function is available in the SF script.

Ignoble61:
@marcelveldt that would be appreciated as I won't have time to chase this up for a week or yet :)

marcelveldt:
@learningit all set now

script.skinshortcuts/addon.xml
```diff
@@ -37,4 +38,8 @@
 		<email></email>
 		<source>https://github.com/BigNoid/script.skinshortcuts</source>
 	</extension>
+	<assets>
+		<icon>icon.png</icon>
+		<fanart>fanart.jpg</fanart>
+	</assets>
```
ronie: this probably slipped in :-)
assets tag should be part "xbmc.addon.metadata" extension point,
and will only work for kodi krypton and up.

```15:23:35 T:6144569344   DEBUG: ADDON: cpluff: 'Suspicious plug-in descriptor content in /var/mobile/Containers/Data/Application/69598EFF-C2DD-4D78-8128-64E076F4A72D/Library/Caches/home/addons/script.skinshortcuts/addon.xml, line 41, column 2 (ignoring unexpected element assets and its contents).'```

Ignoble61: Cheers Ronie. I've moved it to the xbmc.addon.metadata for 1.0.12 - https://github.com/xbmc/repo-scripts/pull/306. If you could confirm it's now in the right place that would be great.

ronie: it's in the right spot for sure, though if it was up to me, i'd simply remove it.
it's only supported on krypton and might still log an error on isengard/jarvis.

tamland: it was *designed* to be compatible with isengard/jarvis..

PR #217, opened 2016-07-29
[script.tvtunes] 6.3.5

robwebset:
Please can I have the following update merged, includes:

v6.3.5 (2016-07-29)
- Improve support when running on Python 2.6
- Support for Kodi v17

Thanks, Rob


robwebset:
Thanks, Rob

PR #218, opened 2016-07-30
[script.cinder] 1.0.2

jpr679:
Added "Run Cinder from here" context menu option.

### DIFF
--- a/script.cinder/addon.xml
+++ b/script.cinder/addon.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<addon id="script.cinder" name="Cinder" version="1.0.1" provider-name="Justin Rush">
+<addon id="script.cinder" name="Cinder" version="1.0.2" provider-name="Justin Rush">
 	<requires>
 		<import addon="xbmc.python" version="2.24.0"/>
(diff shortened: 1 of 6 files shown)


learningit:
Is this commit from the latest source? The version in addon.xml is 1.0.1, the PR is for version 1.0.2.

jpr679:
Forgot to bump up the version number.

learningit:
You need to squash your commits into a single commit.  Please see: https://github.com/ginatrapani/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit

jpr679:
My apologies I noticed a last minute string.po id collision and fixed it. Squashed the commits as well.

script.cinder/addon.xml
```diff
@@ -13,4 +13,12 @@
 		<email>cinder.kodi@gmail.com</email>
                 <source>https://github.com/jpr679/stript.cinder</source>
 	</extension>
+	<extension point="kodi.context.item">
+		<menu id="kodi.core.main">
+			<item library="scriptcontextmenuoption.py">
+				<label>32004</label>
+				<visible>true</visible>
```
tamland: This is not an acceptable condition. You need to narrow it down to something that makes sense for your addon, otherwise it's going to be shown for *every* single list item in kodi.

PR #221, opened 2016-08-07
[script.sonos] 2.0.0

robwebset:
Adds the following changes on top of v1.4.3

v2.0.0 (2016-08-07)
- Check for theaudiodb.com being set in Artist Slideshow
- Fix radio station name
- Add option to list and play Sonos Favourites
- When muting, mute all speakers in group
- When changing volume, change all speakers in group
- Add Sleep Timer Support
- Removed Text to Speech Support

Adding to Helix repo instead of Gotham (where is previously was) as it aligns to the  Artist Slideshow 2.0.0 which went into Helix.

Thanks, Rob


learningit:
Please squash your commits.

robwebset:
Hi @learningit 

Sorry about that - commits now squashed

Thanks, Rob

learningit:
Can you explain what you're doing with the keymap?  Are you replacing the system keymap?

robwebset:
Oh no, not replacing anything - just adding a set of mapping for the sonos controller when the sonos controller is displayed - and then removing it when the controller is closed. (That is the same as was in the previous version)

Thanks, Rob

learningit:
You may want to look at this discussion about keymaps:
 https://github.com/xbmc/repo-scripts/pull/211

robwebset:
Hi @learningit 
I think I'm covered for the concern raised on that thread - as soon as the keymap is copied in, I have a try-catch around the rest of the code to then ensure after that the keymap file is removed before the sonos controller window is shut.
Thanks, Rob

learningit:
That actually doesn't fix it for me. 
I'm of the opinion that the maps really aren't meant for add-ons to modify unless they are using custom Window IDs, which extend the mapping specifically for the add-on , rather than overwrite mappings and then go back and replace them.  The add-on should ask the user if they want the custom mapping and then copy it the keymaps directory.

Read http://kodi.wiki/view/Keymaps#Defaults carefully and you'll realize that your keymaps can be overwritten if the user has a keymap in higher alphabetical order than yours.

There are many ways that Kodi could fail leaving your overwritten keymaps in place and not trigger your try / except logic to put them back.

I'm not singling you out on this issue and I am by no means an expert on the subject, but I feel that whatever goes into the kodi.org repo should reflect a higher standard of coding which other users can use as examples to write their own add-ons correctly.  IMHO it's ok for an add-on to copy a custom keymap into the keymaps directory if they're using custom Windows IDs unique to their add-on, but paging a keymap in and out really just feels wrong.

robwebset:
Hi @learningit 

Unfortunately there is a case where the user (via settings) wants to have the keymap more global, so restricting to window Ids wouldn't cover that usecase.

I am aware they are loaded alphabetically - but for the current case this does work and it does avoid me having to attempt to update the default keymap - which I want to avoid. I am not aware of any other way of mapping keys other than using a keymap.

When you state about asking the user - by using the Sonos addon and then enabling things in the setting, they are actually confirming they want it :)

As I've highlighted previously - the sonos addon has been in the Gotham repo for 2 and a half years and was reviewed, discussed and passed at that time. I have not changed this behaviour in this set of changes - but instead just happen to push it to a later repo.

I hope this clarifies things to your satisfaction? (I have also have other addons that add keymaps - Addon:Sleep for example - this also needs global update and does it in the same way)

Thanks, Rob

learningit:
Unfortunately the above just tells me the history, not whether it's the correct thing to do or not. I've  asked other team members to weigh in on this.

learningit:
I've checked on this issue, and despite it being accepted initially, having a add-on modify keymap files is not what they were designed for.  They are meant to map devices to keys and not for add-ons to do mapping.  The more basic rule is that an add-on should not modify system files or another addon's files or settings.

robwebset:
OK, I think I'll have to pull all my addons into my own repo, this addon is pretty useless without being able to map remote control keys (as is the Sleep Addon). BTW, it doesn't modify keymap files - it just adds a new one.

Also not knowing if something you previously got approved is going to be later rejected by some-one else is just unmanagable.

I'll post another pull request to remove all my addons from the official repo once I've set up my own repo.

MartijnKaijser:
So why don't you use the context menu?  You hacking in the keymap will possibly destroy anything the user has added himself or another addon.
This may not have been an issue in the past but it will (or already is). There's a keymap editor so why don't you tell users to use that one?

Personally i would be pissed if some addon would screw up my kepmapping

robwebset:
It's a controller to control an external Sonos system - so needs to be able to support play, volume etc from a remote control - a context menu is no good.

Never had anybody complain about the keymap causing them issues on their system.

Either way - it doesn't matter any more.

learningit:
I'm sorry you feel that way. The correct way is ask the user on a first-time init to move the custom  keymap file to the keymaps dir. Leave it there, don't roll it in and out where a possibly inconsistent state my result. It may involve using unique Window IDs, I haven't looked that much at the issue.

We (both you and Kodi developers) all learn what works and what causes issues with the core system. I have also had to make changes constantly to my add-ons to keep them up to date with what is learned about or changed in the Kodi environment. This has sometimes involved my slowly fixing more than 50 add-ons due to those requirements.  You are not a new or junior contributor to the kodi.org repo.  Others will be using your add-ons as guides or reference for future new add-ons and we all have a responsibility to try to keep the add-ons in the kodi.org repo up to date as best we can.

MartijnKaijser:
So because we didn't need allow keymap changes anymore you remove every Addon that's not even affected by this. Good job

robwebset:
More a general case - if I'm maintaining my own repo - I may as well have them all in there.

Added to that, next time I submit the "Sleep" addon with bug fixes - it will no doubt hit the same issue ... and I'd guess there would be others.

Rob

PR #222, opened 2016-08-07
[Artist Slideshow] 2.0.2

pkscout:
v.2.0.2
- fixed crash when using fuzzy local path with non-Windows option
- updated German strings


Tadly:
Looking good, thanks

PR #226, opened 2016-08-18
[script.module.iso8601] 0.1.11

emilsvennesson:


learningit:
At first glance, there's an issue with a new submission in the Gotham repo. New submissions are for Jarvis and above. script.module.m3u8 is in Helix, so there may be a case for putting script.module.iso8601 there.  If script.module.m3u8 is the only thing using it, why not just add as a lib folder?
Haven't looked at anything else yet.

emilsvennesson:
@learningit: Oh, I didn't know about the submission issues. I expected this to work on all versions that weren't listed as deprecated (http://kodi.wiki/view/addon.xml) so that's why I submitted it to Gotham. I could re-do it and use the Helix branch instead if you want.

Wouldn't it make more sense to have iso8601 as a separate module? If we're still going to download the library, I think it'd be nice to let other addons use iso8601. I know I could use it in https://github.com/emilsvennesson/kodi-viaplay for example and possibly get rid of the heavy handed dateutil.parser approach.

learningit:
Let me look through this, I have limited time today.  I would propose that iso8601 be added to Helix to be in line with the m3u8 module, but I want to make sure there won't be compatibility/upgrade issues for existing m3u8 module users. Give me a day or two to get back. Thanks.

learningit:
BTW the repo submission matrix is in a different place: http://kodi.wiki/view/Submitting_Add-ons
Just to keep things confused.

emilsvennesson:
@learningit Of course! Take your time. :-)

learningit:
Finally got some time to test, the code seems fine and doesn't seem to have problems with other add-ons that use the m3u8 module. Please close this PR and re-submit into Helix. I will merge this and PR#226 into the Helix repo when that is done. Thanks.

emilsvennesson:
@learningit Great! Thanks a lot. The new PR is at #249. :-)

PR #228, opened 2016-08-19
[script.recap] 1.0.2

robwebset:
Change to the Recap addon to notify the user that it is being removed from the Official Repository.

I thought it best to notify the users that it has moved so they can update their installation cleanly. (Not everyone may be following the forum, so may not know)

Then in a few weeks I'll do a pull to remove the addon completely.

Hope this is OK
Rob


MartijnKaijser:
Since you remove all functional code this Addon is broken. So also set it broken in addon.xml with a description.

robwebset:
Killing this one off - will do another with the broken element

PR #236, opened 2016-08-20
[script.tvtunes] 6.3.7

robwebset:
Change to the TvTunes addon to notify the user that it is being removed from the Official Repository.

I thought it best to notify the users that it has moved so they can update their installation cleanly. (Not everyone may be following the forum, so may not know)

Then in a few weeks I'll do a pull to remove the addon completely.

Hope this is OK
Rob


MartijnKaijser:
Please make the popup a one time thing asap. It's hugely annoying it does this every boot.

robwebset:
Hi @MartijnKaijser 
It's supposed to be annoying :)

The user needs to either
- Switch to use the version in the robwebset repository
- Uninstall the addon if they do not want something that is not in the official repository

Thanks
Rob

PR #238, opened 2016-08-20
[script.pinsentry] Helix Delete

robwebset:
Please can you delete script.pinsentry from the Helix repository.

Please can you also delete all the files at the following location:
http://mirrors.kodi.tv/addons/helix/script.pinsentry/

Thank, Rob


robwebset:
Any reason this wasn't merged to delete the addons - the aim of this is to remove the old versions from the official repo - several of the addons do not conform to the new rules and should be removed.

PR #250, opened 2016-08-23
[script.kodi.loguploader] 0.0.1

ronie:
new logfile upload addon for kodi.

can someone with ios/osx knowledge verify if paths the where the addon looks for crashlogs are correct for those platforms? 
also need to know if the crashlog filename contains the string 'Kodi' on ios/osx.

see: https://github.com/XBMC-Addons/script.kodi.loguploader/blob/master/default.py#L77-L82

^ @Memphiz perhaps?


Memphiz:
Nice one :) - well i think the pathes are correct. On older ios versions (at least 5 and 6) you can additionally search for the word "Latest" because the last crashreport is symlinked to something like "Latest-Kodi.plist". On newer ios versions and osx (ios9) sadly this is not the case anymore. Those are called:

ios9
Kodi-2016-08-23-065523.ips.synced (or without synced - don't bother the file extension).

and osx:
Kodi_2016_08_23_065523_hostnameofthemac.crash

see screenshots from ios5,6,9 and osx11
https://www.dropbox.com/sh/fxh78oh7maw1eca/AABfP1w91fV7Y79hyP_xqyLTa?dl=0

This at least covers all jailbroken devices - i think on non-jailbroken its a bit different again (have none here to test atm...)

Thx for the effort

ronie:
great, thx for the help!

PR #252, opened 2016-08-29
[service.kn.switchtimer] 1.2.4

b-jesch:


phil65:
the files from .idea folder shouldnt get included.

service.kn.switchtimer/addon.xml
```diff
@@ -0,0 +1,36 @@
...
+    <extension point="xbmc.service" library="service.py" start="login" />
+    <extension point="xbmc.python.script" library="handler.py" />
+    <extension point="kodi.context.item">
+        <menu id="kodi.core.main">
+            <item library="addtimer.py">
+                <label>30040</label>
+                <visible>System.HasAddon(service.kn.switchtimer) + Window.IsVisible(tvguide)</visible>
+            </item>
+            <item library="deltimer.py">
+                <label>30041</label>
```
phil65: are these just some shortcuts or does the acton relate to the focused listitem?

phil65: Also, checkin System.HasAddon(service.kn.switchtimer) within a script called service.kn.switchtimer doesnt make too much sense. ;)

service.kn.switchtimer/addtimer.py
```diff
@@ -0,0 +1,17 @@
+import xbmc
+import handler
+
+if __name__ ==  '__main__':
+
+    handler.notifyLog('Parameter handler called: add Timer')
+    try:
+        args = {'channel':xbmc.getInfoLabel('ListItem.ChannelName'),
```
phil65: a better way would be to use sys.listitem instead of xbmc.getInfoLabel('ListItem.xxx').
(because stuff runs async and because listitem.xxx infolabel only works inside media windows)

phil65: Although I am not sure if it is possible to get the channelname that way....
If not, then your way is fine enough.

service.kn.switchtimer/handler.py
```diff
@@ -0,0 +1,161 @@
...
+        notifyLog('Couldn\'t parse date: %s' % (date), xbmc.LOGERROR)
+        notifyOSD(__LS__(30000), __LS__(30020), icon=__IconAlert__)
+        return False
+    return int(time.mktime(dtime.timetuple()))
+
+def readTimerStrings():
+    xbmc.sleep(1000)
+    timers = []
+    for prefix in ['t0:', 't1:', 't2:', 't3:', 't4:', 't5:', 't6:', 't7:', 't8:', 't9:']:
+        if xbmc.getInfoLabel('Skin.String(%s)' % (prefix + 'date')) != '':
```
phil65: hmm normally add-ons shouldnt deal with skin-related stuff... why do you need to deal with skin strings?

service.kn.switchtimer/addon.xml
b-jesch: The action is focused to the related item. [This link](http://www.kodinerds.net/index.php/Thread/49908-Release-service-kn-switchtimer-Ein-Umschalttimer-f%C3%BCr-den-PVR/#post279698) (second screenshot) shows the actions.

service.kn.switchtimer/resources/language/English/strings.xml
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<strings>
```
phil65: xml based language files are deprecated, please convert them to .po files.
https://github.com/alanwww1/xbmc-xml2po
https://github.com/alanwww1/xbmc-xml2po/downloads

phil65:
see comments Thx for contributing!

service.kn.switchtimer/addon.xml
b-jesch: System.HasAddon(...) makes sense if this addon is deactivated.  In this case the context menu item disappears and appear again if addon will activated. Sure it make no sense if addon is deinstalled.

service.kn.switchtimer/addtimer.py
b-jesch: This is the only way I know to get infos from within a PVROSDGuide window. If you look inside a skin you'll find only those named ListItems.xxx.

service.kn.switchtimer/handler.py
b-jesch: All timers are stored as strings within the skin (settings) as extra strings like t0:date, t0:channel etc. The reason is to store them permanent so they will available again after an reset, restart, ... of the system.

service.kn.switchtimer/resources/language/English/strings.xml
b-jesch: Will do that.

service.kn.switchtimer/handler.py
phil65: please save them to the addon_data folder of your script, or, as an easy alternative, as add-on settings.

service.kn.switchtimer/addtimer.py
phil65: yes you might be right, the PVRInfoTag is not available via python afaik.
Ignore this then.

service.kn.switchtimer/handler.py
b-jesch: I can do that but skinners have no chance to access the data if they are stored outside the skin. Doing this way I have to read them from settings/addon data and store them back into the skin again anyway otherwise they are not passed thru the skin. Yes it's skin related stuff. I've talked about that with the skinners in february this year. There's a [long thread](http://www.kodinerds.net/index.php/Thread/48312-Wir-suchen-einen-Skripter/) about this.

service.kn.switchtimer/addon.xml
phil65: The context menu entry always disappears when add-on is de-activated. No need to check for that visible condition.

service.kn.switchtimer/handler.py
phil65: hmm ok, skinners needing access to that info might be a valid reason.
Fine for me then.

phil65: Oh, since this add-on is a service, you could also use window properties for all this. that would be the nicest solution.

service.kn.switchtimer/addon.xml
b-jesch: Ok, then I cut this out

service.kn.switchtimer/handler.py
b-jesch: Yes, stored into window properties is a nice solution, but there's no persistent storage (crash, reboot, etc). To do that a rework is needed at some skins (I already hear the guys from kodinerds whistling ;) ). Maybe a bigger change in the kodi version (Krypton?) is a good deal to  implement those things as properties.

phil65: Storing data persistently should always be done within the add-on, and not spread around all skin config files. If you try to change it in a later version then I am fine with this for now though.  :+1:

b-jesch: Thank you! Some things have grown in history already and they could break only with a bigger change yet. This addon was initial created in a kodi version (dont know which, time runs fast) where all settings of skins are stored into one file (guisettings.xml). I'll implement the additional steps to store/restore settings from/within accessible skin values in a later version.

PR #255, opened 2016-09-02
[service.kn.switchtimer] 1.2.4

b-jesch:
Continued from [here:](https://github.com/xbmc/repo-scripts/pull/252)

* language files changed to Transifex .po
* visible cond of context menu changed


learningit:
Can I ask that you do a diff next time? it's a pain in the rear as the commit comes through as every line of the file is new and it's not easy to see what changes you actually made

b-jesch:
Will try that. I have massive problems to squash commits into one so I've started at zero. Something is going wrong with my dev environment or I must do something wrong. Don't know what. Sorry.

PR #259, opened 2016-09-07
[script.screensaver.football.panel] 1.4.0

enen92:
This PR implements some caching mechanisms for team logos and league data as requested by @zag2me in http://forum.kodi.tv/showthread.php?tid=258252&pid=2408318#pid2408318 in order to reduce the database server load.

Many thanks

### DIFF
--- a/script.screensaver.football.panel/resources/lib/common_addon.py
+++ b/script.screensaver.football.panel/resources/lib/common_addon.py
@@ -30,6 +30,9 @@ addon_path = addon.getAddonInfo('path')
 addon_userdata = xbmc.translatePath(addon.getAddonInfo('profile')).decode('utf-8')
 addon_name = addon.getAddonInfo('name')
 
+addon_userdata_cached_leagues = os.path.join(addon_userdata,"leagues")
+addon_userdata_cached_teams = os.path.join(addon_userdata,"teams")
+
 ignored_league_list_file = os.path.join(addon_userdata,"ignored.txt") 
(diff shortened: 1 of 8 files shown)


zag2me:
Thanks, Is the default 7 days? Does it need to be so short? 

I can't imagine the team logos and details changing too much over the season.

enen92:
I can increase it. What is your suggestion for the list of values?

zag2me:
I'd say 2 weeks as a start, lets see how that goes

script.screensaver.football.panel/resources/lib/common_addon.py
```diff
@@ -59,3 +62,16 @@ def removeNonAscii(s):
 
 def translate(text):
 	return addon.getLocalizedString(text).encode('utf-8')
+
+def filewrite(_file,contents):
+    f = open(_file, "w")
```
phil65: would be nice to use "with" statement for file operations to ensure that they get closed properly in case of error.

script.screensaver.football.panel/resources/lib/common_addon.py
```diff
@@ -59,3 +62,16 @@ def removeNonAscii(s):
...
+    f.write(contents)
+    f.close()
+    return
+
+def fileread(_file):
+    f = open(_file,"r")
```
phil65: same here.

enen92:
Updated PR with @zag2me and @phil65 suggestions

phil65:
Thx!

PR #263, opened 2016-09-16
Added protobuf 3.0.0 addon

WoLpH:
As discussed on Trac earlier: http://trac.kodi.tv/ticket/16878

This is a new release of the protobuf library which includes the missing descriptor_pb2.py fix.

If desired I can create pull requests for the other branches with the older releases as well but I don't think there's any reason to do that right now.

### DIFF
--- a/service.subtitles.subztv.gr/addon.xml
+++ b/service.subtitles.subztv.gr/addon.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<addon id="service.subtitles.subztv.gr" name="SubzTV" version="1.0.0" provider-name="lambda">
+<addon id="service.subtitles.subztv.gr" name="SubzTV" version="1.0.1" provider-name="lambda">
 	<requires>
 		<import addon="xbmc.python" version="2.24.0"/>
 		<import addon="script.module.lamlib" version="1.0.0" />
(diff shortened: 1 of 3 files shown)


ronie:
is this version fully bw compatible to v2?

i ask since i read this on their github page:
```we do not generally recommend that existing users migrate from proto2 from proto3 due
to API incompatibility```

WoLpH:
No, it's not fully backwards compatible unfortunately.

Having that said, I personally haven't encountered any issues with the Hyperion addon when upgrading from proto2 to proto3.

With regards to Python the only difference I could find is the removal of the `HasField` method: https://developers.google.com/protocol-buffers/docs/reference/python-generated#singular-fields-proto2

ronie:
perhaps it should be added as a new addon ( script.module.protobuf3) instead of replacing the old one?

phil65:
yup. new id, please.

WoLpH:
Yep, will do :)
I'll add an update for this one and add the new one as well. Stay tuned!

WoLpH:
For clarity... shouldn't we rename the old one to protobuf2? I think people might inadvertently expect that "protobuf" is current and "protobuf3" is an old release. Unfortunately that still breaks backwards compatibility though :(

ronie:
nope, we don't wanne break any of the addons using protobuf.

WoLpH:
Ok, I believe it's all done now :)

ronie:
1 please squash your commits
2 use "[script.module.protobuf3] v3.0.0" as the commit message
3 please remove protobuf2 from this pr. we already have it in our repo

WoLpH:
The protobuf2 library included is broken however. And heavily outdated.
Should I create a separate pull request for it?

ronie:
yup, please do.

PR #264, opened 2016-09-18
[script.matchcenter] 2.1.1

enen92:
-Do not show dialog while livescores are loading
-Fix english league 1 table
-Group substitutions
-Allow match info display even if no teams match against the database
-Resize windows if scrollbar is not visible
-Ignore leagues from the context menu of the livescores dialog
-better loading controls

As I already submitted 2 updates recently feel free to just review whenever possible

Thanks


script.matchcenter/resources/skins/default/720p/script-matchcenter-Livescores.xml
```diff
@@ -11,17 +11,18 @@
 		<effect type="fade" start="100" end="0" time="240" />
 	</animation>
 	<controls>
-		<control type="image">
+		<control type="image" id="32502">
 			<left>158</left>
 			<top>24</top>
 			<width>966</width>
 			<height>663</height>
+			<visible>IsEmpty(Window(home).Property(loading))</visible>
```
phil65: would be cool to attach these properties to the script window, or, as an alternative, prefix the property name.

enen92: Yap I noticed some problems while reusing the same property names with more than 1 addon. Will do, thanks for remembering.

enen92: By the way @phil65 do you know how I can resize a window depending on the presence (or not) of a scrollbar? Right now I am doing it manually in python but it would be much better if all that stuff could be achieved though xml. Unfortunately I couldn't manage to have conditionals on the window width

enen92:
Code changed to address @phil65 suggestions

script.matchcenter/resources/skins/default/720p/script-matchcenter-Livescores.xml
phil65: what do you mean with window width? the width of a dialog?

PR #266, opened 2016-09-19
[script.module.protobuf3] v3.0.0

WoLpH:
Added the protobuf3 addon as discussed in #263 

@ronie is this ok? :)

PR #268, opened 2016-09-21
script.module.mutagen - Delete

robwebset:
I'd like to propose deleting script.module.mutagen from the repo. I believe this request is consistent with previous deletions of modules as discussed here:

http://forum.kodi.tv/showthread.php?tid=181896&pid=1587067#pid1587067

This module is no longer maintained, and is VERY out of date - having been bumped from release to release from Eden, there is nothing in the repo that uses this module and there is no longer a maintainer to look after it.

Thanks
Rob


robwebset:
I also believe the following backs up why this should be deleted:
https://github.com/xbmc/repo-scripts/pull/266#issuecomment-248167634

Thanks
Rob

MartijnKaijser:
Not acceptable

robwebset:
I have raised some questions on why this can't be removed on the following forum post:
http://forum.kodi.tv/showthread.php?tid=288751
Please can you take a look

ronie:
there are a couple of addons in our repo that include the mutagen module in the addon itself.
those addons don't depend on the script module, but on older versions of kodi, the script module will be used (if it's installed) instead of the internal one.

so no, there's no addon depending on it, removing it from the repo won't break those addons,
but if a newer version gets installed through another repo (which this pr is all about) breakage can occur.

PR #272, opened 2016-09-24
script.skin.helper.colorpicker 1.0.1

marcelveldt:
Colorpicker from skinhelperservice now split into a seperate addon.

Sorry for the large bunch of images included, these are all the colorswatches for the included default set of palettes. While the colorfiles are auto generated by the addon using PIL, I have experienced that the PIL module isn't available on Android so I decided to pre build the color swatches and include them.

If the PIL issue on Android is already solved, please let me know so I can remove the images.

Thanks in advance for adding to the repo !

Cheers,

Marcel

PS: I have some more addons coming up but I'll wait a while before submitting them as I know you guys are busy. Let me know if you need any help.


script.skin.helper.colorpicker/addon.xml
```diff
@@ -0,0 +1,19 @@
+﻿<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+	<addon id="script.skin.helper.colorpicker" name="Skin Helper Service ColorPicker" version="1.0.0" provider-name="marcelveldt">
+	<requires>
+		<import addon="xbmc.python" version="2.13.0"/>
+		<import addon="xbmc.addon" version="12.0.0"/>
+		<import addon="script.module.pil" version="1.1.7"/>
```
phil65: Why the PIL dependency?

script.skin.helper.colorpicker/default.py
```diff
@@ -0,0 +1,58 @@
...
+    for i in range(40):
+        if not (xbmc.getCondVisibility("Window.IsActive(DialogSelect.xml) | Window.IsActive(script-skin_helper_service-ColorPicker.xml) | Window.IsActive(DialogKeyboard.xml)")):
+            break
+        else: xbmc.sleep(100)
+    
+
+#MAIN ENTRY POINT    
+params = getParams()
+
+if params and params.get("ACTION","").upper() == "COLORPICKER":
```
phil65: you can probably leave out the action parameter now? (since it only supports one action)

script.skin.helper.colorpicker/addon.xml
marcelveldt: PIL is used to create the color swatches that can be loaded in a list so skinners can display the picker with their custom colors.

I did this because you can't use colordiffuse with $INFO value inside the itemtemplate

script.skin.helper.colorpicker/default.py
marcelveldt: done

script.skin.helper.colorpicker/addon.xml
phil65: > PIL is used to create the color swatches

no, it's not. ;)

marcelveldt: Maybe you're missing it as the diff might be to large for Github to display properly...

https://github.com/marcelveldt/script.skin.helper.colorpicker/blob/master/resources/lib/ColorPicker.py#L13

https://github.com/marcelveldt/script.skin.helper.colorpicker/blob/master/resources/lib/ColorPicker.py#L73

Offcourse if you have a great idea how to avoid the creating of the images for each color, I'd be very happy to drop PIL immediately :-)
What I already tried was setting a listitem property to the color hex string but I can't use that as colordiffuse within the itemtemplate

phil65: oops, youre right, thx.

phil65: hmm no, you forgot to include ColorPicker.py.

marcelveldt: Sorry, that file is included:
https://github.com/marcelveldt/repo-scripts/tree/script.skin.helper.colorpicker/script.skin.helper.colorpicker/resources/lib

I think it's because all these PNG files that it's too much for Github to handle in the browser

phil65: youre right, github completely eats that file :/

phil65: oh sorry for net gettin back to you, this PR slipped through my fingers.
What does make you think that PIL is not available on android btw? @MartijnKaijser , you know anything about this?

phil65: ...and does it really have to be >1500 colors? :)  This add-on has almost as many files as our complete default skin..

marcelveldt: Well if you think that PIL actually is supported on Android, I can drop the image files all together so they can just be built on demand if needed, that would be a great advantage. I remember from like 2 years ago that the PIL package wasn't working on Android devices. 

The reason I use the image files for the colors in the first place because there doesn't seem to be a method to set the colordiffuse tag with an info property inside the itemlayout. Otherwise I could have filled the list with just a white image and the colordiffuse and I wouldn't be needing PIL and images at all.

marcelveldt: http://forum.kodi.tv/showthread.php?tid=235676&pid=2121971#pid2121971

marcelveldt: Shall I drop this PR and create a new one or just make the changes and commit to this existing PR so our conversation won't get lost ?

phil65: just force-push, as usual.

marcelveldt:
@phil65 Changes are now applied, the PR is updated.

PR #273, opened 2016-09-25
[script.artistslideshow] 2.0.3

pkscout:
v.2.0.3
- removed support for htbackdrops.org (API access no longer available)
- fix for crash when theaudiodb.com doesn't respond with valid JSON


script.artistslideshow/addon.xml
```diff
@@ -10,6 +10,8 @@
 		<provides>executable</provides>
 	</extension>
 	<extension point="xbmc.addon.metadata">
+		<news>2.0.3
+		- removed support for htbackdrops.org (API access no longer avaialble)</news>
```
phil65: typo :)

pkscout: That's the new US spelling of available.  '-)  Extra letters make it more better.

Do you want me to resubmit?

phil65: You could update the PR by force-pushing.
If it is too much hassle for you then I would also be fine to just keep that typo included. What´s your preference?

pkscout:
I'm fine with the typo. Nobody really reads that anyway. ;)

PR #274, opened 2016-09-27
screensaver.playlist 0.0.1

BigNoid:
Simple screensaver that plays a (smart) playlist (.m3u or .xsp)


phil65:
I think the whole functionality is already provided by "Partymode autostart" service?

BigNoid:
There is also a video screensaver add-on, although not on the repo anymore but those all are (also) service addons. I just wanted a simple script that only runs when the screensaver kicks in.

phil65:
Hmm I dont think this really should be done as a screensaver addon since it is not really a screensaver. I would prefer to just extend that "Partymode autostart" add-on, that way users can start a playlist and have a regular screensaver kick in at the same time.
Anyways, just personal opinion, if others think this makes sense then feel free to merge.

phil65:
One more thing to think about: How does kodi behave if this screensaver is chosen and "Partymode autostart" is active at the same time?

BigNoid:
> One more thing to think about: How does kodi behave if this screensaver is chosen and "Partymode autostart" is active at the same time?

To be honest, I think that should be a concern for that add-on as it doesn't register itself as a screensaver.

phil65:
..which makes sense since it allows it to be used in combination with a regular screensaver.

screensaver.playlist/default.py
```diff
@@ -0,0 +1,68 @@
...
+            xbmc.Player().play(self.parse_playlist(playlist, random))
+            xbmc.executebuiltin("PlayerControl(RepeatAll)")
+        elif playlist.endswith('.xsp'):
+            xbmc.Player().play(playlist)
+            xbmc.executebuiltin("PlayerControl(RepeatAll)")
+        else:
+            xbmc.executebuiltin('Notification(%s, %s, %d, %s)' % (ADDON_NAME, ADDON_LANGUAGE(30003), 5000, ADDON_ICON))
+
+    def parse_playlist(self, playlist, random):
+        playlist = open(playlist, 'r')
```
phil65: please use "with" statement for opening files.

BigNoid:
The party mode doesn't kick in when this screensaver start. That script monitors is something is playing.



> Do we really have to parse the m3u manually? Cant we just throw it at kodi player?

Yes, this is the only way to truly randomise the playlist on every start. PlayerControl(RepeatOn) works but than it still always starts with the same video and the sequence after that is random.

phil65:
Ok. One comment concerning code, rest looks good enough.

I´ll leave it to the other guys if we want this functionality implemented as a screensaver. :)

phil65:
Just had a look in docs, did you try something like
temp_playlist = xbmc.Playlist()
temp_playlist.load(PATH_TO_M3U)
temp.playlist.shuffle()
xbmc.Player().play(temp_playlist)

BigNoid:
> Just had a look in docs, did you try something like
> temp_playlist = xbmc.Playlist(xbmc.PLAYLIST_MUSIC)
> temp_playlist.load(PATH_TO_M3U)
> temp_playlist.shuffle()
> xbmc.Player().play(temp_playlis

Completely missed that load function. Much easier that way, thx :thumbsup:

Oh did another test with party mode screensaver function, but that one overrides my screensaver. Still no big deal imo as it makes no sense to have both enabled. I made this btw with playing videos in mind and not music (maybe I should choose a different fanart). Currently there's no screensaver in repo that can play user defined videos.

PR #292, opened 2016-10-25
[script.extendedinfo] 5.5.2

phil65:


script.extendedinfo/resources/lib/dialogs/DialogVideoList.py
```diff
@@ -48,6 +49,7 @@ class DialogVideoList(DialogBaseList, window_type):
                    "with_networks": addon.LANG(32152),
                    "with_keywords": addon.LANG(32114),
                    "first_air_date": addon.LANG(20416),
+                   "with_runtime": xbmc.getLocalizedString(2050),
```
enen92: You could have called LANG method on kodi65 module like you do with all the other cases (too keep everything uniform). If the id is not between 31000 and 33000 the module will call xbmc.getLocalizedString(id_)
https://github.com/phil65/script.module.kodi65/blob/master/lib/kodi65/addon.py#L47

phil65: oops, youre right, KodiDevKit uses xbmc.getLocalizedString by default, that's why this ended up here. Will fix for next update.

PR #297, opened 2016-11-04
Omit the additional info from PRs

Razzeee:
This change will still show the infos when a submitter is filling in a PR, but it will be hidden from the actual PR.


Razzeee:
@MartijnKaijser please review.

I will do a typo correction run after this

hudokkow:
looks good

MartijnKaijser:
i actually left it there by intention

Razzeee:
Okay, why would that be good?

MartijnKaijser:
Cause most do t read what's in that description. Once they complete the PR its in their face and also for other to see

PR #336, opened 2016-12-18
[script.module.web-pdb] 1.3.0

romanvm:
### Description
This is a port of updates from the mainstream `web-pdb` package. I've also changed the order of `extension` tags so that the addon would correctly appear in the Kodi Addons Manager.

### Checklist:

- [x ] My code follows the [add-on rules](http://kodi.wiki/view/Add-on_rules) and [piracy stance](http://kodi.wiki/view/Official:Forum_rules#Piracy_Policy) of this project. 
- [x ] I have read the [CONTRIBUTING](https://github.com/xbmc/repo-scripts/blob/master/CONTRIBUTING.md) document
- [ x] Each add-on submission should be a single commit with using the following style: [sciprt.foo.bar] v1.0.0


phil65:
Thx!

PR #337, opened 2016-12-23
[weather.ozweather] 0.8.8

bossanova808:
### Description
- Add Wind Speed/Direction support for the forecast data
- Bugfixes to cope with more different labels skins use - all thsi weather stuff could really do with some standardisation & proper docs!!

I'm still a bit new to the git subtree stuff, fingers crossed.  I had to use `--allow-unrelated-histories`  when doing this: `git pull --strategy subtree --squash weather.ozweather master`

### Checklist:

- [X] Each add-on submission should be a single commit with using the following style: [sciprt.foo.bar] v1.0.0

### DIFF
--- a/weather.ozweather/addon.xml
+++ b/weather.ozweather/addon.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<addon id="weather.ozweather" name="Oz Weather" version="0.8.7" provider-name="Bossanova808">
+<addon id="weather.ozweather" name="Oz Weather" version="0.8.8" provider-name="Bossanova808">
 	<requires>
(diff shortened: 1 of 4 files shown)


ronie:
looks good, thx!

for your next update, please replace the print statements with xbmc.log()
see: http://kodi.wiki/index.php?title=Add-on_rules#Requirements_for_scripts_and_plugins

bossanova808:
What's the best way to manage print vs. xbmc.log in unit testable modules?  I need prints in those...but I suppose I could write a local function which tests how it is being run eh?

ronie:
those aren't a problem at all, just keep print there. :-)

it's only about messages that end up in the kodi.log file,
those should use the xbmc.log() functionality

PR #344, opened 2017-01-03
[script.speedfaninfo] 0.7.5

pkscout:
v.0.7.5
- added option to run external script to generate log file
- added 'Other' option to sensor display
- added advanced setting for line read size


pkscout:
Please don't do this pull request.  First, I found a bug with AS 2.0.5 that I need to fix first.  Second, I somehow got pull requests for two of my addons into one request, and I know we're not suppose to do this.

PR #354, opened 2017-01-11
[service.subtitles.subtitle] 4.6.0

CaTzil:
### Description
* Add CAPTCH support
* Check item's media type
### Checklist:

Additional information :

### DIFF
--- a/service.subtitles.subtitle/service.py
+++ b/service.subtitles.subtitle/service.py
@@ -125,11 +125,13 @@ def takeTitleFromFocusedItem():
     labelTVShowTitle = xbmc.getInfoLabel("ListItem.TVShowTitle")
     labelSeason = xbmc.getInfoLabel("ListItem.Season")
     labelEpisode = xbmc.getInfoLabel("ListItem.Episode")
+    isItMovie = xbmc.getCondVisibility("Container.Content(movies)") or labelType == 'movie'
+    isItEpisode = xbmc.getCondVisibility("Container.Content(episodes)") or labelType == 'episode'
 
     title = 'SearchFor...'
-    if labelType == 'movie' and labelMovieTitle and labelYear:
+    if isItMovie and labelMovieTitle and labelYear:
         title = labelMovieTitle + " " + labelYear
-    elif labelType == 'episode' and labelTVShowTitle and labelSeason and labelEpisode:
+    elif isItEpisode and labelTVShowTitle and labelSeason and labelEpisode:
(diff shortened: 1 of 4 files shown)


service.subtitles.subtitle/service.py
```diff
@@ -125,11 +125,13 @@ def takeTitleFromFocusedItem():
     labelTVShowTitle = xbmc.getInfoLabel("ListItem.TVShowTitle")
     labelSeason = xbmc.getInfoLabel("ListItem.Season")
     labelEpisode = xbmc.getInfoLabel("ListItem.Episode")
+    isItMovie = xbmc.getCondVisibility("Container.Content(movies)") or labelType == 'movie'
```
phil65: please use listitem.dbtype for this.

burekas7: Some of the addons doesn't set this property properly (DBTYPE) , so this condition is kind of second check in addition to listitem.dbtype.
(By the way, you already approved this code in another addon)

phil65: might be that i overlooked that, i certainly dont spot every small detail when reviewing. Feel free to keep it this way, but it is conceptually wrong and will fail in other cases. Other add-ons doing it wrong is not really a good argument imo, someone has to start doing it right. ;)

CaTzil: Can you explain what you have said about "will fail in other cases."?
What other cases?

phil65: example: link movies to a tvshow, then you get movies in a list with container.content = tvshows. ListItem.Dbtype will be correct in that case, container.content will be misleading.

CaTzil: so, do you suggest to flip the condition, so the first thing that will be checked is the  infolabel, and if that not exists check the other part?

phil65: My suggestion is to just use listitem.dbtype. Dont use container.content() at all here.

CaTzil: Ok,
I`ve reviewed the code again, and the code not affects regular usage of subtitle addon.
It only solves specific scenario for users that want to check if there is a subtitle from context menu.

CaTzil: @phil65 did you check what i`ve said?
The new code not affects regular addon usage.

MartijnKaijser:
Improvements can be done next update

PR #410, opened 2017-02-01
[script.module.future] v0.16.0.1

Tadly:
### Description

I missed the original PR (and would have said something if I didn't) but the hierarchy with this module isn't quite right (from the perspective of the future module).
You'd now have to e.g. `from future.future.utils import PY3` rather than `from future.utils import PY3`
(actually broke one of my add-ons as the kodi repo seems to take precedence over 3rd-Party ones)

I'm happy to talk about this change if there are any concerns...

Oh right, the version number.
We sadly don't have something like a **package-version** (which would be nice) and rather than bringing the version-number out-of-sync with the one `future` uses I decided to add the fourth `.1` Objections?

### Checklist:

Additional information :


Tadly:
The changes in those `__init__.py` are because everything moved up one directory and git confuses them. (I recreated the PR 3 times because I got confused as well)

script.module.future/addon.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<addon id="script.module.future" name="future" version="0.16.0" provider-name="Ed Schofield">
+<addon id="script.module.future" name="future" version="0.16.0.1" provider-name="Ed Schofield">
```
Razzeee: I'm not sure if that's possible, however you can do `0.16.0~whatever`

enen92: I am not sure either but https://github.com/xbmc/repo-plugins/pull/916 was similar and it was merged despite my comment. So I guess it is fine.

Tadly: wasn't sure about that either.
Should I go for `0.16.0~1` than? I'm really clueless on how to give that thing a somewhat proper version number in this case :/
Or `0.16.0~restructured`? I hate version-number that have anything other than `.` and `0-9` :D

Razzeee: @MartijnKaijser will know the correct way

MartijnKaijser: a .1 shouldn't be a real issue but not the preferred "standard".

Tadly: Don't get me wrong, I prefer x.y.z...
I just wanted to prevent the "future module has a new version 0.16.1, now we can't keep the version number in sync".
The additional .1 is the best I could come up with :P

Tadly:
@tyl0re I'm bringing you in as you are the only "official" person using it
(I got forced to by the higher prio of the kodi repo :D)

tyl0re:
@Tadly Have too Look what must change I am not usining it , but i use modules that Uses it, so im Not shure if it still works when changing it, and what must be changed. Would be not so good if i have to change the Moduels, because than that Module Cant be upgraded with just put the Official code to it

by the why is your 
0.16.0.1

When you change the Code from:
-__ver_minor__ = 16
-__ver_patch__ = 0
To 
+__ver_minor__ = 15
+__ver_patch__ = 2

Why making a downgrade, with an Higher Number ?

Tadly:
@tyl0re for the sake of backlogging, I'll throw the stuff from slack in here as well :)

First of, I'm pretty sure you'r dependency will work just fine but testing first wouldn't hurt.

Second, I actually didn't touch any source. All I did is move everything from within `lib/future/*` to `lib/*`
On your end check:
* `script.module.future/lib/future/__init__.py`
* `script.module.future/lib/future/future/__init__.py`
That will hopefully clear up the confusion :)
(hint: because everything moved up one dir, git thinks those files changed rather than being moved)

After recreating the module myself `lib/future/__init__.py` would end up being deleted (the original future module doesn't have that one)
So... want me to get rid of it?

Tadly:
Quick heads-up.
I removed the `lib/__init__.py` and changed the file-permissions (they had 755 which isn't quit right) and force pushed those changes.

@tyl0re could you test if this is working for you?

tyl0re:
Thanks it is working. Could be Merged from my side

PR #415, opened 2017-02-05
[script.module.simplecache] v1.0.11

marcelveldt:
### Description
Simple, stateless caching for kodi python addons.
Will be used in all my projects from now on and also usefull for other devs.
### Checklist:
- [X ] I have read the [CONTRIBUTING](https://github.com/xbmc/repo-scripts/blob/master/CONTRIBUTING.md) document


script.module.simplecache/lib/simplecache.py
```diff
@@ -0,0 +1,282 @@
...
+        self.busy_tasks.remove(__name__)
+        self.win.setProperty("simplecache.clean.lastexecuted", repr(cur_time))
+        self.win.clearProperty("simplecachecleanbusy")
+        self.log_msg("Auto cleanup done")
+
+    def get_database(self):
+        '''get reference to our sqllite database - performs basic integrity check'''
+        try:
+            dbpath = xbmc.translatePath("special://database/simplecache.db").decode('utf-8')
+            connection = sqlite3.connect(dbpath, timeout=30, isolation_level=None)
```
MartijnKaijser: You are not allowed to write in that folder. You mustt use the Addon data one

marcelveldt:
@MartijnKaijser now fixed, changed it to the addon_data folder instead.

script.module.simplecache/README.md
```diff
@@ -0,0 +1,85 @@
...
+</requires>
+```
+
+Now, to use it in your Kodi addon/script, make sure to import it and you can access it's methods.
+
+```
+import simplecache
+
+#get data from cache
+mycache = simplecache.get("MyChunkOfData")
```
ronie: perhaps use a prefixed endpoint example here? i know it's explained a little further down (i would even state it's 'mandatory' instead of labeling it as a 'tip'), but with no prefix, it's very easy to overwrite each others data, or even window properties set by a skin or other addons.

script.module.simplecache/lib/simplecache.py
```diff
@@ -0,0 +1,283 @@
...
+
+        # remove task from list
+        self.busy_tasks.remove(__name__)
+        self.win.setProperty("simplecache.clean.lastexecuted", repr(cur_time))
+        self.win.clearProperty("simplecachecleanbusy")
+        self.log_msg("Auto cleanup done")
+
+    def get_database(self):
+        '''get reference to our sqllite database - performs basic integrity check'''
+        dbpath = "special://userdata/addon_data/script.module.simplecache/"
```
ronie: it's better to use:
`dbpath = xbmcaddon.Addon().getAddonInfo('profile')`
instead of hardcoding the path.

marcelveldt: I'm not so sure about that. When the module is called from a service addon, that will result in deprecated warnings in the log when xbmcaddon.Addon is called with no addon id specified.
The only improvement I can make here is defining a constant with the addon ID, and use xbmcaddon.Addon(ADDONID) but that doesn't make much of a difference imo.

What do you want me to do here ?

script.module.simplecache/README.md
marcelveldt: The internals are all prefixed/unique but good idea indeed to have a prefixed example. I'll change it right away.

marcelveldt:
@ronie both requests are now fixed.

ronie:
thx!

PR #420, opened 2017-02-05
[script.skin.helper.service] 1.1.3

marcelveldt:
### Description
Refactored version of skin helper service. As discussed with several Kodi team members, it's now split up into several seperated addons and modules. Also the code is now fully PEP8 compliant and I monitor it automatically with codacy.
I did not yet check if I can offload some of the features to native Kodi core but that's something as a next step once this refactored version hits the repo.
### Checklist:
- [ X] I have read the [CONTRIBUTING](https://github.com/xbmc/repo-scripts/blob/master/CONTRIBUTING.md) document


ronie:
i recently bumped into an issue with this addon.

the skin helper addon seems to make a lot of calls to the Musicbrainz server when you are scanning your music collection to the library. 
since Musicbrainz throttles requests to 1 request per second, we run into problems when both the kodi music scraper and the skin helper addon are querying Musicbrainz simultaneously.
Musicbrainz starts blocking requests (from either the scraper or the addon) with the unfortunate result that a large section of your music collection fails to be scraped properly.

i'm not sure if this issue is still present in this new, refactored version, but in case it is... is there anything you could do to address this issue?

marcelveldt:
Hi Ronie, true indeed for the previous version (current version on the repo). It was making calls to MB to retrieve the musicbrainz ID's even if that info was already present in the Kodi db. Another error was that it was preloading stuff at kodi library scans, new version only purges the cache at library scans and pulls metadata only if needed/requested by the user and in that case the online metadata would still be in the cache.

For the new version I have completely rewritten it all, all metadata retrieval is now handled by the new metadata module I created which supports several metadata websites. All calls are properly cached and rate limited, the same applies to musicbrainz. Although it shouldn't make too many calls anymore to musicbrainz assuming that the Kodi scraper already has the info in the db.

Long story short... Yes, the issue you're describing is indeed fixed in the new version.

I created the new metadata module in a first attempt to get something general for grabbing the metadata. It seems such a waste to me that there are multiple addons (and even the kodi scraper) all grabbing the same info just for different purposes. 

In the long-term you could have some python or c++ based metadata modules to feed the kodi db with  information. Would be great if we no longer have to create all these window property based workarounds but just "tell" kodi that we want an additional field added to the database. It would be some sort of dynamic fields database instead of the current pretty hardcoded base. Maybe I'm just rambling here but imo this could be something cool to work on for next gen kodi releases. I'll be glad to help out with this.

For now, please check my PR for the updated skin helper addons, especially the new simplecache and metadata modules. This will for now be a huge step forward in a more robust handling of metadata.

ronie:
thanx for the detailed feedback! good to hear the issue isn't present anymore in the new version.

i'm currently writing a new (python based) metadata scraper for music.
in order to gather some ideas on how our scraping process can be improved, i have opened a thread on the forum: http://forum.kodi.tv/showthread.php?tid=306218

feel free to share your thoughts there.
i'm not sure if your addons provide any additional metadata that we don't already have in the music db,
but if it is, perhaps we can look into adding it to the kodi scraper.

marcelveldt:
@ronie great, I will share my ideas on that thread.

script.skin.helper.service/resources/lib/main_module.py
```diff
@@ -0,0 +1,695 @@
...
+        control = self.params.get("control")
+        fallback = self.params.get("fallback")
+        position = self.params.get("position", "0")
+        relativeposition = self.params.get("relativeposition")
+        if relativeposition:
+            position = int(relativeposition) - 1
+        count = 0
+        if control:
+            while not xbmc.getCondVisibility("Control.HasFocus(%s)" % control):
+                if xbmc.getCondVisibility("Window.IsActive(busydialog)"):
```
ronie: you might want to add some sleep here

script.skin.helper.service/resources/lib/main_module.py
```diff
@@ -0,0 +1,695 @@
...
+                    continue
+                elif count == 20 or (xbmc.getCondVisibility(
```
ronie: count will never reach 20 i think?
it's increased by 1 and you break from the loop.

script.skin.helper.service/resources/lib/listitem_monitor.py
```diff
@@ -0,0 +1,585 @@
...
+        log_msg("ListItemMonitor - stop called")
+        self.exit = True
+        self.event.set()
+
+    def run(self):
+        '''our main loop monitoring the listitem and folderpath changes'''
+        log_msg("ListItemMonitor - started")
+        self.get_settings()
+
+        while not self.exit:
```
ronie: it wouldn't hurt to check for abort requested here i think

script.skin.helper.service/LICENSE.txt
```diff
@@ -0,0 +1,339 @@
+                    GNU GENERAL PUBLIC LICENSE
```
ronie: duplicate license file ;-)

ronie:
@marcelveldt have you already had the chance to look into these?

marcelveldt:
@ronie yeah, sorry, I will update the PR tonight. I was busy becoming a father again last week ;-)

script.skin.helper.service/resources/lib/listitem_monitor.py
marcelveldt: well, that check is alread performed by the main_service which controls the additional threads. Once an abort is requested by Kodi, the exit flag will be set.

marcelveldt:
@ronie, PR is updated with the requested changes

ronie:
Great, thx for the massive amount of time you took to split up the addon!

> I was busy becoming a father again last week ;-)

Congrats mate!

marcelveldt:
Thanks ! I will inform skinners that the new version is up on the repo so they can submit updated versions of their skins.

PR #422, opened 2017-02-05
[script.skin.helper.skinbackup] 1.0.12

marcelveldt:
### Description
Another part split off from skin helper service;
Addon to (periodically) backup skin settings, including images and shortcuts.
### Checklist:
- [X ] I have read the [CONTRIBUTING](https://github.com/xbmc/repo-scripts/blob/master/CONTRIBUTING.md) document


script.skin.helper.skinbackup/default.py
```diff
@@ -0,0 +1,118 @@
...
+    def reset(self):
+        '''reset skin settings'''
+        backuprestore = BackupRestore()
+        filters = self.params.get("filter", [])
+        if filters:
+            filters = filters.split("|")
+        silent = self.params.get("silent", "") == "true"
+        backuprestore.reset(filters, silent)
+        xbmc.Monitor().waitForAbort(2)
+        xbmc.executebuiltin("RunScript(script.skin.helper.service,action=checkskinsettings)")
```
ronie: perhaps add a dependency for script.skin.helper.service in your addon.xml ?

script.skin.helper.skinbackup/resources/lib/backuprestore.py
```diff
@@ -0,0 +1,443 @@
...
+        if not filters or (filters and "skinshortcuts" in filters):
+            self.backup_skinshortcuts(temp_path + "skinshortcuts/")
+
+        # backup skin settings
+        if "skinshortcutsonly" not in filters:
+            skinsettings_path = os.path.join(temp_path, u"guisettings.txt")
+            self.backup_skinsettings(skinsettings_path, filters, temp_path)
+
+        # zip the backup
+        zip_temp = u'special://temp/skinbackup-%s.zip' % datetime.now().strftime('%Y-%m-%d %H.%M')
```
ronie: please do not use the kodi temp folder.

addons need to use their own addon_data subfolder for data storage.

script.skin.helper.skinbackup/service.py
```diff
@@ -0,0 +1,50 @@
...
+from resources.lib.colorthemes import ColorThemes
+from resources.lib.backuprestore import BackupRestore
+from resources.lib.utils import log_msg
+import xbmc
+import _strptime
+
+
+class Service():
+    '''Background service for automatic skin backups and/or automatic colortheme switches'''
+
```
ronie: what happens if you switch skins while the service is running?
is there a risk it restores the settings from the previous skin to the current one?

script.skin.helper.skinbackup/LICENSE.txt
```diff
@@ -0,0 +1,339 @@
+                    GNU GENERAL PUBLIC LICENSE
```
ronie: one license file should be enough i think? :-)

script.skin.helper.skinbackup/default.py
marcelveldt: Hmm, it's optional and should only be executed IF skinhelper service is even present. It's cenrtainly not a dependency. I will add the check in the code, thanks.

script.skin.helper.skinbackup/resources/lib/backuprestore.py
marcelveldt: Even for temporary files ? Never knew that so thanks, learned something ;-)
I will change it to a temp folder in the addon_data folder for the addon.

script.skin.helper.skinbackup/service.py
marcelveldt: In theory that shouldn't happen BUT I will double check it and place an additional check in place. Better be safe indeed. Thanks

script.skin.helper.skinbackup/LICENSE.txt
marcelveldt: Hmm, yes, copy/paste error I guess. Will fix it.

script.skin.helper.skinbackup/service.py
marcelveldt: OK, checked the code. Can never happen because it's controlled by skin settings so if you switch to another skin, that skin will have other settings. It's fail safe.

marcelveldt:
@ronie, issues are now addressed and committed to the PR

PR #459, opened 2017-02-21
[script.module.trakt] 2.13.0

Razzeee:
### Description

Updated due to problems. Be advised, this is the same version as in jarvis (https://github.com/xbmc/repo-scripts/tree/jarvis/script.module.trakt) only difference is the lower kodi dependency. The version number is the same, if that's a problem please let me know.

### Checklist:


MartijnKaijser:
That will indeed cause issue iirc. Our repo script will pull some version but not sure which

Razzeee:
Thought that might be the case. Naming it `2.12.0~v1` should work?

Razzeee:
Changed to `2.12.0~v1`

MartijnKaijser:
i have no idea if that will even work or how kodi will compare the versions

Razzeee:
Okay, the original work this is based upon has gone from 2.12.0 to 2.13.0 recently. I'll just update to the new version.

PR #487, opened 2017-04-17
[script.io.benchmark] 1.1.0

Lunatixz:
Test your devices read/write performance.

### Checklist:

Additional information :


script.io.benchmark/default.py
```diff
@@ -0,0 +1,154 @@
...
+def log(msg, level = xbmc.LOGDEBUG):
+    xbmc.log(ADDON_ID + '-' + ADDON_VERSION + '-' + str(msg), level)
+    
+def showText(text, header=ADDON_NAME):
+    xbmcgui.Dialog().textviewer(header, text)
+     
+def okay(line1= '', line2= '', line3= '', header=ADDON_NAME):
+    xbmcgui.Dialog().ok(header, line1, line2, line3)
+   
+try:
```
Tadly: All the following code should be encapsulated in:
```python
if __name__ == '__main__':
    ...
```

script.io.benchmark/default.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+# Plugin Info
+ADDON_ID       = 'script.io.benchmark'
+REAL_SETTINGS  = xbmcaddon.Addon(id=ADDON_ID)
+ADDON_NAME     = REAL_SETTINGS.getAddonInfo('name')
+ADDON_PATH     = (REAL_SETTINGS.getAddonInfo('path').decode('utf-8'))
+ADDON_VERSION  = REAL_SETTINGS.getAddonInfo('version')
+SETTINGS_LOC   = REAL_SETTINGS.getAddonInfo('profile')
+ICON           = os.path.join(ADDON_PATH, 'icon.png')
```
Tadly: `REAL_SETTINGS.getAddonInfo('icon')`

script.io.benchmark/default.py
```diff
@@ -0,0 +1,154 @@
...
+FANART         = os.path.join(ADDON_PATH, 'fanart.jpg')
```
Tadly: `REAL_SETTINGS.getAddonInfo('fanart')`

script.io.benchmark/default.py
Lunatixz: Okay made this change

script.io.benchmark/default.py
Lunatixz: If it's okay, I'll keep this as is... seems like a matter of style over function?

Tadly: using `getAddonInfo` is the generally preferred way. thanks for changing it :+1:

script.io.benchmark/default.py
Tadly: I don't know how much experience you have with python but just so you understand.

If source-code is free floating (not encapsulated within a method) it will be run whenever the file gets invoked.

What does this mean? In your case, if you or someone else would e.g. `import` this file, the whole source would be executed.
This is generally bad (unless you explicitly want this to happen).

In this case, we want the main routine to be run **only** when the script gets **executed**.
This is what `__name__` allows us to determine.

Lunatixz: correct me if i'm wrong... considering this isn't a module or a method meant to be important, ie the base default of the script. Is there something wrong with allow it to execute at once? Traditionally I work with classes but with the above it's simply meant to run on execution and quit. Anything more involved would have been written within individual classes.

script.io.benchmark/default.py
Lunatixz: okay... they are identical but okay :)

script.io.benchmark/default.py
Tadly: as far as I'm aware, it is **always** recommended to use the `if __name__ == '__main__':` clause as your entry point for execution (no matter if it's a file not meant to be imported or not).
Think of other languages like java, c++ etc. all have their way of defining the "main entry point" for what ever needs to be run.

Just because python allows us to have this kind of floating code doesn't mean it's good practices :)

now. In python we work module-based which means each *other* python file (or directory containing more python code) is considered a module.
These typically do **not** have this entry point as they usually are only used via imports (but at the same time, every piece of code-**logic** defined is within classes/methods).

[as an example for *floating* code](https://github.com/linuxwhatelse/plugin.audio.linuxwhatelse.gmusic/blob/master/addon/__init__.py) (couldn't think of another example except one of my addons) a `__init__.py` (declaring the **addon** directory as python module) may very well have some form of *floating* (no clue how to properly call that... I should research that at some point) code for declaration purposes.

**TL;DR**
Encapsulate as much logic within methods and you're fine. :)

script.io.benchmark/default.py
Tadly: The end-result (the path you're getting) yes, the way the path is built/retrieved...
...no idea actually.

Either way. Kodi (which provides the whole framework for us to work with) *should* know best how and where to retrieve things like icons etc. hence why should we do it?

In case core changes we (python add-on devs) can be like *"none of our business"* which is why we leverage things like these to whom has to deal with it anyways :)

script.io.benchmark/default.py
Lunatixz: THX u r correct, I was looking at the incorrect source via my dumb phone ;) I should have verified I was looking at the correct line of code, heh my bad.

Lunatixz:
@Tadly thanks again for your help... and sorry for the miscommunication. I've recently started offloading my personal beta repo addons to the Kodi repository.  Got my wires crossed on what code was being reviewed :)

script.io.benchmark/default.py
Tadly: Looking at the changes, this isn't quite right.
now even your methods are within `if __name__ == '__main__':`

How it should look like is:
```python
import ...
#<more imports>

ADDON_ID       = 'script.io.benchmark'
#<more global definitions>

def repeat_to_length(string_to_expand, length):
    return (string_to_expand * ((length/len(string_to_expand))+1))[:length]
#<all other methodes>

if __name__ == '__main__':
    try:
        params = dict( arg.split( '=' ) for arg in sys.argv[ 1 ].split( '&' ) )
    except:
        params = {}

    #<rest of the source>
```

Hope it is clearer now.

Lunatixz:
@enen92 can you squash and merge :)

Lunatixz:
Is there something I missed holding up this commit?

Tadly:
All the guys are quite busy recently.
@Razzeee maybe you got some time to have a secondary look?

Lunatixz:
Oh, no problem... Thought I might have missed a change :)

Lunatixz:
@Razzeee can you squash and merge for me :)

Razzeee:
I'll try to look at this tonight, sorry for the wait.

script.io.benchmark/default.py
```diff
@@ -0,0 +1,155 @@
+ #   Copyright (C) 2017 Lunatixz
+#
+#
+# This file is part of I/O Benchmark.
+#
+# I/O Benchmark is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
```
Razzeee: I'm a bit confused by this as the addon is listed with an MIT License?

Razzeee:
Otherwise then the license thing it looks good.
But I'm a bit shocked by my slow hardware :)

image

Lunatixz:
Perhaps I need to be more clear about the results?  Two issues. 
1) Read cache will bloat the value... so you need to review the max, min for the actual value. 
2) Everyone has diff hardware, and short from trying to detect hardware or requiring users input their hardware, ie ssd, flash, hdd, usb flash...  I took the best and worst performance benchmarks for every storage class and calculated a median value. Not sure if that was the best solution, taking a average would have made things worse since the performance gap between ssd and usb flash is very high.

script.io.benchmark/default.py
Lunatixz: Sorry copy and paste error, My standard lic is GNU... parts of this plugin are forked so I left its original MIT lic info... I'll correct the issue.

script.io.benchmark/LICENSE.md
```diff
@@ -0,0 +1,21 @@
+The MIT License (MIT)
+
+Copyright (c) 2016 thodnev
```
Razzeee: I would think this should be your nick. But this line should be in `monkeytest.py`
Also add your nick to default.py

Razzeee:
Thanks for your patience!

PR #492, opened 2017-04-24
[service.onmute] 0.0.1

Lunatixz:
Enable closed captioning and subtitles "OnMute".

### Checklist:

- [ ] My code follows the [add-on rules](http://kodi.wiki/view/Add-on_rules) and [piracy stance](http://kodi.wiki/view/Official:Forum_rules#Piracy_Policy) of this project. 
- [ ] I have read the [CONTRIBUTING](https://github.com/xbmc/repo-scripts/blob/master/CONTRIBUTING.md) document
- [ ] Each add-on submission should be a single commit with using the following style: [script.foo.bar] v1.0.0

Additional information :


service.onmute/service.py
```diff
@@ -0,0 +1,215 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with OnMute.  If not, see <http://www.gnu.org/licenses/>.
+
+import sys, os, re, traceback, json
+import xbmc, xbmcplugin, xbmcaddon, xbmcgui
+
+# Plugin Info
+ADDON_ID = 'service.onmute'
+REAL_SETTINGS = xbmcaddon.Addon(id=ADDON_ID)
```
Tadly: using `xbmcaddon.Addon()` will ALWAYS give you an instance for your **current** running addon.
From this instance you can than get the id.
```python
REAL_SETTINGS = xbmcaddon.Addon()
ADDON_ID = REAL_SETTINGS.getAddonInfo('id')
```

service.onmute/service.py
```diff
@@ -0,0 +1,215 @@
...
+    def isMute(self):
+        state = xbmc.getCondVisibility('Player.Muted') == 1
+        log("isMute = " + str(state))
+        return state
+
+        
+    def setMute(self, state):
+        log("setMute = " + str(state))
+        bailout = 0
+        while self.isMute() != bool(state) and bailout < 13:
```
Tadly: Why this loop?
I'd be better to just send the request and validate its response.

Or is there a proper reason?

service.onmute/service.py
```diff
@@ -0,0 +1,215 @@
...
+                else:
+                    self.Monitor.searchSub = self.searchSub
+                
+                '''
+                    missing subs return false positive isSubtitle = True,
+                    compare with hasSubtitle to detect true status.
+                '''
+                # check if user has subs already enabled.
+                if (self.isSubtitle() + self.hasSubtitle()) > 1 and self.isMute() == True and self.autoSub == False:
+                    continue
```
Tadly: Should there maybe be a ~1 sec wait to not stress the system?
Not sure how often this case will/can occur.

service.onmute/settings.py
```diff
@@ -0,0 +1,24 @@
...
+#
+# OnMute is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with OnMute.  If not, see <http://www.gnu.org/licenses/>.
+
+import xbmcaddon
```
Tadly: Not sure if this whole "Script to open settings" is a little overkill?
Someone else please comment on that.

In case it's fine, this...
```python
import xbmcaddon
xbmcaddon.Addon().openSettings()
```
...is really all you need

service.onmute/service.py
Lunatixz: Yes, I'm aware... it's a benign force of habit. Not a requirement?

service.onmute/service.py
Lunatixz: On paper I'd agree, but In use I've found that this condition is missable. Since the goal is to set a mute condition, it's not okay to assume it goes through. I felt a loop with a fail safe was the best option, I'm open to suggestions.

service.onmute/service.py
Lunatixz: Funny you mentioned this while using it last night I reasoned a 2sec delay would be adequate. I'll make the change :)

service.onmute/settings.py
Lunatixz: Yes :) it's a "habbit"/ code style to declare default values. You'll also note I prefer the style " if var == True:" over the short, "If var:" I really hope my minor style preferences aren't a issue. :)

service.onmute/service.py
Lunatixz: I'm going to test this again, and verify the json response.

service.onmute/service.py
Tadly: Na, not a requirement.. Just trying to get your habit away :P (No change needed unless you want to)

service.onmute/service.py
Tadly: Make sure to use the monitor for it, not xbmc.sleep ;)

Lunatixz: I changed the wait for abort value, is that incorrect?

service.onmute/settings.py
Tadly: No issue. Just optimization recommendations :)
The less source-code one has, the less error-prone it will be (and the less to review there is ;D)

service.onmute/service.py
Tadly: hm.. I've never had it fail on me which is why I was asking. (well unless I got an error back of course).

I'd appreciate you testing this again.
If there really is a case where it doesn't work (even though you got a success back) than there might be an issue within kodi itself which we than should investigate

service.onmute/settings.py
Lunatixz: True, I've spent the last four years writing plugin globals like this... Still trying to keep in mind your icon and fanart recommendations :)

service.onmute/service.py
Tadly: The reason I commented on that specific line is that the `continue` will start over your loop (not executing the rest below) hence it will **never** reach the `waitForAbort` at the end.

As I said, I don't know how often this case will actually result in `True` and call `continue`.
Assuming it can multiple times (within a few seconds or even worse for a longer period of time) you have a loop bashing away on your cpu (causing high cpu loads and on slow hardware like e.g. a raspberry it *might* even significantly slow the whole system).

service.onmute/settings.py
Tadly: I learn new things every day from the other members too :)

service.onmute/service.py
Lunatixz: Ahh you are right! Overlooked that line... still going to keep a 2sec waitforabort as well.

Tadly: Don't worry. You don't have to do it on mobile either. Just come back to it when you have time and a PC handy :)
Not like I (or others for that matter) will always have immediate time either.

service.onmute/settings.py
Lunatixz: Ditto!!

Razzeee: I also think it's a bit overkill and maybe a bit confusing for users. But if @Lunatixz really want it I'm not against it.

service.onmute/addon.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<addon id="service.onmute" version="0.0.1" name="OnMute" provider-name="Lunatixz">
+    <requires>
+        <import addon="xbmc.python" version="2.25.0"/>
+        <import addon="script.module.simplejson" version="3.3.0"/>
```
Razzeee: I don't think you use this

service.onmute/service.py
```diff
@@ -0,0 +1,216 @@
...
+# You should have received a copy of the GNU General Public License
+# along with OnMute.  If not, see <http://www.gnu.org/licenses/>.
+
+import sys, os, re, traceback, json
+import xbmc, xbmcplugin, xbmcaddon, xbmcgui
+
+# Plugin Info
+ADDON_ID = 'service.onmute'
+REAL_SETTINGS = xbmcaddon.Addon(id=ADDON_ID)
+ADDON_NAME = REAL_SETTINGS.getAddonInfo('name')
```
Razzeee: Your not using this?

service.onmute/service.py
```diff
@@ -0,0 +1,216 @@
...
+ADDON_PATH = REAL_SETTINGS.getAddonInfo('path').decode('utf-8')
```
Razzeee: Your not using this?

service.onmute/service.py
```diff
@@ -0,0 +1,216 @@
...
+ADDON_VERSION = REAL_SETTINGS.getAddonInfo('version')
+LANGUAGE = REAL_SETTINGS.getLocalizedString
```
Razzeee: Your not using this?

service.onmute/settings.py
Lunatixz: I can change it... but when you really break it down it's a style preference.  If I had a function` f(x=0)` and I wrote `x=0 f(x)` chances are you'd ignore it, rather than suggest I change it to `f(x)`. However In this case it's part of kodis api so it's drawing attention. 

If it's a real deal issue I'll gladly change it, let me know :)

service.onmute/addon.xml
Lunatixz: Is json packed with kodi?

service.onmute/service.py
Lunatixz: its used in log function

service.onmute/settings.py
Lunatixz: You guys are getting me nervous... in a few months i'm going to submit a very large script (PseudoTV Live) with over 10k+ lines of code. Go easy on me ;)

Razzeee: Well, don't be nervous. But we're trying to get you to do a good style, makes reading, maintaining and reviewing a lot easier. :)

And as said, this is just a suggestion, nothing more.

Lunatixz: I 100% understand... which is why i'm taking baby steps offloading my beta repo to Kodi.  I'm trying to get a feel for what's expected. THX again for all the help.

service.onmute/service.py
```diff
@@ -0,0 +1,203 @@
...
+                    self.Monitor.searchSub = False
+                else:
+                    self.Monitor.searchSub = self.searchSub
+                
+                '''
+                    missing subs return false positive isSubtitle = True,
+                    compare with hasSubtitle to detect true status.
+                '''
+                # check if user has subs already enabled.
+                if (self.isSubtitle() + self.hasSubtitle()) > 1 and self.isMute() == True and self.autoSub == False:
```
Razzeee: This is kinda weird.
`(self.isSubtitle() + self.hasSubtitle()) > 1`
is the same thing as
 `self.isSubtitle() == True and self.hasSubtitle() == True`
or
 `self.isSubtitle() and self.hasSubtitle()` <-- I would prefere this :)

so it why arent you using
`(self.isSubtitle() + self.hasSubtitle() + self.isMute()) > 2`?
or
 `self.isSubtitle() == True and self.hasSubtitle() == True`

Lunatixz: Is it okay we leave styling untouched?  I think its important the developer comprehend his own code. When I contribute to other projects I always try and submit the code in the devlopers style. There are a lot of ways to write the same thing.  We can agree that if a == 1, if a == True and if a all mean the same thing...

service.onmute/addon.xml
Tadly: afaik json **is** simplejson and json is packed with pythons stdlibs.
simplejson however is updated more frequently and apparently way faster, so unless you really need speed (at which point cjson might by of interest too), just stick with pythons default json module.

You might see things like...
```
try:
    import simplejson as json
except ImportError:
    import json
```
...where they basically use simplejson if available but fall back to json in case it isn't.
Again, unless you **know** you need the benefits of simplejson, there's no real reason to use it

service.onmute/service.py
```diff
@@ -0,0 +1,203 @@
...
+                else:
+                    self.Monitor.searchSub = self.searchSub
+                
+                '''
+                    missing subs return false positive isSubtitle = True,
+                    compare with hasSubtitle to detect true status.
+                '''
+                # check if user has subs already enabled.
+                if self.isSubtitle() == True and self.hasSubtitle() == True and self.isMute() == True and self.autoSub == False:
+                    xbmc.sleep(2000)
```
Tadly: what did we say about xbmc.sleep? :) :P
change that to `self.Monitor.waitForAbort(2)` and than I think we are good to go

service.onmute/service.py
Razzeee: No problem with styling at all, if it's up to me it can go in like this. But it's pretty weird if you mix compare styles in the same line, really screams to cause problems somewhen :)

service.onmute/addon.xml
Razzeee: AFAIK that's not needed anymore since v17, as the internal python implementation now is good enough? (not sure if that's the cause)

Lunatixz: Good to know thanks

service.onmute/service.py
Lunatixz: K, drop some knowledge on me @Tadly :) because i've always been confused by the new monitor class... `self.Monitor.waitForAbort(2)` What exactly does this do? in all examples it's used to sleep between loops? In this case if I was writing a normal while loop, I'd want the loop to sleep for 2secs, so is it incorrect to use waitforabort(2)? will that stall a user system?

service.onmute/service.py
Lunatixz: I agree I don't like mixing either, but for those boolean variables I was using a operator, so in my mind after that I should always view those two as a integer... Thank for understanding style, Only reason Its brought up by me is the aforementioned script I will be submitting, It would take days to rewrite it in a different style.

service.onmute/service.py
Tadly: I honestly don't know the specifics (internals) about either, `xbmc.sleep` or the `Monitor`.
However, the Monitor is a newer implementation for things like sleeping, checking whether kodi wants to exit etc. which is why the general consent is to use the over  the *legacy* stuff.
I might even (at some point) suggest internally to deprecate that stuff and see how that goes.

To answer
> `self.Monitor.waitForAbort(2)` What exactly does this do?

Exactly what the docs say it does:

>Block until abort is requested, or until timeout occurs.
>If an abort requested have already been made, return immediately. Returns True when abort have been requested, False if a timeout is given and the operation times out.

In your case it waits for 2 seconds and than continues unless e.g. the user exited kodi (well, requested to exit) than it will unblock immediately.

I hope I explained that properly :)

Lunatixz: > return immediately
IC so `self.Monitor.waitForAbort(2)` can break while it's waiting not after... So can I go crazy and set a sleep for 900 secs? would that be safe?

Tadly: Yeah sure..
I [do have services](https://github.com/linuxwhatelse/service.linuxwhatelse.notify/blob/master/service.py#L117-L120) where the mainloop (the `while not self.Monitor.abortRequested():`) itself does literally nothing apart from `self.Monitor.waitForAbort(60)` to keep the service alive.
How long you sleep is up to you. It should be a sensible value though and generally the higher the better.
How how is **to high** I do not know though.

This is also why we really dislike seeing `time.sleep` being used.
Remember, if kodi wants to exit and a service doesn't end within 5 sec. kodi **will** try to kill it which in turn prevents the service from properly *shutting down*

Lunatixz: THX, I didn't realize I could "Safely" use it like this... Now I have to re-think some existing code to take advantage of this.

PR #523, opened 2017-06-09
[script.service.tvheadend-route-latencies] 1.0.0

iwis:
### Description
Add-on checks quality of network connection between Kodi and Tvheadend by pinging some important routers and servers that are between them.

### Checklist:

Additional information :


script.service.tvheadend-route-latencies/addon.xml
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<addon id="script.service.tvheadend-route-latencies"
+       name="Tvheadend route latencies"
+       version="1.0.0"
+       provider-name="iwis">
+    <requires>
+        <import addon="xbmc.python" version="2.20.0" />
```
romanvm: It is v `2.25.0` for Krypton

script.service.tvheadend-route-latencies/default.py
```diff
@@ -0,0 +1,163 @@
...
+    total_requests = 1 + sec/wait_sec              #number of requests that will be sent
+    total_answers  = 0
+    packet_loss    = 0
+    total_time     = 0
+    s = requests.Session()
+    try:
+        s.get(url, timeout=5).text                 #make a TCP connection to speed up proper requests
+    except requests.RequestException:
+        pass
+    time.sleep(1)                                  #be nice
```
romanvm: Please use `xbmc.sleep()` here and in other cases

script.service.tvheadend-route-latencies/default.py
```diff
@@ -0,0 +1,163 @@
...
+        control.setAnimations([('WindowOpen',  'effect=fade start=0   end=100 time=300',),
+                               ('WindowClose', 'effect=fade start=100 end=0   time=300',)])
+
+    def __init__(self):
+        super(TVNetworkHealthWindow, self).__init__("Tvheadend route latencies")
+        try:  #comment it temporary to see detailed log in case of errors
+            self.testlength = int(ADDON.getSetting("testlength"))  #how long a single test will take
+
+            #--------------- IP addresses of consecutive routers and servers ---------------
+            local_router    = xbmc.getInfoLabel('Network.GatewayAddress')
```
romanvm: For some network configurations this infolabel may return incorrect data.

iwis: I noticed that on Android - xbmc.getInfoLabel('Network.GatewayAddress') returns empty string and System information > Network > Gateway is also empty in Kodi GUI. I described it in README.md. Is there another way I can get network gateway on Android?

iwis: I checked in https://github.com/xbmc/xbmc/blob/master/xbmc/network/linux/NetworkLinux.cpp that Kodi reads gateway address directly from /proc/net/route file. Unfortunately on Android this file does not contain a line for gateway now. This is why we get no info. Gateway can be read using https://pypi.python.org/pypi/netifaces/ library, but I don't know if there is possibility to use it in Kodi.

script.service.tvheadend-route-latencies/addon.xml
iwis: Fixed and tested.

script.service.tvheadend-route-latencies/default.py
iwis: Changed and tested.

script.service.tvheadend-route-latencies/default.py
romanvm: Platform and version specific wheels for `netifaces` suggest that it has binary components so it is not acceptable for a Kodi addon. But `Network.GatewayAddress` limitation is not exactly a show stopper, and the rest looks good.

PR #535, opened 2017-06-17
[script.module.pycountry] 17.5.14b

Lunatixz:
### Checklist:

Additional information :


learningit:
Still looking into this. If you want it for your own use and not wait on adding it as a pypi library, you'll need to rename it to avoid conflicts if and when it's added as a pypi library.

Lunatixz:
what does pypi library have to do with kodis repo? i'm confused... distros do not cohabitate with kodis repo, am I wrong? almost all modules hosted here are also hosted via pypi. I'm almost sure it's a non issue.

learningit:
I've discussed this with Team Kodi members, I was pretty sure but not certain of the rules. I have confirmed them with the team.
It can be an issue. Please explain what you've modified.
Definitely the version number needs to follow the pypi version or the module name needs to differ.
Generally to be accepted in the kodi.org repo, everything in the name/lib directory should be identical to the pypi version.  The version in the addon.xml should also be identical to the pypi version. 

You can have a look at an existing script.module such as:
https://github.com/xbmc/repo-scripts/tree/gotham/script.module.beautifulsoup4

Lunatixz:
Can you expand on the conflict that arises? I'm trying to understand the inner workings for Kodi and was unaware that there is a issue with script.modules vs their existing distros. I've submitted a number of modules in the past, this is the first i'm hearing of this.

Below is an example of modicatoin made to allow file access within Kodis VFS:
https://github.com/Lunatixz/XBMC_Addons/blob/master/script.module.pycountry/lib/pycountry/__init__.py#L19

I generally use the version number identical with the sources release, unless I made amendments, where I add a subversion tag 'a', 'b', etc.

THX

Lunatixz:
BTW, The only reason for submitting this module was in hope that other developers would use it, rather than individual plugin developers including hefty tables with each submissions. pycountry includes various  tables for every region in a easy to use format. Below is an example from plutotv:
https://github.com/Lunatixz/XBMC_Addons/blob/master/plugin.video.plutotv/country.py

enen92:
IIRC script modules take an higher priority than those installed on the system for platforms using system python. Usually when you package an existing python module as a script module you just take the module code (from pypi) and give credits to the original author. If you change the original module code, the module is no longer the same and so, as a means to avoid confusion, you should rename/rebrand it (if the license complies).
So, if the code has any changes, the best solution is to rename it/fork it and you to be the author.

Lunatixz:
k i'm going to rename it to pycountries and resubmit.

Issue #478, opened 2017-04-04
[Update] script.tag-generator


wellspokenman:
### Add-on details:

- **General**
  - Add-on name: Tag Generator
  - Add-on ID: script.tag-generator
  - Version number: 0.5
  - Kodi/repository version:
https://github.com/wellspokenman/tag-generator/commit/881409737f525f9c1a3cbcae5e330692aa505440
- **Code location**
  - URL:https://github.com/wellspokenman/tag-generator
  - Revision/tag: 8814097
  - Branch: master

### Description of changes:
Added support for Trakt watchlists, corrected some URLs. Dependencies for this functionality included under lib. Tested on ubuntu and windows, Kodi 16 and 17. 

### Checklist:

- [ X] My code follows the [add-on rules](http://kodi.wiki/view/Add-on_rules) and [piracy stance](http://kodi.wiki/view/Official:Forum_rules#Piracy_Policy) of this project. 
- [ X] I have read the [CONTRIBUTING](https://github.com/xbmc/repo-scripts/blob/master/CONTRIBUTING.md) document
- [ X] Each add-on submission should be a single commit with using the following style: [plugin.video.foo] v1.0.0

Additional information :
- Submitting your add-on to this specific branch makes it available to any Kodi version equal or higher than the branch name with the applicable Kodi dependencies limits.
- [add-on development](http://kodi.wiki/view/Add-on_development) wiki page.
- Kodi [pydocs](http://kodi.wiki/view/PyDocs) provide information about the Python API
- [PEP8](https://www.python.org/dev/peps/pep-0008/) codingstyle which is considered best practice but not mandatory.
- This add-on repository has automated code guideline check which could help you improve your coding. You can find the results of these check at [Codacy](https://www.codacy.com/app/Kodi/repo-plugins/dashboard). You can create your own account as well to continuously monitor your python coding before submitting to repo.
- Development questions can be asked in the [add-on development](http://forum.kodi.tv/forumdisplay.php?fid=26) section on the Kodi forum.

Razzeee:
The news in your addon.xml should only include the latest change

wellspokenman:
The wiki was not explicit about that. See https://github.com/wellspokenman/tag-generator/commit/522e01a2891e0b7c83993f399ccb1bdd3c29c343

Razzeee:
@wellspokenman thanks for letting me know, I just updated the wiki

wellspokenman:
[...]

PR #545, opened 2017-07-04
[script.tag-generator] 0.0.5

Razzeee:
@wellspokenman
https://github.com/xbmc/repo-scripts/issues/478


Razzeee:
the trakt library used is another one, not the same

wellspokenman:
- The icon doesn't have to change
- There is a good reason those modules are embedded. Requests, simplejson and oauth are required outside of Kodi by the included Trakt module. This is not the same as the Trakt addon - it does the oauth work now required after the changes to the Trakt API.

Razzeee:
Here are some improvements, but it will need to target the krypton branch, as the dependencies are only available after that version.

Modules talking to inlined modules work fine btw. I haven't had or can remember any problems.

https://github.com/wellspokenman/tag-generator/pull/2

Razzeee:
Should I update this to the newest version? And should it still be 0.5.0 then or would you do a new version  @wellspokenman ?

wellspokenman:
@Razzeee I think 0.5.1 seems appropriate. No new features for the user, so a minor change. It seems that my skillset is inadequate for contributing to the main repo, so I'll refrain from doing so until I can address that. Thank you for bridging the gap.

Razzeee:
Are you sure? I can arrage everything that's needed as it's only a new PR and a bump of the version :)

wellspokenman:
I meant, yes please and thank you to doing this one (0.5.1) - we've both put in too much time to back out now. I just meant to say that I won't be doing this again until I can be more autonomous during submission.

PR #548, opened 2017-07-07
[script.twittertweets] 1.0.0

Lunatixz:
### Checklist:

Additional information :


script.twittertweets/gui.py
```diff
@@ -0,0 +1,94 @@
...
+    if isinstance(string, list):
+        string = (string[0])
+    elif isinstance(string, (int, float, long, complex, bool)):
+        string = str(string) 
+    
+    if isinstance(string, basestring):
+        if not isinstance(string, unicode):
+            string = unicode(string, 'utf-8')
+        elif isinstance(string, unicode):
+            string = string.encode('ascii', 'ignore')
```
romanvm: Is it OK that non-ASCII chars will be swallowed?

script.twittertweets/gui.py
```diff
@@ -0,0 +1,94 @@
...
+        else:
```
romanvm: This clause will never be reached.

script.twittertweets/service.py
```diff
@@ -0,0 +1,141 @@
...
+        ''' 
+        gen. 140char mock sentence for skin test
+        '''
+        a = ''
+        for i in range(1,141):
+            a += 'W%s'%random.choice(['',' '])
+        return a[:140]
+        
+
+    def chkFEED(self, user):
```
romanvm: Maybe it's worth to use Twitter API instead of parsing html pages?

script.twittertweets/gui.py
Lunatixz: What's weird about? I'm always looking to improve my code. Function is to make sure whatever string is passed to xbmc.log is in a proper format.

script.twittertweets/service.py
Lunatixz: I took a look at the API and IMO would first require I make a proper Kodi module. Its a lot of work for what I need. As you can see parsing parameters are fairly simple. It the site changes it's HTML often I'll move to API.

script.twittertweets/gui.py
romanvm: As I said, with `.encode('ascii', 'ignore')` all non-ASCII chars will be removed. If you need a proper format for `xbmc.log()`, use `.encode('utf-8')` like in the string below that will never be reached because of a logic error in your if/elif/else construction.
Also, `xbmc.log()` does not accept non-encoded `unicode` objects, so the string above also does not make sense.
The safes option is to store all text as `unicode` and encode it with `utf-8` before passing to Kodi Python API.

Lunatixz: I will reevaluate the function... it was written hastily in-order to avoid unicode errors when I was dealing with multiple types of encodings since it's not needed for this plugin I'll remove it.

Thanks for the info...

script.twittertweets/service.py
romanvm: Maybe this will help: https://github.com/xbmc/repo-scripts/tree/krypton/script.module.twitter It's in the official repo.

Lunatixz: I didn't notice that module... THX I'll take a look for a future update. As it is I'd like to keep beautifulsoup since the script is simple and I hope requires little maintenance. I will definitely consider moving to the API for long term stability if twitter changes their site around.

romanvm:
LGTM

romanvm:
Now please rebase all commits into one

Lunatixz:
I'll need to look into how, I've been trying to figure out rebase -i ...

When I get some free time I'll play around with it, THX for the review.

Lunatixz:
I can not figure out rebase, I made of mess of this commit and will have to close and resubmit.

PR #564, opened 2017-07-29
[script.module.neverwise] v1.1.1

phate89:
### Description
Update needed for a plugin addon

### Checklist:

Additional information :

### DIFF
--- a/script.module.neverwise/addon.xml
+++ b/script.module.neverwise/addon.xml
@@ -1,16 +1,17 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<addon id="script.module.neverwise" name="NeverWise Kodi toolkit" version="1.0.9" provider-name="NeverWise">
+<addon id="script.module.neverwise" name="NeverWise Kodi toolkit" version="1.1.1" provider-name="NeverWise">
   <requires>
     <import addon="xbmc.python" version="2.14.0"/>
-    <import addon="script.module.beautifulsoup" version="3.2.1"/>
+    <import addon="script.module.beautifulsoup4" version="4.3.2"/>
+    <import addon="script.module.dateutil" version="2.4.2"/>
   </requires>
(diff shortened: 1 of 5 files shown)


script.module.neverwise/addon.xml
```diff
@@ -1,16 +1,17 @@
...
+    <import addon="script.module.dateutil" version="2.4.2"/>
   </requires>
   <extension library="lib" point="xbmc.python.module"/>
   <extension point="xbmc.addon.metadata">
     <platform>all</platform>
-    <summary lang="it">Strumenti per gli addons audio/video di Kodi.</summary>
-    <summary lang="en">Toolkit for Kodi audio/video addons.</summary>
-    <description lang="it">Questi strumenti sono utilizzati negli addons per Kodi sviluppati da NeverWise.</description>
-    <description lang="en">These toolkit are used in the addons for Kodi developed by NeverWise.</description>
+    <summary lang="it_IT">Strumenti per gli addons audio/video di Kodi.</summary>
```
enen92: I don't think this works for gotham (only helix and above). Do not forget you're submitting to the gotham branch so that is the lowest version you intend to support. However, script modules do not show anywhere in the kodi interface so this is not really a problem. Can you do a quick test in gotham to confirm there are no problems?

phate89: You're right. It's not my code so I missed this change.
I tested it with gotham and it works well (I tested it with the addon submitted to isengard). Anyway if I have to push it to helix isn't a problem since the addon that needs the update is in isengard repo.

enen92: Thank you. You don't need to have the dependencies on the same repository. The wiki is down at the moment but when it is back online you can check in (http://kodi.wiki/view/Addon.xml) which is the minimum api version supported by each kodi version. That means that a new version of kodi will support all the addons targeting lower versions of kodi. The opposite is not true tough :)
merged

PR #585, opened 2017-08-18
script.colorbox 2.0.4

BADMS:
2.0.4


script.colorbox/addon.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<addon id="script.colorbox" name="colorbox Script" version="2.0.3" provider-name="BADMS">
+<addon id="script.colorbox" name="colorbox Script" version="2.0.4" provider-name="BADMS">
```
enen92: You changed the license type so please also update the `license` tag in the addon.xml

script.colorbox/resources/lib/Utils.py
```diff
@@ -94,37 +109,6 @@ def Shuffle_Set(amount,timed=40):
         npegs = ''.join(npeg)
         HOME.clearProperty('Colorbox_shuffle.' + npegs)
         time.sleep(timed)
```
enen92: Can you please use xbmc.sleep here instead of python's time? If the time for the sleep is bigger than 5 seconds please use `xbmc.Monitor().waitforabort(timeinseconds)`

BADMS:
Ok, hopefully done, thanks.

PR #599, opened 2017-09-17
[service.addondata.sync] 0.1.1

basrieter:
### Description

This is a simple Kodi service add-on that sync (bi-directional) add-on data. Currently only Dropbox is supported but I was planning on adding more back-end systems later. Details on the syncing can be found here on my website: https://www.rieter.net/content/kodi-sync/

Perhaps a specific thing to mention is that I have built in a kill-switch feature (using a DNS TXT record) that allows me to disable all sync activity for a specific add-on version. I did this to be able to stop syncing in case a serious bug is found that could cause profile corruption. This is the DNS TXT record _addon_sync_version.rieter.net_.

### Checklist:

Additional information :


learningit:
I'm sorry to say that this add-on is not suitable for the kodi.org repo.
Nothing against the concept or coding, but we do not allow add-ons which modify the settings on other add-ons.
This type of an add-on can cause a user confusion because it can change things like usernames, passwords, tokens and other data to a state which is not easily recoverable by the end-user.
For those reasons, we are not in a position to put this add-on into the kodi.org repo.

basrieter:
That is really too bad. Would it be an option to make it opt-in by default? There are currently 2 options: full sync or selective sync. In the latter the user had to select what add-on data to sync? I would allow the user to actually actively select what to sync?

enen92:
In my opinion this is conceptually wrong from the start. This should be a service or script that exposes an interface other addons can use to sync settings. Addon developers can then opt to provide the necessary code to register in your service and to push/pull settings from your addon (which will be in sync with cloud services). If you document your module correctly it might be even possible to have a stardard implementation that makes functionality such as a context menu item (similar to this one) possible per addon. By blindingly sync addon settings by defaut and hiding it from the addon authors you're making it really easy for other addons to break without the end user knowing or even the developer (harder to provide support). This is a cool feature and a usual request however it has to be implemented properly in order not to violate our repository rules. @Razzeee points out you can use https://github.com/ruuk/script.module.addon.signals to implement a proper script for this goal. Do not take the rejection personally please.
Best regards

basrieter:
Don't worry, not taking it personally at all, just a bit disappointed, as I did not read the rules carefully enough. I really liked the idea to for this add-on. But as you describe, perhaps this is a bit too much of a sync-all method. I am really interested in looking into this and did not really looked at it from this interesting perspective. 

So if I understand correctly it would make more sense to implement a proper interface to sync data using the signals add-on. Enabling any add-on to signal some cloud storage? And to adhere to the rules, it is essential that each add-on itself stores any settings that are retrieved from the cloud using my add-on? I am a bit lost with the context menu suggestion? Would you suggest that I provide a basic Python file and some standard context menu XML entries for the addon.xml that any add-on could include to have some standard context menu entries?

enen92:
I am stating that if devs apply a constant setting to inform they support cloud sync (eg. A boolean setting named "supports-cloud-sync") it is possible to enable a context menu item in the addon listing to call your script and execute the syncing logic. Just like you do right now (iirc).

As for the rest of your comment you are correct. I would not mind at all to adapt my addons to use your module/service/script to cloud sync their settings if you provide a way to do so. Other addons will also support it as long as users request so.

basrieter:
OK, I am trying to put my thoughts in text for some about 30 minutes now, but it seems a bit of a challenge to write them down in a way they are understandable. I might be overthinking it, so please let me know. 

----------------------------

Just to be clear (and save me some unnecessary efforts):

1. **Another** add-on includes an add-on settings `supports-cloud-sync` in their settings.xml
1. That setting is used in in a visibility-condition in **my** addon.xml `<extension point="kodi.context.item">` element and makes an entry show that calls a script in **my** add-on and adds **their** add-on id to the `allowed-for-syncing` list I keep? So the only thing an add-on developer needs to do is include the setting and then the context menu item will appear because of the fact that my add-on is installed? (_If this is what you mean, how would the visibility condition look like? I could not find any condition that is based on an add-on setting?_)
1. From that point on I automatically sync all the user_data files for that add-on just as I do now (and am thus allowed to read/write in that add-ons user_data folder)?

Besides this, I could think about providing an interface based on the https://github.com/ruuk/script.module.addon.signals (or perhaps similar to it) that allows the storing of only actual add-on settings. It could provide:

* `sync_cloud_setting(setting_id, local_setting_value) -> returns synced_value`

The only thing that might be a challenge in such an implementation is that I don't really know when what setting was updated locally. I currently use the change date of a file (remote vs local) to determine what is the most recent file. I can always store the change-datetime remotely, but if somebody calls a `sync_cloud_setting`, I need to figure out whether the remote or the local setting is the most recent?

basrieter:
Ok, what I have done at the moment, is check whether an add-on has a setting `cloud_sync_enabled` set to "true" and only then I sync the add-on settings folder for that add-on. Does that make sense?

Razzeee:
I don't have much say in this, but I personally still think an addon writing cross addon folders is the last thing we want.
I would try to go with addon signals and do something like this in addon that send their settings (brace for pseudo code):

This code would be pretty c&p for everybody, but some authors likely would want some settings to NOT sync.
```
def onSettingsChanged():
 foreach Setting:
  addonSignals.SendChangedAddons

def onReceiveSettings():
  foreach Setting:
    jsonRPC/PythonApi.updateSetting (make sure to not trigger method above)
```

Your side
````
def onReceiveSettingChanged():
  cloudProvider.storeSetting

def onSyncRequest():
  addonSignals.SendSettings(cloudProvider.getSettings(addonid))
```

And yes, problem here will be figuring out when to sync what. And how to do a handshake between addons, so you can dynamically add new ones.

basrieter:
Thanks @Razzeee for your insight. Was also thinking out that, but then wondered when an add-on would call this? There is no real way of knowing when settings were updated within Kodi. This could even happen without the add-on actually running. 

Locally I could keep track of what settings I have received at what time. That would give me an indication on whether there was a local change or not. And a rough time frame in which the change occurred: between the last sync and the moment of the OnSettingsChanged event. But that could be large time span and if there were some remote changes it will be hard to figure out which change was the actual most recent one. 

But perhaps I will create a SyncClient and a SyncService. The SyncClient is something another add-on can instantiate in their add-on. As it is then allowed to read the modified time from the settings.xml file. Something like this:

```python
class SyncClient():
    def __init_(*args):
        self.settingsToSync = args
        
        # register callback (OnCloudSettingsChanged) for the addonSignals here        
        
    def UpdateSettingsInCloud():
        addon = xbmcaddon.Addon()
        
        # determine the modified date of the settings.xml file (needs a proper 
        # translate special://<path_to_addon>/settings.xml)
        lastModified = os.path.getmtime("settings.xml")
        syncData = {}
        for setting in self.settingsToSync:
            value = addon.getSetting(setting)
            syncData[setting] = value
        
        syncedSettings = addonSignals.SendSettings(syncData, lastModified)
     
    def OnCloudSettingsChanged(settings):
        # Callback method
        addon = xbmcaddon.Addon()
        for setting, value in settings:
            value = addon.setSetting(setting)        
    
```
An add-on would then only do:

```python
client = SyncClient("setting_a", "setting_b")
# at this point update for settings will automatically be received. 
# a manual update can be triggered via:
client.UpdateSettingsInCloud()
```

The only thing is that for updates to be received the add-on should be running. And many add-on will not for a long time (except services). So I might look into a request-reply system, instead of a _fire-and-forget and hope the add-on is still running whenever an callback with updated settings is received_-system. 

```python
client = SyncClient("setting_a", "setting_b")
# sync the settings with the cloud before doing anything
client.SyncSettings()
```

`SyncSettings` would include all code to read the settings, send them to the Cloud storage add-on and receive back any updated settings, which will then be stored.

Razzeee:
Haven't thought about addons not running all the time :(
Maybe it's a bit more sane to extend core then? But we need to be careful, as settings can (and sadly will) contrain sensitive data.

basrieter:
The not constant running thing was one of the reasons for my current file approach. It only works on filesystem level and can thus sync without add-ons running. With extending "the core" you mean Kodi itself, which will be in c++? Just asking, as my Python, Javascript (AngularJS) and C# are excellent, but c++ is rather lacking....

Regarding the sensitive data: I can always include encryption (I am actually doing that already in my own add-on, Retrospect).

Razzeee:
Yes, core is C++ (I'm coming from the C# world too, so it's possible ;) )

I think there are to many problems with direct access to files and there's a good reason we don't want it in the official repo. And some day (very) maybe it won't be possible anyway if all addons get sandboxed ;)

basrieter:
What I will think I will try to do is create a new `Addon` class inheriting from `xbmcaddon.Addon` that overrides the `getSetting` and `setSetting`. That way, an add-on dev could very simply do a cloud sync of settings by using the new Addon class.

* When instantiating the class, I will call 
  * the super/base `__init__`
  * the cloud service add-on to send it the current settings and receive possible updates
  * start a long-poll to get updates on changed cloud settings. 
* On the `setSetting` I call the base/super class `setSetting` to update the setting in Kodi and then call the cloud service add-on to send it the new value.
* If the long poll triggers, the new settings will be received and stored using the base/super `setSetting`. 

The cloud storage add-on basically has 3 functions then:

1. periodically sync settings to and from the cloud and keep the resulting set in a local cache
1. notify any add-on with a long poll if a setting was changed
1. receive updates from add-ons for which settings were just changed via the setSetting, store them in the cloud. Not really sure if this should be synchronous or async.

Does this make sense? I would base the calls between the add-ons on Flask and Requests as I would rather have a point-2-point connection to send all the settings.

basrieter:
@learningit @Razzeee I am reworking this and was wondering whether it would be allowed to get/set another add-ons settings using xbmcaddon.Addon("other.addon.id).getSetting and xbmcaddon.Addon("other.addon.id).setSetting? 

What I was thinking was to have an add-on register itself via either http calls so that my add-on know that it wants to sync. The other add-on will then have a "cloud_sync_enabled" and "cloud_sync_settings" setting. The latter is a comma separated list of settings to sync. From my cloud-sync add-on I would then us the xbmcaddon.Addon("other.addon.id).getSetting/setSetting to read and update that add-on setting with cloud data?

Razzeee:
Calling setSetting for another addon, will still be a problem in my eye. But we'll need to talk about this internally I guess.

basrieter:
I already thought so. So I moved that part. See here for my current implementation:

https://bitbucket.org/basrieter/service.addondata.sync/src/sync_module_3

Razzeee:
looks promissing, but I lack time for a real review.
Not too sure about the implications of having a http service.

basrieter:
That is good to hear @Razzeee. THe http service is only listening on localhost, so not available outside the Kodi service. However, the functionality is only there for communication and,  if really required, could be replaced with similar synchronous communication means. 

Who could comment on that part?

learningit:
This is still something that probably won't make it into the kodi.org repo as I understand it. A couple of reasons: as a service, it needs at least one seen using it to get into the repo; it's still a third party add-on writing into another add-ons settings; the user (not a developer) needs to be aware that something is messing with the user's settings and allow the user to deny access.

basrieter:
That is really too bad to hear. From my point of view, the add-on dev has the option to let the user disable the whole functionality via their own add-on's settings. They can make even make it an **opt-in** by set the 

```<setting id="cloud_sync_enabled" default="true" type="bool" label="00000" />```

to default to false. 

And the add-on dev's are 100% in control what settings are being synced by setting them in the `cloud_sync_settings` setting. 

And yes, in theory it is a library that a dev, willingly includes in their add-on and that library will then read/write their add-on's settings. However, 

* The add-on doing it, will respect the wishes of the add-on developer in respect to what settings are synced, 
* It will only sync settings, no files
* The fact that it is in the Kodi repository should at least give some 'confidence' that this add-on is inspected by Team Kodi before updates and thus will not abuse their data. 

So, I really have the impression that the sync feature is something that is very much appreciated by many multi-instance Kodi users (such as myself). So what would it take to get it within the acceptance criteria of Team Kodi and the official repository?

learningit:
Having another addon write settings is definitely not acceptable for the kodi.org repo.
You're kind of missing the point I raised, the end user (not the add-on dev) needs to be in control of their data.
I'm not saying that this isn't a good feature, it's just not acceptable in the kodi.org repo in its current state.

basrieter:
I understand your perspective. That is why my question was what it would require to work around this issue? Should I only provide a library that accepts and returns settings key value pairs and that add-on dev's can than code the lines to update them in their settings? 

Besides this, I still see the end-user as 100% in control at the moment: The end-user needs to setup their own dropbox account in the service, otherwise it won't work, and then tell the add-on it is allowed so sync settings. The only thing I take way is the lines of code of reading and storing the information that a user (and dev) has agreed upon to share?

basrieter:
So, out of curiosity, would be OK if I just provided my sync service add-on and post code samples on how to call the API from another add-on?

learningit:
You can always "release" the service in the kodi.org forum and see what add-ons make use of it.

basrieter:
But then a user will need to install both my service add-on and the actual add-on using it. Mine will not be simple to install, so that would not make it easy to try and win ppl over. 

But I already have a thread in the forum. I will update it and see what happens.

basrieter:
True, but then how would an end-user be installing my add-on? Downloading it from somewhere else will not really be what is within the reach of most users? 

@enen92 you initially commented on my PR, any thoughts?

learningit:
There are many, many add-ons and services in use and supported in the forum which are not in the kodi.org repo for one reason or another.

basrieter:
True, I had one myself, but I guess there won't be a lot of "dependency" add-ons that are not in the repository? Because without those dependencies in the Kodi repo, none of the official accepted add-ons using the dependency will install or work without a tech savvy person installing them?

PR #605, opened 2017-09-29
[script.module.codequick] 0.9.0

willforde:
### Description
Codequick is a framework for kodi add-on's. The goal of this framework is to simplify add-on development. This is achieved by reducing the amount of boilerplate code to a minimum, automating tasks like route dispatching and sort method selection. Ultimately allowing the developer to focus primarily on scraping content from websites and passing it to kodi.

Codequick is written to support both python2 and python3 and is fully unicode aware.

This is a replacement for script.module.xbmcutil.
All my add-ons now depend on this module.

I will soon add it to the forums for others to use, if they so wish.

### Checklist:


script.module.codequick/lib/codequick/__init__.py
```diff
@@ -0,0 +1,37 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+# Execution start time, used later to report total execution time
+from __future__ import absolute_import
+start_time = __import__("time").time()
+
+__all__ = ["run", "Script", "Route", "Resolver", "Listitem", "utils", "storage"]
```
learningit: We strongly discourage the use of \_\_label\_\_ style lables, that syntax should only be used for Python internals

script.module.codequick/lib/codequick/route.py
```diff
@@ -0,0 +1,146 @@
...
+        folder_counter = 0.0
+        for listitem in raw_listitems:
+            if listitem:
+                # noinspection PyProtectedMember
+                listitem_tuple = listitem._close()
+                listitems.append(listitem_tuple)
+                if listitem_tuple[2]:
+                    folder_counter += 1
+
+        # Guess if this directory listing is primarily a folder or video listing.
```
learningit: Not sure if this is the right way to do this. You should be setting the infolabel "mediatype" for each listitem to the correct value. I'm not sure if you're doing this because of mixed directory or you have something else in mind here.

willforde: listitem_tuple contains (path, xbmcgui.listitem, isfolder).
This checks if it is listing folders or video items, so to set xbmcplugin.setContent differently.
"files" when listing folders and "videos" when listing media items.

learningit: I can't see where you are setting the "mediatype" label which is used by newer skins.

learningit:
I ran out of time looking at this. There's a lot to review in one setting. Someone else on the team should have a look. There seems to be a fair bit of duplication with other modules we have in the repo already, so I'm not really sure what to say. I'm going to leave it to someone else to look over the youtube stuff, etc.

script.module.codequick/lib/codequick/route.py
willforde: Sorry. Wasn't sure what you were on about. To be honest I forgot about "mediatype". It would be set by the add-ons that use codequick. If needed I could set "mediatype" to default to "video", if it was not set by the add-on. I will change that bit of code to take the infolabel "mediatype" into consideration.

script.module.codequick/lib/codequick/__init__.py
romanvm: It's actually a standard way to limit wildcard imports.

learningit: My bad. Sorry about that, I thought I had gone back and deleted the comment. I realized that was valid later as I read through the code.

willforde: I already removed the __label__ you were on about.

willforde:
I have changed Route.__content_type to take "mediatype" into consideration when guessing content_type.
Also "mediatype" will default to '"video" or "music" if not set by add-on.

script.module.codequick/lib/codequick/listing.py
```diff
@@ -0,0 +1,792 @@
...
+        item.label = u"[B]%s[/B]" % Script.localize(SEARCH)
+        item.art.global_thumb("search.png")
+        item.info["plot"] = "Search for video content."
+        item.set_callback(SavedSearches, route=callback.route.path, **params)
+        return item
+
+
+# Import callback functions required for listitem constructs
+from codequick.youtube import Playlist as YTPlaylist
+from codequick.search import SavedSearches
```
romanvm: done

script.module.codequick/lib/codequick/resolver.py
```diff
@@ -0,0 +1,264 @@
...
+            raise ValueError("resolver returned invalid url of type: '%s'" % type(resolved))
+        else:
+            raise ValueError(self.localize(NO_DATA))
+
+        # Send playable listitem to kodi
+        xbmcplugin.setResolvedUrl(self.handle, True, listitem)
+
+
+# Now we can import the listing module
+from codequick.listing import Listitem
```
romanvm: done

script.module.codequick/lib/codequick/route.py
```diff
@@ -0,0 +1,165 @@
...
+
+        Any number of sort methods can be given as multiple arguments.
+        Normally this should not be needed as sort methods are auto detected.
+
+        :param int methods: One or more kodi sort methods.
+
+        .. seealso:: The full list of sort methods can be found at.\n
+                     https://codedocs.xyz/xbmc/xbmc/group__python__xbmcplugin.html#ga85b3bff796fd644fb28f87b136025f40
+        """
+        self._manual_sort.update(methods)
```
romanvm: done

script.module.codequick/lib/codequick/script.py
```diff
@@ -0,0 +1,352 @@
...
+
+    @CacheProperty
+    def profile(self):
+        """The add-on's profile data directory path."""
+        return self.get_info("profile")
+
+    @CacheProperty
+    def path(self):
+        """The add-on's directory path."""
+        return self.get_info("path")
```
romanvm: done

romanvm:
LGTM, except the fact that all this is overcomplication like using a quarry truck to bring groceries from the nearest store.

PR #655, opened 2017-11-04
Add addon-check as submodule and setup travis-ci

Razzeee:


Razzeee:
This one actually found broken png and dies on it. Not sure how to handle that.

MartijnKaijser:
Ping the author?

Razzeee:
I've prepared the needed fix for the test suite. But we should probably
also ping the author.

Razzeee:
Authors email address isn't valid (anymore). So what are we going to do? Check if it works? Mark as broken?

Rechi:
replace it with https://github.com/xbmc/repo-scripts/blob/dc0fbe652559859daaee41278d4ac5ee4f26eef1/script.service.radiovis/icon.png

PR #659, opened 2017-11-06
[script.module.clouddrive.common] 1.0.0

cguZZman:
### Description
- Module will be the base of all the cloud drive addons.
- OneDrive is the first addon ready to use this module (PR xbmc/repo-plugins#1493 pending). Google Drive, Dropbox and Box are coming.

### Checklist:


script.module.clouddrive.common/addon.xml
```diff
@@ -0,0 +1,19 @@
...
+<addon id="script.module.clouddrive.common" name="Cloud Drive Common Module"
+	version="1.0.0" provider-name="Carlos Guzman (cguZZman)">
+	<requires>
+		<import addon="xbmc.python" version="2.25.0" />
+	</requires>
+	<extension point="xbmc.python.module" library="/" />
+	<extension point="xbmc.service" library="service.py" start="login" />
+	<extension point="xbmc.addon.metadata">
+		<platform>all</platform>
+		<summary lang="en">Cloud Drive Common Module</summary>
```
Razzeee: Can you port to the new locales? This would be `en_GB` instead of `en`

script.module.clouddrive.common/resources/language/Hebrew/strings.po
```diff
@@ -0,0 +1,278 @@
+# Kodi Media Center language file
```
Razzeee: Your translations are using the old folder style, would be great if you can migrate to the new one. Here's an example:
https://github.com/xbmc/repo-scripts/tree/krypton/script.trumptweets/resources/language/resource.language.en_gb

cguZZman:
Hey guys, hope you can find some time to review the module and merge it.
If you need any change just let me know.

script.module.clouddrive.common/README.md
```diff
@@ -0,0 +1,11 @@
+# script.module.clouddrive.common
+
+
+## Usage
```
Lunatixz: Could you expand on the modules usage? Is it primarily for https://github.com/xbmc/repo-plugins/pull/1508

Documentation for other developers could be useful.

script.module.clouddrive.common/changelog.txt
```diff
@@ -0,0 +1,2 @@
+v1.0.0 - Nov 6, 2017:
+- First release
```
Lunatixz: FYI changelogs are no longer used, use the <news> addon.xml parameter.

script.module.clouddrive.common/clouddrive/common/remote/errorreport.py
```diff
@@ -0,0 +1,29 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#-------------------------------------------------------------------------------
+
+import urllib
+
+from clouddrive.common.remote.request import Request
+
+
+class ErrorReport(object):
```
Lunatixz: Sending an error report over the net should be optional. Please replace with either a user prompt to send the data or a settings.xml option to enable reporting (default=disabled).

script.module.clouddrive.common/README.md
cguZZman: It's primarily for xbmc/repo-plugins#1493 and xbmc/repo-plugins#1508. And I have 2 more coming (Dropbox and Box).
I'll add a manual later in case somebody wants to add another cloud provider. 
But is this a blocker? Hope not.

script.module.clouddrive.common/changelog.txt
cguZZman: Ok. I'll move it to the news section.

script.module.clouddrive.common/clouddrive/common/remote/errorreport.py
cguZZman: It's optional. The setting is there in settings.xml ("report_error") and the check is in https://github.com/cguZZman/repo-scripts/blob/8da551458c3a7b6ff65345d3c04647c2e346deeb/script.module.clouddrive.common/clouddrive/common/ui/addon.py#L539
It's enabled by default though. 
I'll move the check to the ErrorReport class and make it disabled by default.

script.module.clouddrive.common/README.md
Lunatixz: No it's not a block... The module should provide some data to aid any potential developers seeking to use it.... You can add one later :)

script.module.clouddrive.common/clouddrive/common/remote/errorreport.py
Lunatixz: Great, I must have missed it...

cguZZman:
@Lunatixz changes done.
Let me know.

Lunatixz:
LGTM

Lunatixz:
@MartijnKaijser requesting a revert until further evaluation of the heroku component can be made.

MartijnKaijser:
Ok

cguZZman:
Do you guys have any question I can help with?About the heroku component.

Lunatixz:
@cguZZman Heroku is the issue... It was an oversight that wasn't detected until today. It will be discussed internally and I will keep you updated. In the meantime you might consider moving authentication back to python using a more orthodox method.

cguZZman:
Just in case you guys make the revert, the addons should reverted too. They have dependency on this module.

Would you elaborate in what an orthodox method would be? 
Just clarifying thinks a little bit more, I made the login component to avoid a lot of pain to the user having to create their own app registration account and doing all the steps manually like I have seen in other OAuth2 addons.
Kodi not having a browser and the cloud drive providers not having a way to authenticate devices without a browser make the process harder.

I guess the issue is about security, so let me give you guys more information about it:
- All communication between the kodi device and the login app is over SSL. So it's encrypted.
- The pin is generated by the cryptographically strong random number generator in java: https://docs.oracle.com/javase/8/docs/api/java/security/SecureRandom.html
- The pin is a 24 bits length number, so there are ‭16,777,215‬ possible numbers. it's an in-memory only number (never stored) and lives for 10 minutes only. I can reduce it to 2.5 minutes if required.
https://github.com/cguZZman/kodi-login/blob/75101ad09b12e9f536a0c4a8ddc1e7bb1fa75e78/src/main/java/com/syncinator/kodi/login/controller/rest/PinController.java#L49
- To retrieve the information about the pin, the kodi device must provide the pin plus a random password generated by the same cryptographic generator. The password is a 1024 bits length random number.
https://github.com/cguZZman/kodi-login/blob/75101ad09b12e9f536a0c4a8ddc1e7bb1fa75e78/src/main/java/com/syncinator/kodi/login/controller/rest/PinController.java#L51
- So in short, if somebody wants to guess those 2 numbers, they would have to guess it between 2^1048 possible combinations in 10 minutes. It has the same strength as many of the current cryptography used today. But I can incremented it to 2048 if needed.
- No information is stored in the heroku server. 
- The login app it's pretty simple and straight forward, using open standards and obviously the code is there for audit. But it has a security level that my self would trust if it's hosted anywhere by anyone. 

I'm designing and change to throttling brute force attacks to improve the security. It will accept a very low number of request per seconds and blocked after a very low number of failed attempts. But as I said the current state gives me confidence.

cguZZman:
Guys, just been proactive, I made the security improvements to the login app:
- Decreased the life time of a pin to 2.5 minutes.
- Increased the random password to a 2048 bits length.
- And the best one is that now the pin info request and all the process must be done within the same network as the kodi device. So an attacker must get access to the same network first before starting to guess the 2^2048 possible passwords in less than 2.5 minutes in the very best scenario.

https://github.com/cguZZman/kodi-login/commit/5feff541476270cc8e684a663e267d0d878b8155

Also if needed, I can add you as a collaborator in heroku so you can see and audit the deployment.

enen92:
@cguZZman the problem is not the security you implement "server-side" in your heroku app but the fact you're shifting authentication to a server you own instead of giving control to the user. The existing addons have a troublesome authentication process since everything happens under the user control.
This fact was not detected during the pull request review process but imho the PR should be reverted as well as the addons using the module. In the meantime please try to create a login process in which all the required logic happens in users' system.
ping @MartijnKaijser @Razzeee @ronie

cguZZman:
I don't see the issue of having the authentication process in an owned server. This is part of OAuth2. One of the options of OAuth2. That's why the resource owner (Google Drive, OneDrive) gives the app a client_id and a secret, because a web application or server indeed can keep secret the secret, the password you were given. This is not possible when you move the authentication to the client side when the device doesn't have a browser component. And it's when all the addons have to ask the user to create their own app client_id and secret, which is one of the pains I'm trying to resolve here.
That's why the user is asked for the permission to let the app access their information (which the server app does **not**). If the user consent the access, that's it, authorization code flow done, and the app can do what it was designed to do.
So at the end it's about the trust. And that's why I keep everything open so everybody can see and trust the addon.
In fact, one of the options I have is to move the ownership of this app to you guys. I posted this message days ago: https://forum.kodi.tv/showthread.php?tid=324495. 
You could own the app or hosted the app in your server and it would be more trustful for everybody.

learningit:
It's not that this isn't clever or a reasonable solution for what you want to do. 
It's simply not acceptable to have a service in the kodi.org repo that does third party authorization without the explicit permission of the resource owners to run that service on their behalf.  Just because it behaves well today, doesn't mean that it will in the future.
I'm fairly certain I can speak for the rest of team in saying that there really isn't any desire to have kodi.org run a third-party autho service either.

This just means that this script and the associated add-ons utilizing it will not be accepted into the kodi.org repo, you can continue provide the service and add-ons from your own repo.

tamland:
How is this 'without explicit permission' and 'shifting authentication to a server'? Correct me if im wrong but if I'm reading what @cguZZman is saying correctly this is just a normal Oauth2 client doing OAuth2 flow (except the pin part). If so, there will be a consents screen where you will be asked to whether you want to grant the 'hekoku app' access to your data. Granting third party services access this way is the whole point of OAuth.

cguZZman:
@learningit If the service is moved to the kodi server, the registration would be changed too. It would be a kodi registration. Owned by the kodi team. Run by the kodi team. And ready to be used by kodi addons or third party addons.
At the end, the point I'm trying to do here is to give the users more easy options to access their media thru kodi. Without having to have a certain technical level. 

Yea, the own repo option has been always there, but I'm a fan of having the addons in the official repo. Installing an external repo is another process I always avoid (I don't have any) and certainly requires manual intervention or steps.

Anyway, thanks guys for sharing your point of view. Hope you will give this another thought. And keep the service. I'll check the options

enen92:
@cguZZman can't the heroku app (if written in python) ran on the user hardware? For example if you create a httpserver using `basehttpserver` or a web framework such as `script.module.bottle` which is also available in the kodi repo?

learningit:
@tamland Who is guaranteeing that the heroku app is valid and remains that way? just because a dev says it's ok doesn't make it so forever.

cguZZman:
@enen92 it can but it would be useless because kodi does not have a browser. The app needs a callback url which in the case of runing in the kodi device it would be "localhost". Another computer or device would not be able to get the callback because there is nothing runing in "localhost".

@learningit I understand that and that's why the option of transferring the ownership is there. If the open source and audit is not enough.

learningit:
Again, I don't have an issue with what's being done here per se. I understand the process quite well. I do have an issue about who is the trusted provider here, hence my comment about "explicit permission" of the content providers. I really don't think that kodi.org wants to become the "trusted provider". That's just my opinion, others may want to sign up for that.

enen92:
@cguZZman the webserver would be running on the localhost

cguZZman:
@enen92 yes, it would be in the localhost of the kodi device. But Google Drive or OneDrive will redirect the final step to the "localhost" of the device which is doing the navigation for the authentication. And because kodi does not have a browser, you have to do that navigation in another device, which does not have the service running in localhost.

tamland:
You guys are clearly not fully understanding what this is.. I suggest you first try it to see that there is no need for any trust here other than the user choosing to trust the service after vising Google Drive etc. and reviewing the permissions. We are definitely not banning every other service that does this on general basis. That would make several existing plugins not allowed.

The only real issue as far as I can see is that the client and url is 'kodi-login'. This is not good. You should not use the Kodi name.

cguZZman:
@tamland fair thing. I did't see that issue before. I'll make a PR right away to change the name in case you guys want to keep the addon.

Lunatixz:
@tamland There is absolutely no security allowing a outside, unverified heroku app as a middle man handling authentication.

Kodi has absolutely no control of the code running on heroku so IMO it should be excluded from the repository.

cguZZman:
@Lunatixz all OAuth2 apps are middle mans handling authentication. That's what OAuth2 is. 
IMHO, that's more secure than an app asking the user and password directly.
Only Google Drive addon has the unverified flag right now, but that's up to the user to allow it or not. Not even google is blocking the app, because the decision is a user's decision.
p.s. OneDrive app has no problem.

Lunatixz:
@cguZZman Explain to me the point of Kodi verifying plugins that redirect users to unverified code?

Your heroku app is running unverified code and your plugin's only function is to direct users to it... sure it has other functions after the fact... however its main function is a redirect to unverified code.

cguZZman:
@Lunatixz not agree at all about the function of the heroku app. It's only for authentication and that's it. To get the access token.
All operations are performed directly to the cloud provider after that. Directly from the kodi device to Google Drive or OneDrive.
Again, it's only used when an access token is needed.

Lunatixz:
@cguZZman does Kodi run the Heroku applet? or does it want to run a Heroku applet? If both answers are no that leaves us in a position where you are running code that at any time beyond the scope of Kodi can be changed... If that statement is correct you'll find yourself in the conundrum we are currently in.

basrieter:
Is there an option for OneDrive to use some API key? As Dropbox has?

cguZZman:
@Lunatixz the login app is not an applet. It's a wep application like all webapps out there.
Java is not applets if that is where it came from.
The login app It's an app like any other OAuth app. Like when Spotify uses your facebook or google account to ID you. They are not using your files tho, just your ID info.
This webapp is only to get a token, not id nor files, so the python code running in the kodi device can access the files info.

I understand the point about the code review, that's why I said I can transfer the ownership. The source code can be hosted in a git repo owned by you and anybody could create PRs like any addon.
It does not requires nothing extra than a code review like any addon and once merged it is automatically deployed by heroku or any cloud native platform or PaaS.

cguZZman:
@basrieter it is using an API key, just like Dropbox. In fact I have the dropbox addon in the queue to be added to the repo, but let's see if it make it to it.

Lunatixz:
@cguZZman it's symantecs... General point should be understood. Kodi's repository harbors code that was thoroughly vetted. In this case heroku component was missed... Heroku for all practical purposes runs code that can not be verified.

I think what you've done is novel and great, however it doesn't comply with the official repository rules.

You can still host this plug via your own private repository and remain in the Kodi forum...

tamland:
@Lunatixz There are no rules about plugins not being allowed to access third party services that runs 'unverified code'. By that logic the yoututbe plugin would have to be removed because it is too 'running unvetted and unverified code' on yoututbe.com.

tamland:
And fyi the Trakt plugin does exactly the same thing. It too is directing the user to a webpage running 'unverified code' where you need to enter a pin and login to your account.

tamland:
You can add [script.module.dropbox](https://github.com/xbmc/repo-scripts/pull/664) to the list. This too is using the DropboxOAuth2FlowNoRedirect which is basically the same thing.

tamland:
@cguZZman As you can see from that module, dropbox does have a more official way of using OAuth on devices without a browser. So does google: https://developers.google.com/identity/protocols/OAuth2ForDevices Is this usable for you? If so it would probably make people less skeptical..

learningit:
I'm a bit curious as to why we need another Dropbox or Google Drive addon in the repo when we already have existing functioning ones that seem to be reasonably well maintained?

cguZZman:
@tamland that dropbox module requires the client_id and secret. Every user would have to create an app registration to be usable, which is one of the points I'm trying to resolve here. 
And the google oauth2 for devices, I indeed reviewed that link before and if you go to the final of that page, you will see the only scope allowed for devices is the `https://www.googleapis.com/auth/drive.file` scope which gives the app permission to only files that the app has created or the user has explicitly shared with the app. The permissions required by my addon are `drive.readonly` and `drive.photos.readonly` to have read-only access to all the files and photos.
You can see what I'm saying about the scopes here: https://developers.google.com/drive/v2/web/about-auth

@learningit As I said, the dropbox addon requires you to create an app registration. 
And there is **no** other addon for Google Drive.

cguZZman:
@tamland  and team, I just renamed the login service in the meantime you reach a final decision about the module. PR: https://github.com/xbmc/repo-scripts/pull/691

learningit:
@cguZZman We are discussing this internally. I think we understand what you're doing and it raises some issues that we need to discuss and decide as a team. You are correct, there isn't a Google Drive addon in the repo.  There is one in the forum from the dmdsoftware user that's been in use for quite sometime and there are some service modules in the repo. 

Internally we need to discuss what it means to allow these types of services. We will come back to you once the discussion comes to an outcome.  The add-on(s) may be removed or marked as "broken" in the repo in the meantime, until the outcome is known.

cguZZman:
Hi Kodi Team, it's been a while, and I'm wondering if you reached a decision about my addons. There are a lot of fixes and improvements I have pending to commit. 
Hope you have some good news.
Thanks

PR #661, opened 2017-11-07
[script.module.oauth2client] 1.4.11

robweber:
### Description

Got a pull request for this that fixes some time issues affecting other addons that use this as a dependency. Didn't see any reason this couldn't stay at the current xbmc.python version level unless you'd rather I bring it up to Jarvis or Krypton. I can re-pull to those repos if needed and adjust the value. 

### Checklist:

Additional information :

### DIFF
--- a/script.module.oauth2client/lib/oauth2client/client.py
+++ b/script.module.oauth2client/lib/oauth2client/client.py
@@ -81,6 +81,8 @@ GOOGLE_APPLICATION_CREDENTIALS = 'GOOGLE_APPLICATION_CREDENTIALS'
 # The ~/.config subdirectory containing gcloud credentials. Intended
 # to be swapped out in tests.
 _CLOUDSDK_CONFIG_DIRECTORY = 'gcloud'
+# The environment variable name which can replace ~/.config if set.
+_CLOUDSDK_CONFIG_ENV_VAR = 'CLOUDSDK_CONFIG'
 
 # The error message we show users when we can't find the Application
 # Default Credentials.
@@ -530,7 +532,6 @@ class OAuth2Credentials(Credentials):
     request_orig = http.request
 
     # The closure that will replace 'httplib2.Http.request'.
-    @util.positional(1)
     def new_request(uri, method='GET', body=None, headers=None,
                     redirections=httplib2.DEFAULT_MAX_REDIRECTS,
                     connection_type=None):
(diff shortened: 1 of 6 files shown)


robweber:
Sorry for the 2 commits, I noticed after the fact that the Google version listed is actually 1.4.11

Razzeee:
While this doesn't perfectly match https://github.com/google/oauth2client/releases/tag/v1.4.11 it should be fine to merge.

But still, those are pretty old versions of the package :/

robweber:
Correct, there is one custom patch on top. There was an issue in the original Google implementation that caused issues parsing datetime objects with the expiry token. [Here is a link to the PR](https://github.com/robweber/script.module.oauth2client/pull/1) where I merged it in. 

My goal here was just to get things working for addons needing this as a dependency ASAP. I'm going to work on updating the whole thing with the newest from the Google repo.

Razzeee:
That's weird my diff to version 1.4.11 showed only two differences, none of which is the change from the PR

script.module.oauth2client/lib/oauth2client/client.py
```diff
@@ -530,7 +532,6 @@ def authorize(self, http):
     request_orig = http.request
 
     # The closure that will replace 'httplib2.Http.request'.
-    @util.positional(1)
```
Razzeee: 1.) This is still in 1.4.11 for me

script.module.oauth2client/lib/oauth2client/crypt.py
```diff
@@ -38,10 +38,37 @@ class AppIdentityError(Exception):
...
+
+  Args:
+    None
+  Returns:
+    None
+  Raises:
+    ImportError if OpenSSL is unavailable.
+
+  """
+  try:
```
Razzeee: 2.) This is how it looks in 1.4.11

````
  try:
    _ = imp.find_module('OpenSSL')
    return
  except ImportError:
    import OpenSSL
```

robweber:
Weird. Based on some time stamps I think I figured out what happened. Looks like the last time I updated from the source was June 19, 2015 (you're right, that is an old version!). Just looking at the crypt.py file, the tag for 1.14.11 says the last update was May 19, 2015. So well before I updated. If you look at the same file for [1.14.12](https://github.com/google/oauth2client/blob/v1.4.12/oauth2client/crypt.py) though the changes you're seeing are in that file, updated May 26, 2015. 

My guess is that I pulled this from the master branch of the oauth2client repo. The changes to the files were made post the 1.14.11 tag, but the version number in the version history wasn't updated yet. Mostly my mistake for not just pulling from the tagged branch. This addon is a dep for a different one I work on so I'm sure I was just in a hurry since I needed this to get it to function. 

How do you want me to proceed here. We can kill this whole thing and I can just update from the newest tagged if you want.

Razzeee:
Entirely up to you, just wanted to bring this to attention, as we prefere the modules to match upstream if possible.

MartijnKaijser:
preferred is from tagged version to exclude possible confusion

robweber:
Works for me, I'll close this myself and re-pull the most current tagged version from the source.

PR #665, opened 2017-11-12
Gotham tests config

Razzeee:
Much more features now and we're disabling three tests for gotham that we don't want to run.


.tests-config.json
```diff
@@ -0,0 +1,5 @@
+{
+    "check_legacy_language_path": false,
+    "check_legacy_strings_xml": false,
+    "check_license_file_exists": false
```
Razzeee: This is the only one of the deactivated tests that might be a discussion

PR #698, opened 2017-12-11
[script.artwork.downloader] 14.4.0.0

burekas7:
Description

* New version 14.4.0.0
* API Keys are configured now in the settings. The default keys doesn't working anymore.
* Each one should create his own key for Fanart.tv and TMDB.
* Fix Fanart.tv scraper results for Fanart and Poster types (Movies and TV shows)
* Additional fix of @JohnyBee for themoviedb.org provider ('tmdb.py'): "Artwork Downloader" was cannot cope with movies search by release date.
* Additional options for limiting downloaded artwork are visible again in Settings.
* Additional code fixes.
* More info here: https://forum.kodi.tv/showthread.php?tid=114633&pid=2676059#pid2676059

Checklist:

Additional information :

* The default API keys doesn't working anymore. Each one should create his own key for Fanart.tv and TMDB.
* Entering the API keys is done now through the addon's settings ("API Keys" category). You don't need to touch the code anymore.
* Important Note: After the installation make sure to update your own API Keys from fanart.tv (Project key only) and themoviedb.org sites (TVDB API key is already set as default, it's global and working)
* Don't use TMDB as movies scraper because you will get Provider ID error. Use Universal Movie Scraper instead.


burekas7:
What does "The Travis CI build failed" mean and what should I need to do?

Razzeee:
You might want to look at your `settings.xml` as the checker reports it as invalid xml

`PROBLEM: Invalid xml found. /home/travis/build/xbmc/repo-scripts/script.artwork.downloader/resources/settings.xml`

burekas7:
But the settings.xml file is ok and valid, I have not found something wrong there. 
The settings in kodi for this addon works well.
You can see screenshot here:
https://forum.kodi.tv/showthread.php?tid=114633&pid=2676059#pid2676059

Rechi:
The invalid syntax is the double hyphen in the following line 
https://github.com/xbmc/repo-scripts/blob/bc7e934dbdb24520dbab057a0860d5ed065b24d9/script.artwork.downloader/resources/settings.xml#L12

https://www.w3.org/TR/REC-xml/#sec-comments
> For compatibility, the string " -- " (double-hyphen) MUST NOT occur within comments.

burekas7:
@Rechi @Razzeee 
Hi, I have fixed the invalid xml issue.
Thanks.

Lunatixz:
Reviewing the changes, IMO the only needed change is the API-Keys moving to settings... Most of the additions are unnecessary (a collection of community hacks). For example https://github.com/xbmc/repo-scripts/pull/698/files#diff-17a1a78d483e38a2dd746c2cbe28290fR180 which I commented on in the forums a number of months ago unnecessary hammers the TMDB website. https://forum.kodi.tv/showthread.php?tid=114633&pid=2629171#pid2629171

burekas7:
@Lunatixz 
Basically this is not my fix, but I checked it with and without this, and it worked the same.
I also wasn't sure why we need it, but I assumed that it could only help if it doesn't make problems.

script.artwork.downloader/default.py
```diff
@@ -681,6 +681,7 @@ def _gui_mode(self, currentmedia):
                    background = setting['background'])
         # Look for argument matching artwork types
         for item in sys.argv:
+            #log('HHH    ' + item) #burekas debug
```
basrieter: Why add a commented line?

script.artwork.downloader/lib/old_apply_filters.py
```diff
@@ -0,0 +1,298 @@
+#!/usr/bin/python
```
basrieter: Is this file ever used? I don't think so, so I would suggest to just not include it?

script.artwork.downloader/lib/provider/tmdb.py
```diff
@@ -138,16 +141,47 @@ def get_image_list(self, media_id):
                 return image_list
 
 
+#def _search_movie(medianame,year=''):
```
basrieter: I would suggest to remove this? If we ever wanted to know what this method did, we can always use the GIT history?

Lunatixz:
@burekas7 its problematic, and so are many of the changes.... You want to triple the amount of api hits per movie... under the assumption that a movie the year may be incorrect, This is cumbersome process. there can only be one correct year.

I've monitored the issues with this script and IMO many of the hacks found on the forum are unnecessary and only look to fix unique issues that arise per user case.

script.artwork.downloader/lib/provider/tmdb.py
burekas7: Basically I didn't touch any of this beside the first one you mentioned that remains there due to a small debug that I was try to investigate. As I said I just fix the API Keys issue and add them to the Settings, and fix some Fanart.tv scraper issues.

burekas7:
@Lunatixz 
If you ask me, I don't care if it's there, I can remove this if you really think it's necessary. As I said, I just thought that it could be helpful.
And I see, you right about the triple amount of the API, maybe I will just remove this.
Thank you.

script.artwork.downloader/lib/provider/tmdb.py
burekas7: In addition, look about @Lunatixz comment, probably I will return this instead the new method there.

script.artwork.downloader/lib/old_apply_filters.py
burekas7: I don't familiar with this file as I wrote you.
I will check it, if it actually never used maybe I will remove it.

script.artwork.downloader/default.py
burekas7: This remained there due to a small debug that I was try to investigate about specific case (In TV Shows, when you press on DiscArt it's still looking for 'banners' as arg) . But I still haven't found the reason why. Not so critical right now.
I will remove it. Thanks.

burekas7:
Pay attention to the new commit about version 14.4.0.0.

According to the last comments I have made some code fixes.

burekas7:
@Lunatixz

Finally I have decided to leave the function with the triple check.
The problem there was about wrong indentation when I copy it.
Now if it succeeds on the first time with the current year, it won't check again for 'year-1' or 'year+1'. It will break the loop and continue with the tmbd_id that was found.

Lunatixz:
@burekas7 a indentation issue is easily fixed however as is I would reject this merge...

basrieter:
I have to agree with @Lunatixz. If you have your meta data or folder naming incorrect, then the user should fix their stuff. It should not be the scraper that tries to fix the incorrect (or accurate) information.

burekas7:
@Lunatixz 
OK. Never mind.
I will return the original function as it was.

script.artwork.downloader/resources/settings.xml
```diff
@@ -9,7 +9,7 @@
         <setting type="sep" />
         <!--
         <setting label="32023" type="bool"      id="service_startup"    default="false"/>
-        <setting label="32019" type="labelenum" id="service_startupdelay" subsetting="true" values="00|10|20|30" default="--" visible="eq(-1,true)"/>
+        <setting label="32019" type="labelenum" id="service_startupdelay" subsetting="true" values="00|10|20|30" default="40" visible="eq(-1,true)"/>
```
Lunatixz: Default is out of range from the setting, please correct... ie 00|10|20|30|**40**

Otherwise LGTM

burekas7: Done.

burekas7:
@MartijnKaijser 
When this PR going to be merged?
Thanks.

Lunatixz:
@burekas7 There are no issues with your PR, however its come to my attention this script maybe deprecated by a fork made a number of years ago: https://github.com/rmrector/script.artwork.beef

The future of artwork downloader is undecided and maybe replaced in the long term with artwork beef... The team appreciates your help maintaining the script, at this time your PR won't be merged.

burekas7:
@Lunatixz 
But artwork beef isn't even exist now in Kodi official repository, while artwork downloader is there. 
And why do not confirm the merge to be in parallel to beef? It's working good for a long time.

Lunatixz:
@burekas7 I understand, if/when artwork.beef becomes available to the repo, artwork downloader will be deprecated. IF artwork.beef doesn't merge with the repository; then we can discuss the future of artwork downloader and future PRs.

MartijnKaijser:
Sorry i don't accept any updates for this add-on unless it's one to mark it broken.

burekas7:
@MartijnKaijser 
I don't understand what do you mean?
I just add an option for insert the API keys through the settings.

PR #679, opened 2017-11-26
[service.watchedlist] 1.2.1

SchapplM:
### Checklist:

- [x] My code follows the [add-on rules](http://kodi.wiki/view/Add-on_rules) and [piracy stance](http://kodi.wiki/view/Official:Forum_rules#Piracy_Policy) of this project. 
- [x] I have read the [CONTRIBUTING](https://github.com/xbmc/repo-scripts/blob/master/CONTRIBUTING.md) document
- [x] Each add-on submission should be a single commit with using the following style: [script.foo.bar] v1.0.0

Fixed Dropbox API change: https://github.com/SchapplM/xbmc-addon-service-watchedlist/issues/30.
This Pull Request follows PR https://github.com/xbmc/repo-scripts/pull/664 from @micahg. He pushed the update to branch krypton (presumably due to stricter dependencies for the requests-addon).
This addon was on branch isengard before, but to be able to match the dropbox dependency, I also moved it to krypton. I could also try to keep it in isengard for compatibilty (all features except dropbox would keep working).
The diff should only contain the latest changes of https://github.com/SchapplM/xbmc-addon-service-watchedlist/commits/master.

PR #719, opened 2017-12-23
[script.module.buggalo] 1.1.7

SchapplM:
### Checklist:

Following the discussion in https://github.com/xbmc/repo-scripts/pull/679#discussion_r156290860, I would like to merge an update to script.module.buggalo. The addon does not seem to be maintained any more. The addon does not seem to be used by more than 3 addons, but it is in my opinion very helpful for more efficient addon development.

As described in my PR https://github.com/twinther/script.module.buggalo/pull/6 and in the new [buggalo Readme](https://github.com/SchapplM/script.module.buggalo/blob/email_ssl/README.md), for bug report emails to work, login credentials to an email server have to be placed in plain text in the source code of the addons.
Ideally, every developer would create email accounts for his addons and set a different password for the smtp authentification than for account login.

I am not sure how fast these credentials could be used by spambots, if passwords have to be changed often or if this in conflict with rules for addons (e.g. regarding safety).
I did not see any other way though to create the email notification, which is very helpful in the early phase of addon development.

### DIFF
--- a/script.module.buggalo/addon.xml
+++ b/script.module.buggalo/addon.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<addon id="script.module.buggalo" name="Buggalo Exception Collector" version="1.1.6" provider-name="twinther">
+<addon id="script.module.buggalo" name="Buggalo Exception Collector" version="1.1.7" provider-name="twinther">
     <requires>
         <import addon="xbmc.python" version="2.1.0"/>
         <import addon="script.module.simplejson" version="2.0.10"/>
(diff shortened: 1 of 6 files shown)


script.module.buggalo/addon.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<addon id="script.module.buggalo" name="Buggalo Exception Collector" version="1.1.6" provider-name="twinther">
+<addon id="script.module.buggalo" name="Buggalo Exception Collector" version="1.1.7" provider-name="twinther">
```
enen92: please add yourself as author as well

script.module.buggalo/lib/buggalo.py
```diff
@@ -122,6 +133,10 @@ def onExceptionRaised(extraData=None):
     heading = getRandomHeading()
     data = client.gatherData(etype, value, traceback, extraData, EXTRA_DATA)
 
-    d = gui.BuggaloDialog(SUBMIT_URL, GMAIL_RECIPIENT, heading, data)
+    if not GMAIL_RECIPIENT == None and EMAIL_CONFIG == None: # for backwards-compatibility
+        EMAIL_CONFIG = {'recipient':GMAIL_RECIPIENT}
+    print SUBMIT_URL
```
enen92: do not use prints, use the xbmc.log method (with debug level)

SchapplM:
I updated the requested files

PR #721, opened 2017-12-23
changed line endings from windos-style (CRLF) to linux-style (LF)

SchapplM:
When working with repo-scripts in linux, the windows file endings in this file are reported as a change in the file all the time. It is not possible to change branches without deleting the file, which is very annoying.

> error: Your local changes to the following files would be overwritten by merge:
> 	script.video.randomtv/resources/language/resource.language.en_gb/strings.po
> Please, commit your changes or stash them before you can merge.
> Aborting

I also reported this to the addon-repo in https://github.com/gmccauley/script.video.randomtv/pull/1.


enen92:
I'd wait to see if it goes upstream first. 
IMO, this update doesn't really justify a version bump in the addon.xml, It is just a question of syncing this with the repo of the author so future updates contain the updated language file.
However, I'd like to know what others think and how to proceed. ping @MartijnKaijser

Razzeee:
Our repo guidlines state that files should have linux linendings as far as I remember. And there is no version bump in this PR which should also be fine. I don't think there is any harm.

MartijnKaijser:
since changes are also upstream we merge this.

PR #734, opened 2017-12-28
[service.watchedlist] 1.2.3

SchapplM:
Please merge the latest fixes from https://github.com/SchapplM/xbmc-addon-service-watchedlist

### DIFF
--- a/service.watchedlist/lib/watchedlist/utils.py
+++ b/service.watchedlist/lib/watchedlist/utils.py
@@ -1,26 +1,31 @@
 """
 This file contains additional utility functions
 """
-import xbmc, xbmcgui, xbmcvfs, xbmcaddon
 import os
 import time
-import sys
-import buggalo
 import re
 import random
 import json
 
+import xbmc
+import xbmcgui
+import xbmcvfs
+import xbmcaddon
+import buggalo
+
 _addon_id = u'service.watchedlist'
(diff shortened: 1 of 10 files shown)


service.watchedlist/lib/watchedlist/watchedlist.py
```diff
@@ -568,6 +578,7 @@ def get_watched_xbmc(self, silent):
                           "id": 1})
                 if json_response.has_key('result') and json_response['result'] != None and json_response['result'].has_key('tvshows'):
```
romanvm: .has_key() method is deprecated and should not be used. Correct syntax is: `if key in dict`

service.watchedlist/lib/watchedlist/utils.py
```diff
@@ -65,7 +65,9 @@ def setSetting(name,value):
     _Addon.setSetting(name,value)
 
 def getSetting(name):
-    return _Addon.getSetting(name)
+    # The returned value from getSetting is not unicode, only string
+    # This function returns unicode
+    return _Addon.getSetting(name).decode('utf-8')
```
romanvm: Make sure that the setting was stored as utf-8 encoded string

SchapplM:
I implemented the requested changes (obfuscation, fix deprecated) and also let python code tools autopep8, pycodestyle and pylint search for code style violations and fixed most of them.

PR #740, opened 2018-01-01
[service.watchedlist] 1.2.4

SchapplM:
minor fix and extended exception-handling.
I got some bug report emails where `SystemExit` exception was raised.
I guess this comes from the user exiting Kodi by the shutdown menu.
However I do not understand, why the bug report dialogue can still be displayed (the whole bug report process takes a few seconds).
Also I put `monitor.waitForAbort` everywhere, so exiting Kodi should not raise an exception in python.

Is catching `SystemExit` manually the right solution here?


Lunatixz:
LGTM

PR #741, opened 2018-01-02
[script.trakt] Update Library after minimum percent of watching

burekas7:
Description

* A new option for running the Update Library after minimum percent of watching.
(Pretty similar to the "Rating" option)

* In details:
 A new category is added in settings: Update Library. 
Includes:
-- Enable/Disable Update Library after watching Movie (Default = Disabled)
-- Enable/Disable Update Library after watching TV show episodes. (Default = Disabled)
-- Minimum percent watched to run Update Library.

* Related links from the kodi forum:
https://forum.kodi.tv/showthread.php?tid=220547&pid=2681799#pid2681799
https://forum.kodi.tv/showthread.php?tid=220547&pid=2684120#pid2684120

Checklist:

Additional information :
* I have tested and verified all the cases:
1. Stopped: Below minimum percentage - Not running update library.
2. Stopped: Above minimum percentage - Is running update library.
3. Ended: Is running update library.
4. Enable Update Library after watching Movie: Is running update library.
5. Disable Update Library after watching Movie: Not running update library.
6. Enable Update Library after watching TV show episodes: Is running update library.
7. Disable Update Library after watching TV show episodes: Not running update library.

### DIFF
--- a/script.trakt/addon.xml
+++ b/script.trakt/addon.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<addon id="script.trakt" name="Trakt" version="3.1.12" provider-name="Trakt.tv, Razzeee">
+<addon id="script.trakt" name="Trakt" version="3.1.13" provider-name="Trakt.tv, Razzeee">
 	<requires>
         <import addon="xbmc.python" version="2.24.0"/>
         <import addon="script.module.simplejson" version="3.3.0"/>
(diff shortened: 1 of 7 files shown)


Razzeee:
Nice implementation, but this change is not in upstream and probably will never be, as it tries to get around a problem that needs to be fixed. The code path should not be needed at all and it adds a code smell.

burekas7:
@Razzeee 
I know that the original purpose was not sounds good.
But it could still be helpful on some cases, and as I said, the defaults are off. So I don't see any issue here.
I hope you will consider it again.

Razzeee:
I want us to fix the original problem, wherever it is.

PR #746, opened 2018-01-10
[script.common.plugin.cache] 2.5.9

Rechi:
### Description
- fixes error for windows uwp (https://forum.kodi.tv/showthread.php?tid=326796)
- some python3 adjustments

PR #756, opened 2018-01-18
[script.common.plugin.cache] 2.5.10

Rechi:
Fix `self` usage in non instance methods

@Lunatixz @rmrector should fix the issue reported at #746


Lunatixz:
LGTM

PR #820, opened 2018-03-17
[script.tubecast] 1.0.2

enen92:
minor fix

### DIFF
--- a/script.tubecast/resources/lib/tubecast/youtube/player.py
+++ b/script.tubecast/resources/lib/tubecast/youtube/player.py
@@ -35,11 +35,19 @@ class CastPlayer(xbmc.Player):
     def onPlayBackStarted(self):
         self.from_yt = False
         self.playing = True
-        self.youtubecastv1.report_playback_started(self.video_id, int(self.getTime()), self.ctt, self.list_id, self.current_index)
+
+        if not self.getTime():
+            playing_time = 0
+        else:
+            playing_time = self.getTime()
+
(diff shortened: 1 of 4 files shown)


script.tubecast/resources/lib/tubecast/youtube/player.py
```diff
@@ -35,11 +35,19 @@ def getPlayingStatusCode():
     def onPlayBackStarted(self):
         self.from_yt = False
         self.playing = True
-        self.youtubecastv1.report_playback_started(self.video_id, int(self.getTime()), self.ctt, self.list_id, self.current_index)
+
+        if not self.getTime():
```
Lunatixz: I believe getTime will raise a exception not return a boolean or nonetype.
https://codedocs.xyz/xbmc/xbmc/group__python___player.html#ga487f3e98ff5530d8292f70067fac91fe

I usually use try: getTime() except: 0

enen92: https://forum.kodi.tv/showthread.php?tid=329153&pid=2715067#pid2715067

Lunatixz: IC the log, int() requires a value; however getTime() only returns a int value or raises an exception. I just felt it was safer to cover the exception since you know to expect it.

enen92: I am not sure either. I think recently something changed and playbackstarted is triggered right when the demuxer is created. I can change it though. Looks safer as you point out. Thanks

PR #824, opened 2018-03-26
[service.subtitles.legendaszone] 0.1.3 and [service.subtitles.pipocas] 0.1.5

highlandr:
The only change is from  http -> https

Cheers


enen92:
@highlandr unfortunately you need to create two pull request (one for each addon) to keep the git history clean

highlandr:
hmmm. ok. when i get time to pick up the old stuff.
rules are rules even if they are too strict.
cheers.

PR #830, opened 2018-04-03
[service.subtitles.pipocas] 0.1.5

highlandr:
Update [service.subtitles.pipocas] 0.1.5

http to https - Mandatory.


Rechi:
https://github.com/xbmc/repo-scripts/blob/master/PULL_REQUEST_TEMPLATE.md
> Each add-on submission should be a single commit with using the following style: [script.foo.bar] v1.0.0

PR #849, opened 2018-05-02
[service.subtitles.subdivx] 0.3.2

ramiro:
### Description
Fix a race condition with a Kodi API call introduced in v0.3.0 which causes problems for many users. Also, reduce verbosity of old temporary dirs non-fatal removal failures.
### Checklist:

### DIFF
--- a/service.subtitles.subdivx/service.py
+++ b/service.subtitles.subdivx/service.py
@@ -14,7 +14,6 @@ import re
 import shutil
 import sys
 import tempfile
-import time
 from unicodedata import normalize
 from urllib import FancyURLopener, unquote, quote_plus, urlencode, quote
 from urlparse import parse_qs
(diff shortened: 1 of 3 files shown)


service.subtitles.subdivx/service.py
```diff
@@ -584,10 +591,10 @@ def main():
         # Send end of directory to XBMC
         xbmcplugin.endOfDirectory(int(sys.argv[1]))
 
-        _cleanup_tempdir(workdir)
+        time.sleep(2)
```
enen92: please do not use the time module. Use either `xbmc.sleep()` (small time steps) or preferably `xbmc.Monitor().waitforabort()`

ramiro: Thanks for the review!

I think I've adressed the issue. Followed advice from https://kodi.wiki/view/Service_add-ons

enen92: yeap, perfect. Thank you

PR #857, opened 2018-06-24
[script.trakt] 3.2.1

Razzeee:
### Description

Code wise a small update, but one specific to leia+ so it needs to be added to this branch.

### Checklist:


script.trakt/resources/settings.xml
```diff
@@ -0,0 +1,86 @@
+<?xml version="1.0" encoding="utf-8" standalone="yes"?>
```
enen92: wondering if we should not start to request developers to adopt the new settings format which is supported since leia. What do you think? It is not documented on the wiki yet (I plan to create a new page for it when I have some time) but it is quite straightforward (https://github.com/xbmc/xbmc/blob/master/system/settings/settings.xml)

enen92: Also, shouldn't we bump the python api for leia as well? There are new methods that will only work in leia and others that have been removed. However `2.25.0` will be installed (manually) in both Krypton and Leia.
@Razzeee

Razzeee: Documentation is needed and it would be even better to have some kind of automatic conversion.

Bumping sounds like a good idea, but I'm not the export on that.

PR #858, opened 2018-06-26
[script.tubecast] 1.0.7

enen92:
Got tired of the `trying to load setting definitions from old format...` on the log


enen92:
Ups, looks like this is Leia only.

PR #861, opened 2018-07-02
[script.myepisodes] 1.2.8: mark as broken

maximeh:
### Description
Mark addon as broken; a new version will be available for Krypton and up.
### Checklist:

### DIFF
--- a/script.myepisodes/addon.xml
+++ b/script.myepisodes/addon.xml
@@ -18,6 +18,7 @@
     <license>GNU GENERAL PUBLIC LICENSE. Version 2, June 1991</license>
     <source>https://github.com/maximeh/script.myepisodes</source>
     <email>maxime dot hadjinlian at gmail dot com</email>
+    <broken>Not compatible with MyEpisodes anymore. Please update Kodi and this addon</broken>
   </extension>
 </addon>
 


Rechi:
@maximeh I also found versions in gotham and helix branch. I think those should be marked broken too.

maximeh:
@Rechi okey will create a PR for each of them then.

Rechi:
Bumping gotham branch to a higher version then helix would be enough, but it's up to you.

maximeh:
@Rechi they would work with the new version of the addon ? I am wondering because of the dependencies.

Rechi:
I don't get what you ask. If an addon is in gotham branch all it dependencies has to be available in that branches. If no higher version is available in a 'higher' branch the version from lower branch is used. Btw the version numbers listed in addon.xml are min versions, so they will always match in 'higher' branches.

See https://kodi.wiki/view/Submitting_Add-ons#Allowed_submissions and https://kodi.wiki/view/Addon.xml#version_attribute_2

maximeh:
Agreed. But since the 2.0.0 version of my addon has min version for the dependencies that are only available in Krypton; it won't work if I bump the Gotham/Helix branch to use the 2.0.0 version of my addon; so I have no other choice than to mark them as broken.
Or did I misunderstood something ?

The current version of the addon is broken in gotham/helix so I can easily mark them as broken; but I can't easily upgrade them to the latest version.

Rechi:
I meant something different, you only have to mark gotham version broken if you bump the version to a higher one then currently in the helix branch. That helix version will be automatically overwritten, as version has preference over branch where the addon is in.

maximeh:
Ahhh ! I did not knew that.
Okey, that's indeed better only one PR to do instead of two :)
Thanks for the tips

PR #869, opened 2018-07-13
[script.module.xmltodict] 0.11.0

c4tz:
### Description
This PR updates script.module.xmltodict to version 0.11.0.

I tried to keep the changeset as small as possible, but removed `ez_setup.py` and `.travis.yml` as well as the tests because they are not needed inside the addon.

Tested with Kodi 17.6 on Arch Linux.

I need this update because I'm currently developing and addon which depends on the changes from 0.9.1 and 0.9.2 in xmltodict.

### Checklist:


enen92:
Thanks a lot for updating the module

PR #879, opened 2018-08-20
[script.module.clouddrive.common] 1.2.1

cguZZman:
### Description
- Auto export and watch folder
- Fixes 

### Checklist:

Additional information :


script.module.clouddrive.common/clouddrive/common/export.py
```diff
@@ -0,0 +1,110 @@
...
+    def create_strm(driveid, item, file_path, content_type, addon_url):
+        item_id = Utils.str(item['id'])
+        item_drive_id = Utils.default(Utils.get_safe_value(item, 'drive_id'), driveid)
+        f = KodiUtils.file(file_path, 'w')
+        f.write(addon_url + '?' + urllib.urlencode({'action':'play', 'content_type': content_type, 'item_driveid': item_drive_id, 'item_id': item_id, 'driveid': driveid}))
+        f.close()
+        return True
+        
+    def save_items_info(self, exportid, items_info):
+        f = KodiUtils.file(self.get_items_info_path(exportid), 'w')
```
enen92: use `with`

script.module.clouddrive.common/clouddrive/common/service/export.py
```diff
@@ -0,0 +1,245 @@
...
+                changes = changes_by_drive[driveid]
+            else:
+                self.provider.configure(self._account_manager, export['driveid'])
+                changes = self.provider.changes()
+                changes_by_drive[driveid] = changes
+            items_info = self.export_manager.get_items_info(item_id)
+            if items_info:
+                if changes and not Utils.get_safe_value(export, 'exporting', False):
+                    Logger.debug('*** Processing changes for export "%s" in %s' % (export['name'], export['destination_folder']))
+                    while True:
```
enen92: this is a forever blocking loop, use monitor to listen for requested aborts

script.module.clouddrive.common/clouddrive/common/ui/addon.py
```diff
@@ -300,11 +306,144 @@ def _list_drive(self, driveid):
...
+                url = self._addon_url + '?' + urllib.urlencode(params)
+                list_item = xbmcgui.ListItem(item_name)
+                context_options = []
+                params['action'] = '_run_export'
+                context_options.append((KodiUtils.localize(21479), 'RunPlugin('+self._addon_url + '?' + urllib.urlencode(params)+')'))
+                params['action'] = '_remove_export'
+                context_options.append((KodiUtils.localize(1210), 'RunPlugin('+self._addon_url + '?' + urllib.urlencode(params)+')'))
+                list_item.addContextMenuItems(context_options)
+                listing.append((url, list_item, True))
+        xbmcplugin.addDirectoryItems(self._addon_handle, listing, len(listing))
```
enen92: this stuff theoretically should be handled but the plugins that consume the module, not by the module itself

script.module.clouddrive.common/clouddrive/common/ui/addon.py
```diff
@@ -548,14 +614,15 @@ def play(self, driveid, item_driveid=None, item_id=None):
             for subtitle in item['subtitles']:
                 subtitles.append(DownloadServiceUtil.build_download_url(driveid, Utils.default(Utils.get_safe_value(subtitle, 'drive_id'), driveid), subtitle['id'], urllib.quote(Utils.str(subtitle['name']))))
             list_item.setSubtitles(subtitles)
-        xbmcplugin.setResolvedUrl(self._addon_handle, True, list_item)
+        if not self.cancel_operation():
+            xbmcplugin.setResolvedUrl(self._addon_handle, True, list_item)
```
enen92: same, dependency on xbmcplugin does not make sense on a script.module

script.module.clouddrive.common/resources/skins/default/1080i/export-main-dialog.xml
```diff
@@ -0,0 +1,356 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
enen92: similarly, skinning stuff is functionality of a script extension point...not a script.module

script.module.clouddrive.common/clouddrive/common/ui/addon.py
cguZZman: This module is used as common library for cloud drive plugins. 
OneDrive and Google Drive addons (and others in the future) do the same logic in almost all the things. Only the access to its respective API changes. So each addon 'inherit' all the common logic from here and then do its respective stuff.
So for me it doesn't make sense to repeat the same code over and over when a concept of 'module' already exist within the kodi ecosystem.

script.module.clouddrive.common/clouddrive/common/ui/addon.py
cguZZman: Same explanation, the addon will inherit this class.

script.module.clouddrive.common/resources/skins/default/1080i/export-main-dialog.xml
cguZZman: Same as before. All the addons needs these custom windows. Why repeat and copy the xmls in all the addons when this 'module' can abstract them.

script.module.clouddrive.common/clouddrive/common/export.py
cguZZman: Working on it

script.module.clouddrive.common/clouddrive/common/service/export.py
cguZZman: Working on it

cguZZman:
@enen92 see my comments. The idea is to abstract a lot of code for the plugins.
In some way doing the same as service.smartthings, script.module.codequick or script.module.kodiswift.

enen92:
its okay for me. Just stating it could be better, you don't need to repeat all the code, call the module to get the listitems and add the directories in your plugin...as simple as that. For the skinning stuff you're right, it's better in the module: a script module is a script anyway.

script.module.clouddrive.common/clouddrive/common/service/export.py
cguZZman: Just reviewing this 'while True'. It's not a forever blocking loop. Python doesn't have a do-while loop (just in case, no offence: it's a while but you need the code executed at least one time). 
So the code could be read as:
do {
 #bla bla
} while not (changes_done and changes_retry)

I know the while True makes a lot of noise (it did for me when you mentioned it), but reviewing it makes me remember why I put it there.
If I put a monitor.onAbortRequested() there is like puting the abort check in all whiles and fors.
Let me know

script.module.clouddrive.common/clouddrive/common/export.py
cguZZman: Did it in this way because xbmcvfs.File doesn't have the `__exit__` method where I suppose the file should be closed. 
if I use `with KodiUtils.file(file_path, 'w') as f` I'm getting the exception `AttributeError: __exit__`.
Let me know

script.module.clouddrive.common/clouddrive/common/service/export.py
enen92: ok

cguZZman:
@enen92 since `xbmcvfs.File` doesn't support the `with `statement, I implemented the try-catch version, which is ensuring the file is closed.

About calling the module to get the listitems and add the directories in the plugin that would cause the plugins to have unnecessary methods which are already abstracted in the common module. I hope this will be non-blocking issue and you can merge it.
Let me know

PR #889, opened 2018-08-31
[script.module.codequick] 0.9.5

willforde:
### Description
* Mostly documentation changes.
* Added Time to live option to persistent storage.
* Added support for fetching localized strings using "msgid" as the reference. i.e. gettext
* Added some extra functions to utils, and removed some others.
* Lots of other minor changes.
* Tested with Kodi v18 Leia.

Sorry in advanced for the large commit changes log.

### Checklist:

### DIFF
--- a/script.module.codequick/lib/codequick/storage.py
+++ b/script.module.codequick/lib/codequick/storage.py
@@ -2,7 +2,9 @@
 from __future__ import absolute_import
 
 # Standard Library Imports
+from collections import MutableMapping, MutableSequence
 from hashlib import sha1
+import time
 import sys
 import os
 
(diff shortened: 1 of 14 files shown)


script.module.codequick/lib/codequick/storage.py
```diff
@@ -89,70 +94,119 @@ def flush(self):
             self._stream.write(content)
             self._hash = current_hash
             self._stream.flush()
+        else:
```
enen92: not needed

script.module.codequick/lib/codequick/support.py
```diff
@@ -86,52 +83,93 @@ def emit(self, record):
...
+    def _parse_po(self):  # type: () -> dict
+        self._loaded = True
+
+        # Check if path needs to be translated first
+        addon_path = addon_data.getAddonInfo("path")
+        if addon_path[:10] == "special://":  # pragma: no cover
+            addon_path = xbmc.translatePath(addon_path)
+
+        # Location of strings.po file
+        strings_po = os.path.join(addon_path, "resources", "language", "resource.language.en_gb", "strings.po")
```
enen92: you're not allowed to directly parse/mess with those files. Use the API only please. I know you're trying to add support for strings instead of numeric id's but nevertheless, you need to use another approach that does not rely on reading the strings.po file directly

willforde:
I have removed the parsing code, and changed it so that the user can manually add string references using `utils.string_map`.
I also took this chance to re-order methods in storage.py to remove duplicate code.

enen92:
Thanks much for removing the strings stuff! Merging

PR #894, opened 2018-09-20
[script.module.clouddrive.common] 1.2.8

cguZZman:
### Description

- Resume points and watched status for strm files in library

### Checklist:

Additional information :

### DIFF
--- a/script.module.clouddrive.common/clouddrive/common/service/player.py
+++ b/script.module.clouddrive.common/clouddrive/common/service/player.py
@@ -16,31 +16,33 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #-------------------------------------------------------------------------------
+import datetime
+import re
+import threading
+import urllib
+
 from clouddrive.common.ui.logger import Logger
 from clouddrive.common.ui.utils import KodiUtils
-from clouddrive.common.service.source import SourceService
-import re
-import urllib
-from clouddrive.common.remote.request import Request
-from clouddrive.common.remote.errorreport import ErrorReport
-from clouddrive.common.service.download import DownloadServiceUtil
 from clouddrive.common.utils import Utils
-import threading
+
 
 class PlayerService(object):
     name = 'player'
 
     def __init__(self, provider_class):
+        from clouddrive.common.service.source import SourceService
         self.abort = False
         self._system_monitor = KodiUtils.get_system_monitor()
         self.provider = provider_class()
         self.addonid = KodiUtils.get_addon_info('id')
-        
+        if KodiUtils.get_info_label('System.BuildVersion').startswith('17.'):
+            KodiUtils.set_home_property('iskrypton', 'true')
         self.addon_name = KodiUtils.get_addon_info('name')
         self.url_pattern = 'http.*:%s/%s/%s/.*' % (KodiUtils.get_addon_setting('port_directory_listing'), SourceService.name, urllib.quote(self.addon_name))
         Logger.debug(self.url_pattern)
         self.player = KodiPlayer()
         self.player.set_source_url_matcher(re.compile(self.url_pattern))
+        self.player.set_addonid(self.addonid)
     
     def __del__(self):
         del self._system_monitor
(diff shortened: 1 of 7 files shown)


script.module.clouddrive.common/clouddrive/common/service/player.py
```diff
@@ -60,27 +62,91 @@ def stop(self):
 
 class KodiPlayer(KodiUtils.kodi_player_class()):
     def __init__(self, *args):
-        pass
```
enen92: Generally it is a bad practice to set variables on the home window. Can you use another alternative such as an hidden setting or a property on the listitem itself?

script.module.clouddrive.common/clouddrive/common/service/player.py
```diff
@@ -60,27 +62,91 @@ def stop(self):
...
+        self.iskrypton = KodiUtils.get_home_property('iskrypton') == 'true'
 
     def set_source_url_matcher(self, source_url_matcher):
         self.source_url_matcher = source_url_matcher
+    
+    def set_addonid(self, addonid):
+        self.addonid = addonid
         
     def onPlayBackStarted(self):
```
enen92: For leia please use onAVStarted

script.module.clouddrive.common/clouddrive/common/ui/addon.py
```diff
@@ -95,6 +96,7 @@ def __init__(self):
         self._system_monitor = KodiUtils.get_system_monitor()
         self._account_manager = AccountManager(self._profile_path)
         self._pin_dialog = None
+        self.iskrypton = KodiUtils.get_home_property('iskrypton') == 'true'
```
enen92: Please use our api

script.module.clouddrive.common/clouddrive/common/ui/utils.py
```diff
@@ -95,7 +98,8 @@ def execute_json_rpc(method, params=None, request_id=1):
         cmd = {'jsonrpc': '2.0', 'method': method, 'id': request_id}
         if params:
             cmd['params'] = params
-        return json.loads(xbmc.executeJSONRPC(json.dumps(cmd)))
+        cmd = json.dumps(cmd)
+        return json.loads(xbmc.executeJSONRPC(cmd))
```
enen92: This is prone to errors. Try to catch the exception

script.module.clouddrive.common/clouddrive/common/ui/utils.py
```diff
@@ -249,7 +278,90 @@ def rmdir(f, force=False):
         import xbmcvfs
         return xbmcvfs.rmdir(f, force)
     
+    @staticmethod
+    def read_content_file(file_path):
+        content = None
+        f = None
+        try:
+            f = KodiUtils.file(file_path, 'r')
```
enen92: Please use python's with statement to guarantee the file handler is closed

script.module.clouddrive.common/clouddrive/common/service/player.py
cguZZman: I'm using the home window because I didn't found a better alternative. A hidden setting would make persistent a value that I need for a short period of time only. Making it permanent in the settings will complicate the functionality.
A property in the listitem would be nice if worked, but it didn't. I set the property before calling setResolvedUrl and then try to read it with the infolabel ListItem.Property() when the playback has started, but it always returns empty. I'm guessing the listitem was changed for another instance at playing time.
So the property in the home window is the only solution it's working for me right now.

script.module.clouddrive.common/clouddrive/common/service/player.py
cguZZman: The problem is I need a callback function that works in both versions. I think I can change it sometime after leia is finally released.

script.module.clouddrive.common/clouddrive/common/ui/utils.py
cguZZman: The try-catch is always in the main method calling this util method. I don't want to eat the exception or handle it here because each method calling this one will handle it.

script.module.clouddrive.common/clouddrive/common/ui/utils.py
cguZZman: Did it in this way because xbmcvfs.File doesn't have the __exit__ method where the file should be closed. This is not python's vanilla file object. This is kodi's.
if I use with KodiUtils.file(file_path, 'r') as f I'm getting the exception AttributeError: __exit__.
Same comment as in previous pr: https://github.com/xbmc/repo-scripts/pull/879#discussion_r211298919

script.module.clouddrive.common/clouddrive/common/ui/addon.py
cguZZman: I'll change this.

cguZZman: Ok. I took a look and this is the thing: 
I'm actually using the api to set that property. 
I need it like this because I'm using this property to set the visibility of a setting:
`visible="!String.IsEmpty(window(home).property(iskrypton))"`
See the settings.xml here: https://github.com/xbmc/repo-plugins/pull/2052/files

Some settings will be visible only for krypton and others won't. 

I tried to read and use an info label directly there, but the condition `String.StartsWith(System.BuildVersion,17.)` never worked.
So I had to set that property, which uses the same infolabel and finally worked.

cguZZman:
@enen92 I responded to each comment. Let me know. Thanks

script.module.clouddrive.common/clouddrive/common/ui/utils.py
enen92: Nevermind, though it was standard file io

PR #919, opened 2018-10-23
[service.subtitles.rvm.addic7ed] 3.0.0

romanvm:
Subtitles service for Addic7ed.com. It supports only TV shows.

v.3.0.0
- Changed addon ID to avoid clashes with similar addons.
- Changed icon and fanart to generic versions.
- Converted code to Python 3 compatible.
- Code refactoring and cleanup.


service.subtitles.rvm.addic7ed/addic7ed/core.py
```diff
@@ -0,0 +1,253 @@
...
+    - thumbnailImage: a 2-letter language code (e.g. 'en') to display a country
+      flag.
+    - 'hearing_imp': if 'true' then 'CC' icon is displayed for the list item.
+    - 'sync': if 'true' then 'SYNC' icon is displayed for the list item.
+    - url: a plugin call URL for downloading selected subs.
+    """
+    for item in subs_list:
+        list_item = xbmcgui.ListItem(
+            label=item.language, label2=item.version,
+            thumbnailImage=xbmc.convertLanguage(item.language, xbmc.ISO_639_1)
```
enen92: please use `list_item.setArt()` since `thumbnailImage` is deprecated

romanvm: I'm not sure if this rule applies to subtitles API, because `thumbnailImage` here is not an actual image, but a language code, like `'en'` or `'uk'`.

romanvm: UPD: `setArt()` seems to work too. I'll do some more test and update the PR.

romanvm:
Done

enen92:
Thanks

PR #924, opened 2018-10-28
[script.module.trakt] 3.0.0

Razzeee:
### Checklist:

### DIFF
--- a/script.module.trakt/addon.xml
+++ b/script.module.trakt/addon.xml
@@ -1,19 +1,19 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
 <addon id="script.module.trakt"
        name="trakt.py"
-       version="2.14.1"
+       version="3.0.0"
        provider-name="Fuzeman, Razzeee">
   <requires>
-    <import addon="xbmc.python" version="2.1.0"/>
-	<import addon="script.module.requests" version="2.12.4"/>
-	<import addon="script.module.six" version="1.9.0"/>
-	<import addon="script.module.arrow" version="0.6.0"/>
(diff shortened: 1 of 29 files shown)


enen92:
^^ I think we should revert that change for some of the fields (email and forum at least)

MartijnKaijser:
See https://github.com/xbmc/addon-check/issues/133

Rechi:
Even if you don't care about `language`, `forum` and `email` validation, `summary` and `description` still need a `lang="en"` attribute.

Rechi:
@Razzeee as already mentioned `summary` and `description` tag still need a `lang="en"` attribute.

script.module.trakt/addon.xml
```diff
@@ -1,19 +1,19 @@
...
+    <import addon="script.module.six" version="1.9.0"/>
+    <import addon="script.module.arrow" version="0.6.0"/>
   </requires>
   <extension point="xbmc.python.module"
              library="lib" />
   <extension point="xbmc.addon.metadata">
-    <summary>Python interface for the trakt.tv API</summary>
-    <description>Packed for Kodi from https://github.com/fuzeman/trakt.py</description>
+    <summary lang="en_GB">Python interface for the trakt.tv API</summary>
+    <description lang="en_GB">Packed for Kodi from https://github.com/fuzeman/trakt.py</description>
```
Rechi: AFIK, gotham isn't compatible with language_Country identifiers.

enen92: True..missed that. This won't show on the GUI anyway.
Maybe we should have different schemas also for metadata in the addon check instead of relaxing it to a "allow all" situation

PR #981, opened 2019-01-28
[script.service.checkpreviousepisode] 0.3.3

bossanova808:
### Description

Fix for bug I introduced with previous version just released 0.3.2 (bug with Kodi versions < Leia).

Apologies @Rechi for another update so quickly but I apparently misinterpreted something the original dev did.  I'm doing this on less than ideal setups so it's a tad awkward,  but have this morning tested with both Krypton and Leia and all is working as expected.

Also, made that change to forum url as previously requested.

Slightly improved language & logging.

### Checklist:


script.service.checkpreviousepisode/addon.xml
```diff
@@ -14,7 +14,7 @@
     <platform>all</platform>
     <language></language>
     <license>GNU GENERAL PUBLIC LICENSE. Version 2, June 1991</license>
-    <forum>http://forum.xbmc.org/showthread.php?tid=151531</forum>
+    <forum>https://forum.xbmc.org/showthread.php?tid=151531</forum>
```
Rechi: You missed the important part of the requested change.
```suggestion
    <forum>https://forum.kodi.tv/showthread.php?tid=151531</forum>
```

PR #996, opened 2019-02-07
[README.md] fix broken Add-on rules link and add Kodi Addon checker

anxdpanic:


anxdpanic:
If the Kodi Addon checker link is a welcome addition I will also pr updates for repo-plugins

PR #1008, opened 2019-02-19
[script.service.checkpreviousepisode] 0.3.4

bossanova808:
### Description

v0.3.4
* Add new option to ignore particular shows
* Move 'browse show episodes' to a setting rather than on every playback interaction, which felt janky to me
* Improve language across the board
* N.B. - new dependency - script.module.yaml

**NOTES / QUESTIONS**
There is a new dependency - script.module.yaml - I am not sure when this dependency was added to the Kodi repo so it's possible that I'll need to move the target forward from Gotham for this.  Also - I presume when Kodi installs the update, it will automatically install the new dependencies?

Also - I have used all new strings in this, as the previous language was not very 'native' and a bit all over the place (and several string were not used).  Once this is out I will ask in the forum thread for translators to submit updated translations and then do a new version with those once I have collected them together.  Is that ok?

### Checklist:

### DIFF
--- a/script.service.checkpreviousepisode/addon.xml
+++ b/script.service.checkpreviousepisode/addon.xml
@@ -1,7 +1,8 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<addon id="script.service.checkpreviousepisode" name="XBMC Check Previous Episode" version="0.3.3" provider-name="Lucleonhart, bossanova808">
+<addon id="script.service.checkpreviousepisode" name="XBMC Check Previous Episode" version="0.3.4" provider-name="Lucleonhart, bossanova808">
   <requires>
     <import addon="xbmc.python" version="2.14.0"/>
+    <import addon="script.module.yaml" version="3.11.0"/>
 	<import addon="xbmc.addon" version="12.0.0"/>
   </requires>
   <extension point="xbmc.service" library="default.py" start="login"></extension>
--- a/script.service.checkpreviousepisode/changelog.txt
+++ b/script.service.checkpreviousepisode/changelog.txt
@@ -1,4 +1,9 @@
-v 0.3.3
+v0.3.4
+* Add new option to ignore particular shows
+* Move 'browse show episodes' to a setting rather than on every interaction, which felt janky to me
+* Improve language across the board
+
+v0.3.3
 * Krypton fixes for previous
 
 v0.3.2
--- a/script.service.checkpreviousepisode/default.py
+++ b/script.service.checkpreviousepisode/default.py
@@ -3,6 +3,10 @@ import xbmcaddon
 import xbmcgui
 import os
 import json
+import yaml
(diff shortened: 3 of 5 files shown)


bossanova808:
I seem to be having some issues with `script.module.yaml` side of things...I couldn't find it in the kodi scripts repo here on github in any branch (but may have missed it, and I am never sure how good githubs searching is).  In dev here I had to manually install it from the source repo...where I do see it: https://mirrors.kodi.tv/addons/leia/script.module.yaml/  

I have it working on one machine, and I can `import yaml` fine on that one.  On two other test machines, despite manually installing it in the exact same way, I can not seem to `import yaml`

Anyone have any ideas about that?  It's very odd...

Rechi:
- It's located at https://github.com/xbmc/repo-scripts/tree/gotham/script.module.yaml.
- Dependencies are only installed if you install from repo or zip.
- No idea what goes wrong without logs.

bossanova808:
Thanks @ Rechi

- Ok, requirements can stay the same then.
- I thought that was the case for dependencies.  But a manual install is just to copy it into the addons folder, right (works for every other addon, just checking it is not different for python modules)
- Full debug log here: https://www.dropbox.com/s/1dxeqh7twcgbnc2/kodi.log?dl=0

You'll see it says:
`19:56:46.956 T:7728  NOTICE: ADDON: script.module.yaml v3.11.0 installed`
...etc etc
19:56:53.210 T:3984   ERROR: EXCEPTION Thrown (PythonToCppException) : -->Python callback/script returned the following error<--
                                             - NOTE: IGNORING THIS CAN LEAD TO MEMORY LEAKS!
                                            Error Type: <type 'exceptions.ImportError'>
                                            Error Contents: No module named yaml
                                            Traceback (most recent call last):
                                              File "C:\XBMC\Leia\Kodi\portable_data\addons\script.service.checkpreviousepisode\default.py", line 7, in <module>
                                                import yaml
                                            ImportError: No module named yaml
```

The addon.xml for this is as per this pull request.

Rechi:
Have you installed your add-on via zip file? If not, the additional dependency might not have been installed correctly.
Furthermore, if you only extracted script.module.yaml to the correct place it isn't enabled and also it's dependencies (listed in addon.xml) are not installed by Kodi.

bossanova808:
@rechi - thanks again, that was it...disabled in the dependencies thing.

None of this affects the PR above...folks getting the update from the master repo (once this is merged) will then get the dependency correctly

...I have now just tested it successfully (with the manual yaml install) on another machine and all is well.  Thanks for your help, much appreciated.

bossanova808:
Actually testing it more on a systemat that already had this add-on, I note that my approach to one of the settings changes could be improved, so please hold off merging this... Will improve it tomorrow & update this PR.

bossanova808:
Ok @Rechi all updated and all seems good to go...tested and working fine here on two machines.

PR #1024, opened 2019-03-23
[script.module.inputstreamhelper] 0.3.4

emilsvennesson:
### Description
+ python2_3 compability (mediaminister, Rechi)
+ Option to disable inputstreamhelper in settings.xml
+ calculate disk space on the tmp folder (dawez)
+ Support for Unicode paths in Windows (WallyCZ)
+ Italian translation (pinoelefante)
+ Dutch translation (dnicolaas)
+ Greek translation (Twilight0)
+ Russian translation (vlmaksime)

### Checklist:

Additional information :


dagwieers:
@emilsvennesson I think by requiring `kodi-six` in this addon, every addon that is using **inputstreamhelper** also needs to add `kodi-six` in its requirements. At least on my system it was unable to find kodi_six because its path was not automatically included for the addon (not being part of the addons requirements).

That's why I would not add `kodi_six` as a requirement and simply add the needed PY2/PY3 provisions in your own code, like we did for **plugin.video.vrt.nu**.

emilsvennesson:
@dagwieers You don't need to add it as a dependency for your add-on. It will automatically get installed because the module depends on it.

I'm going to stick with kodi-six (for now at least) because it saves us the headache of dealing with the py2/3 unicode mess.

dagwieers:
Sure, it gets installed, but in my case it wasn't able to import it because the addon did not have it in its requirements. So having it installed is not sufficient, every addon needs to require it in addon.xml for the kodi_six library path to be included so the addon can import it. (I know this is weird, but I got bitten by this for **script.module.pysocks**)

dagwieers:
For **plugin.video.vrt.nu** we don't have any unicode issues by just using:
```python
from __future__ import unicode_literals
```

dagwieers:
I am just saying that I think all addons will start failing as soon as this gets out...

emilsvennesson:
That doesn't make any sense. It would mean that all add-ons would have to add the dependencies of all its modules. 

I've tried it with plugin.video.viaplay and it works just fine on all of my Kodi installs.

dagwieers:
I straced Kodi and that was what I saw. Having a library installed doesn't ensure its path will be included for python. See: https://github.com/pietje666/plugin.video.vrt.nu/pull/93#issuecomment-476287321

PR #1052, opened 2019-04-29
[script.module.urllib3] 1.25.2

Freso:
### Description

* Bump [`urllib3`](https://pypi.org/project/urllib3/) to version 1.25.2. See upstream [`CHANGES.rst`](https://github.com/urllib3/urllib3/blob/1.25.2/CHANGES.rst) for upstream changes.

This update fixes at least one bug from upstream urllib3 1.25.1 known to affect Kodi users: https://github.com/Freso/script.module.urllib3/issues/2

### Checklist:

### DIFF
--- a/script.module.urllib3/addon.xml
+++ b/script.module.urllib3/addon.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
 <addon id="script.module.urllib3"
        name="urllib3"
-       version="1.25.1"
+       version="1.25.2"
        provider-name="shazow">
   <requires>
     <import addon="xbmc.python" version="2.25.0"/>
--- a/script.module.urllib3/changelog.txt
+++ b/script.module.urllib3/changelog.txt
@@ -1,3 +1,7 @@
+[B]Version 1.25.2[/B]
+
+- urllib3 1.25.2
+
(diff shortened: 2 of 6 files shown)


basrieter:
This PR (and most likely the 1.25.1 too) really causes a very long loading time on Raspberry Pi systems. It increased from 0.5 seconds to around 4 seconds. For many video add-ons this is causing a very slow user experience.

ronie:
@Freso due to the issues 1.25.2 causes on the Raspberry Pi we are considering to revert urllib3 to 1.24.3

as far as i can tell https://nvd.nist.gov/vuln/detail/CVE-2019-11324 was already fixed in 1.24.2 (https://github.com/urllib3/urllib3/compare/a6ec68a...1efadf4), so we should be safe there.

would really appreciate your thoughts!

basrieter:
Perhaps we can include 1.24.3? I am currently trying out 1.24.3 with some of my users. I will let you know the result.

Freso:
We can downgrade to 1.24.3, yeah, if there is a way to do this. 1.24.2 is the last version to include the security flaw (so 1.24.2 doesn’t have the fix), but 1.24.3 was published which is basically just 1.24.2 + CVE fix.

I’ll make a PR for 1.24.3 in a few minutes.

PR #1054, opened 2019-05-05
[script.module.urllib3] 1.25.2.1 (1.24.3)

Freso:
### Description

* Dump (or what’s opposite of bump? :thinking:) [`urllib3`](https://pypi.org/project/urllib3/) to version 1.24.3.

Due to urllib3 v1.25.x’s inclusion of `rfc3986` it’s compiling a bunch of regexes every time `urllib3` is imported… which it is every time `requests` is imported, which is imported by a ton of Kodi add‐ons.

The best solution here would be either for `urllib3` (or `rfc3986`) to be smarter about the regex compiling, or for Kodi to not need to reimport Python libraries on every view change… However, neither of those are here‐and‐now fixes and upstream `urllib3` have released a version of 1.24.x which includes the security fix that was previously only available in the 1.25.x releases, so downgrading to 1.24.3 until this is handled better either by urllib3, its consumers, or Kodi itself.

See https://github.com/xbmc/repo-scripts/pull/1052 and https://github.com/Freso/script.module.urllib3/commit/b67a392276dd8bf04731c76fab8a240ebfad955c for more background/links.

### Checklist:


Freso:
[`filter()`](https://docs.python.org/3/library/functions.html#filter) is also available to Python 3, so I’m not going to change this code from upstream to appease @TravisBuddy…

basrieter:
@Freso I think we need to bump the version of the add-on (not urllib3 itself) to be higher than what we now have. It will make the version numbers out-of sync. Perhaps for now use 1.25.2.1?

ronie:
^ yup, what @basrieter suggest is the way to go.

we can't push addons with a lower addon version to kodi users.
the version needs to be higher than what they have currently installed.

please set version="1.25.2.1" in the addon.xml file

Freso:
Done.

Freso:
I’m partial to agree, but never the less, downgrading is an immediate bandaid that can be applied right now, while upstream figure things out. Either way, the PR is here now. I’m not strongly opinionated whether to downgrade or keep as‐is, so whatever y’all decide to do with this, I’m happy with.

romanvm:
Maybe version number inside addon.xml should be changed to `1.25.2.1.24.3` to indicate this version is a "special" one?

basrieter:
Is that a valid version number @romanvm?

romanvm:
> Is that a valid version number @romanvm?

@basrieter I don't know about any limitations in addon version strings. IIRC Kodi does basic lexicographic comparison when deciding which version is higher.

Freso:
> Maybe version number inside addon.xml should be changed to `1.25.2.1.24.3`

If people agree to this, I’m happy to oblige.

basrieter:
> If people agree to this, I’m happy to oblige.

I think it does make things more clear.

Freso:
@Rechi and @ronie?

Freso:
Looks like urllib3 1.25.3 will have a fix for this, as per https://github.com/urllib3/urllib3/issues/1590#issuecomment-491876601 – and I’m guessing they’ll put this out soon. I’m inclined to close this PR and make a new one for 1.25.3 when they release it.

basrieter:
> Looks like urllib3 1.25.3 will have a fix for this, as per [urllib3/urllib3#1590 (comment)](https://github.com/urllib3/urllib3/issues/1590#issuecomment-491876601) – and I’m guessing they’ll put this out soon. I’m inclined to close this PR and make a new one for 1.25.3 when they release it.

That would be super news!

dagwieers:
A 15% overhead on a 3500ms load time still means 500ms for loading one library.
I just checked on LibreELEC v9.0.2 and importing Requests takes 4860ms, so 15% would still be 729ms.
I'd like to see the actual results after release on Raspberry Pi.
IMO a flag to restore v1.24 behaviour would be best.

basrieter:
@dagwieers So we should wait for 1.25.3 I think!

sethmlarson:
Hey all, just dropping in to thank you all for your patience with this issue. The maintainers of urllib3 aren't commonly users of Raspberry Pi and devices like it. It would be great to get your perspectives and use-cases more frequently. Feel free to drop in with suggestions or PRs with how things can be improved in the future and potentially how these types of issues can be avoided in the future.

Rechi:
Upstream has fixed the slow import the correct way, so no need for a downgrade.

basrieter:
Great! @fresco would it be possible to get version 1.25.3 into the repo? I see it was release a week ago (https://pypi.org/project/urllib3/#history)?

Freso:
1.25.3 PR: https://github.com/xbmc/repo-scripts/pull/1071

dagwieers:
Importing urllib3 v1.25.3 takes about 800ms on RPi, whereas importing both urllib and urllib2 together only takes 180ms. So IMO this release is still an unacceptable regression on RPi that slows down almost every add-on needlessly.

Freso:
`urllib` (Python 3) and `urllib2` (Python 2) are Python standard library modules, not 3rd party modules. Comparing `urllib3` to these is a false and misleading equivalency.

That said, I’m open to reopening this PR (or making a new one) for 1.25.3.1.24.3 if the Kodi maintainers say this should be the way.

dagwieers:
@Freso The purpose is not to mislead, but to give a reference so that you can compare numbers coming from the same device. On the upstream ticket (https://github.com/urllib3/urllib3/issues/1590#issuecomment-498887591) I also added this as a reference:

> In comparison, urllib3 v1.24.1 takes only 260ms to import, which makes v1.25.3 about 3x slower to be imported.

In the meantime I am investigating and making changes to the regular expressions to make them perform better. Will need to wait until this evening to test the effect on a Raspberry Pi. However, I think we either need a way to delay the re.compile(), or a way to disable the RFC validation as I doubt we can make big improvements to the regexps to avoid the overhead of re.compile().

PR #1055, opened 2019-05-07
[script.service.checkpreviousepisode] 0.3.5

bossanova808:
### Description

* Fixed error with non ascii chars in show title
* Better addon profile path handling, create profile directory if doesn't exists
* (Thanks to Andrew Beam)

### Checklist:


bossanova808:
Hmm, so that doesn't work...not sure how to do PRs to two addons at once...will jsut do them one by one I guess!

PR #1063, opened 2019-05-20
[script.module.requests] 2.22.0

Freso:
### Description

* Update `requests` module to 2.22.0. See upstream [release history](http://docs.python-requests.org/en/master/community/updates/) for upstream changes.

**Note:** `requests` 2.22.0 drops official support for Python 3.4. I don’t know what the minimal supported Python version for Kodi M* will be. If this is going to be Python 3.4(.x), then don’t merge this. If it’ll be Python 3.5 or later, then this should be fine.

### Checklist:


basrieter:
Ok, from the changelog of urllib3 I read:

```Requests now supports urllib3 v1.25.2. (note: 1.25.0 and 1.25.1 are incompatible)```

However, in the addon.xml the dependency is still on 1.22?

Freso:
[The documentation](https://kodi.wiki/view/Addon.xml#.3Crequires.3E) states:
> The dependency must be provided with the minimum version number your script/skin requires.

The minimum required version of `urllib3` for `requests` 2.22.0 is [1.21.1](https://github.com/kennethreitz/requests/blob/v2.22.0/setup.py#L47), so as I understand it there’s no need to bump the `version` attributes. I could downgrade it, but I also don’t want to risk downgrading to a version that may have never existed.

Of course, I may also be misreading/misunderstanding the `addon.xml` documentation.

basrieter:
The thing is, with 1.21.1 as min-version, it is also OK to have 1.22 installed, which is no longer compatible?

Freso:
The statement/declaration "`'urllib3>=1.21.1,<1.26,!=1.25.0,!=1.25.1'`" says that `requests` (2.22.0) is compatible with "all versions of `urllib3` starting from (and including) version 1.21.1 and prior to (and thus excluding) version 1.26, except for versions 1.25.0 and 1.25.1". This means that using `urllib3` 1.22.x will(/should, anyway!) still work 100% fine with `requests`. Prior to 2.22.0, that line read `'urllib3>=1.21.1,<1.25'` and as such theoretically it wasn’t compatible with 1.25.x at all, which is the main (only?) thing that was changed in 1.22.0 compared to 1.21.0.  (1.25 had no backwards incompatible changes that made `requests` break, so even though `requests` didn’t declare support for 1.25, it’s been working just fine regardless. The only change any users may notice is that `requests` will not warn about mismatched dependency versions (which, again, is just that: a warning—no deterioration of functionality…).

basrieter:
Ok, so we are good there. Great!

PR #1071, opened 2019-06-03
[script.module.urllib3] 1.25.3

Freso:
### Description

* Bump [`urllib3`](https://pypi.org/project/urllib3/) to version 1.25.3. See upstream [`CHANGES.rst`](https://github.com/urllib3/urllib3/blob/1.25.3/CHANGES.rst) for upstream changes.

This update reduced the initiation time when `import`ing the `urllib3` module, which especially affects Kodi users on low‐end platforms like Raspberry Pis. See also https://github.com/xbmc/repo-scripts/pull/1054

### Checklist:

### DIFF
--- a/script.module.urllib3/LICENSE.txt
+++ b/script.module.urllib3/LICENSE.txt
@@ -1,19 +1,21 @@
-This is the MIT license: http://www.opensource.org/licenses/mit-license.php
+MIT License
 
-Copyright 2008-2016 Andrey Petrov and contributors (see CONTRIBUTORS.txt)
+Copyright (c) 2008-2019 Andrey Petrov and contributors (see CONTRIBUTORS.txt)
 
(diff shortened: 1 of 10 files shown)


basrieter:
Can we merge this @Rechi @ronie ?

dagwieers:
Importing urllib3 v1.25.3 takes about 800ms on RPi, whereas importing both urllib and urllib2 together only takes 180ms. So IMO this release is still an unacceptable regression on RPi that slows down almost every add-on needlessly.

Rechi:
@dagwieers you have to report that to upstream (https://github.com/urllib3/urllib3). The Kodi repo only uses the official releases of this library.

basrieter:
I think the most import thing is that it's performance improved significantly with this version and is again comparable with 1.24.x. Can you confirm that?

That it is slower than just urllib or urllib3 is something that might have always been.

PR #1073, opened 2019-06-04
[script.module.kodi-six] 0.1.0

romanvm:
### Add-on details:

- **General**
  - Add-on name: Kodi Six
  - Add-on ID: script.module.kodi-six
  - Version number: 0.1.0
  - Kodi/repository version: gotham

- **Code location**
  - URL: https://https://github.com/romanvm/kodi.six
  
Wrappers around Kodi Python API that normalize handling of textual and byte strings in Python 2 and 3.

### Description of changes:

0.1.0:
- Implemented lazy patching of Kodi API functions and classes.

### Checklist:

### DIFF
--- a/script.module.kodi-six/addon.xml
+++ b/script.module.kodi-six/addon.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
 <addon id="script.module.kodi-six"
        name="Kodi Six"
-       version="0.0.2"
+       version="0.1.0"
        provider-name="Roman V.M.">
   <requires>
     <import addon="xbmc.python" version="2.14.0"/>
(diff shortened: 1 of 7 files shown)


romanvm:
Since this version includes important performance optimization, I decided to push it to `gotham` as well.

Rechi:
@romanvm Are you sure that you want to release this as version `0.1.0` too. So you never know which version one actually has.
Also keep in mind that it is not guaranteed that the Leia version is published to the Leia and higher repos in this case.

romanvm:
I want to publish this version in all branches that are still open for updates because it includes an important performance fix, especially for less performant platforms like Pi.
What is the best way to do it? Maybe assign lower version number than in Leia but higher than here, e.g. 0.0.5?

Rechi:
Everything higher then `0.0.2` and lower then `0.1.0` is fine. To be able to exactly identify the used version I wouldn't use `0.0.3` as that was already used in Leia branch.

romanvm:
Ok, thanks. Then I'm closing this PR and will create a new one.

PR #1084, opened 2019-06-28
[script.module.cache] 1.0.0

FraserChapman:
### Description

Simple Python HTTP Cache using sqlite3

* Stores cached data in sqlite3.Blobs.
* Calculates lifetime, freshness as virtual columns.
* Obeys cache control directives, immutable, no-store, etc.
* Supports etag, last_modified, etc for conditional requests.
* Can be used as a generic key/blob data store when used without directives.

### Checklist:


FraserChapman:
I realise that "No addon is using it" applies here - however I have a PR for repo-addons/leia for a new add-on that does use it - should issue the PR in the next hour or so...

FraserChapman:
Just removed .gitignore

script.module.cache/lib/cache.py
```diff
@@ -0,0 +1,311 @@
...
+try:
+    import _pickle as cpickle
+except ImportError:
+    import cPickle as cpickle
+
+logger = logging.getLogger(__name__)
+
+HTTP_DATE_FORMAT = "%a, %d %b %Y %H:%M:%S %Z"
+ADDON = xbmcaddon.Addon()
+ADDON_PROFILE = ADDON.getAddonInfo("profile")
```
romanvm: Should be decoded into UTF-8 in Python 2

script.module.cache/lib/cache.py
```diff
@@ -0,0 +1,311 @@
...
+CACHE_URI = xbmc.translatePath(path.join(ADDON_PROFILE, "cache.sqlite"))
+
+
+def httpdate_to_datetime(input_date):
+    # type: (str) -> Union[datetime, None]
```
romanvm: For function/methods that may return `None` you should better use `-> Optional[datetime]`

script.module.cache/lib/cache.py
```diff
@@ -0,0 +1,311 @@
...
+        # type: (list) -> None
+        """Saves stored data"""
+        with Cache(self.db) as c:
+            c.set(self.key, json.dumps(data, ensure_ascii=False))
+
+    def append(self, data):
+        # type: (str) -> None
+        """Add string to the store if it doesn't exist"""
+        current = self.retrieve()
+        if data not in current:
```
romanvm: If strings in the list are unique, consider using a set instead of a list. Presence check in a set is an O(1) operation vs O(n) in a list. The only problem with set is that it does not maintain the element's order.

script.module.cache/lib/cache.py
```diff
@@ -0,0 +1,311 @@
...
+        self.connection = None
+        self.db = CACHE_URI if db is None else db
+        if not xbmcvfs.exists(self.db):
+            xbmcvfs.mkdirs(path.dirname(self.db))
+        self._open(self.db)
+
+    def get(self, uri):
+        # type: (str) -> Union[None, sqlite3.Row]
+        """Retrieve a partial entry from the cache"""
+        query = ("SELECT blob, last_modified, etag, immutable, "
```
romanvm: Consider using multiline strings `"""..."""` for SQL queries.

script.module.cache/lib/cache.py
```diff
@@ -0,0 +1,311 @@
...
+        # type: (str, tuple) -> sqlite3.cursor
+        if values is None:
+            values = ()
+        try:
+            # Automatically commits or rolls back on exception
+            with self.connection:
+                return self.connection.execute(query, values)
+        except (sqlite3.IntegrityError, sqlite3.OperationalError) as e:
+            logger.debug(e.message)
+            return
```
romanvm: `return` is not needed here.

script.module.cache/lib/cache.py
FraserChapman: Done

script.module.cache/lib/cache.py
FraserChapman: Done

script.module.cache/lib/cache.py
FraserChapman: Ah my thinking was sets aren't supported by JSON...so a list kind of made sense as it can be passed to json.loads/json.dumps ... but thinking about it now - the data doesn't need serialising further at all as a pickled blob of a set is fine...indeed this removes the dependency on json too - done and done :)

script.module.cache/lib/cache.py
FraserChapman: Done

script.module.cache/lib/cache.py
FraserChapman: Done

FraserChapman:
Thanks! Have made all the changes suggested - thanks for the review :)

FraserChapman:
Just updated the comments and readme to reflect the changes from review.

FraserChapman:
Ah the decode("utf-8") ... think that is it.

FraserChapman:
Small fix in debug message

script.module.cache/lib/cache.py
```diff
@@ -0,0 +1,307 @@
...
+            self.connection.cursor().close()
+            self.connection.close()
+
+    def _create_table(self):
+        # type: () -> None
+        query = """CREATE TABLE IF NOT EXISTS data (
+                uri TEXT PRIMARY KEY NOT NULL,
+                blob BLOB NOT NULL,
+                http_date TIMESTAMP NOT NULL,
+                age INTEGER DEFAULT 0,"
```
Rechi: invalids SQL syntax (`"`)

FraserChapman: Ah thanks - fixed.

script.module.cache/lib/cache.py
```diff
@@ -0,0 +1,307 @@
...
+        """Saves set of strings"""
+        with Cache(self.db) as c:
+            if type(data) is set:
+                c.set(self.key, data)
+
+    def append(self, item):
+        # type: (str) -> None
+        """Add string to the store"""
+        current = self.retrieve()
+        current.add(item)
```
Rechi: Another Python 3 incompatibility added in yesterdays changes.
```
  File "<...>\addons\script.module.cache\lib\cache.py", line 102, in append
    current.add(item)
AttributeError: 'str' object has no attribute 'add'
```

FraserChapman: Pretty sure that is because of a cached string value, "current' should be of type Set...I think clearing the cache would fix this. In any case I have amended the retrieve method to ensure the value is a set. Done

FraserChapman:
Ah - just made the other query strings use multiline strings """ too :)

PR #1108, opened 2019-08-01
[script.module.enum34] 1.1.6

jmooremcc:
### Description

An enumeration is a set of symbolic names (members) bound to unique, constant values
This module is the backported version of the enum34 package that runs under Python 2.7
Do not use this module with Python versions greater than 3.3

### Checklist:

Additional information :


enen92:
Why can't this add-on be used with versions higher than 3.3? Kodi matrix will bundle python 3.7.3

jmooremcc:
> Why can't this add-on be used with versions higher than 3.3? Kodi matrix will bundle python 3.7.3

The reason it shouldn't be used with Python versions higher than 3.3 is because starting with version  3.4, enum became part of the standard library. Using this module on versions higher than 3.3 will generate errors which I have personally experienced.

Rechi:
Create a PR with the add-on using this module.
We will merge this PR only shortly before merging the add-on using this.

jmooremcc:
> Create a PR with the add-on using this module.
> We will merge this PR only shortly before merging the add-on using this.

Done. PR #1114

dagwieers:
@Rechi We need this module also for **script.module.netflix**, and are looking to remove it from the codebase and add it as a dependency. What is keeping this from being merged?

enen92:
@dagwieers script modules need at least one add-on in the repo upon which they depend (or ready to be accepted). The add-on the module depends on is still being reviewed/adjusted. Please check this is the last version of the module and that it works under python 3.7 if you plan to use it.

dagwieers:
@enen92 It is the last version, I had it packaged almost identically. The Netflix add-on is not Python 3 ready, we are working on that. Hopefully this week we will be finished and have something to propose.

jmooremcc:
> @Rechi We need this module also for **plugin.video.netflix** (which we are preparing to include in repo-plugins), and are looking to remove it from the codebase and add it as a dependency. What is keeping this from being merged?
> 
> PS I actually started to create a package to be added here (https://github.com/dagwieers/script.module.enum34), when I noticed this was already proposed.

@dagwieers you should be aware that the enum34 module should not be used under Python 3.3+. The reason why is because the enum module is already included in the standard library and using the module will create runtime errors.

dagwieers:
@jmooremcc I know.

dagwieers:
@jmooremcc Wouldn't it be easier to call this package `enum34` so we can do:
```python
try:  # Python 3
    from enum import Enum
except ImportError:
    from enum34 import Enum
```
Or is there an easier way to avoid this conflict?

jmooremcc:
> @jmooremcc Wouldn't it be easier to call this package `enum34` so we can do:
> 
> ```python
> try:  # Python 3
>     from enum import Enum
> except ImportError:
>     from enum34 import Enum
> ```
> 
> Or is there an easier way to avoid this conflict?

That might work. 
However, my plan for the next version of Kodi is to simply remove the enum34 module from my addon.xml file as a required module. This will eliminate the enum34 module from being loaded and all of my enum import statements will work without modification since they will then refer to the standard library.

dagwieers:
@jmooremcc We implemented the above and ship enum34 as enum34 as part of the plugin. There's no other feasible solution if script.module.enum34 does not ship as enum34 :-( This current submission will never work in practice. Can you make this modification?

jmooremcc:
@dagwieers I've actually run my code under V19 Kodi that's running Python Version 3.7 and all I had to do was remove the requirement for the script.module.enum34 module from my addon.xml file. All of my modules that utilize the enum module have the simple import statement:
>`from enum import Enum`

When I include enum34 module as a requirement and installed the module, I got the following error in the log file:

> EXCEPTION Thrown (PythonToCppException) : -->Python callback/script returned the following error<--
                                             - NOTE: IGNORING THIS CAN LEAD TO MEMORY LEAKS!
                                            Error Type: <class 'AttributeError'>
                                            Error Contents: module 'enum' has no attribute 'IntFlag'

This proves that using the enum34 module under Python 3.7 creates a conflict between the standard library version and the enum34  module.

One other thing I found out was that the nightly build I downloaded and installed for V19-Alpha1 on 9/13/19 does not appear to be running Python 3.7. It appears to be running Python 2.7. The version I'm running that does contain Pyton 3.7 is "KodiSetup-20190605-baee55d4-PR16116-merge-x64" which I acquired back on 6/25/19. 

I also want to ask @enen92  or @Rechi  if they know the xbmc.python version number for Kodi Matrix (V19)? I haven't been able to find that information so far.

jmooremcc:
@dagwieers One other note: When I uninstall the enum34 module when running the 9/13/19 nightly build, I get an error in the log file that says the enum module could not be found. This is further evidence that that the 9/13/19 nightly build is not running Python 3.7.

dagwieers:
@jmooremcc I don't know about you, but we intend to have a single plugin to work on both Python 2 *and* Python 3. And in that case you cannot offer enum34 as enum, because it will fail on Python 3. So leaving it as an optional requirement will make the plugin fail on Python 2 because it requires enum34, and potentially will break on Python 3 if it does get installed. So that is clearly not an option. So maybe close this request, and do as you please. We settled for shipping it with the plugin (as enum34).

dagwieers:
And yes, the nightly builds don't use Python 3 yet. And we shouldn't rely on that anyway.

So *pretty please* change this module to use enum34 as package name, or close this request, because the request as-is does not make any sense.

enen92:
Closing as this looks stalled.

jmooremcc:
Hi,
Originally this module was required by PR 1172 so that I would not have to make this module as part of my addon.
I don’t know why it’s stalled but once I remove enum34 from my addon code, I will need this module.

I seem to be caught in a circular trap. On the one hand this module is refused because no addon requires it.
Then on the other hand, my addon (PR 1172) is refused because the enum34 module does not exist in the repository.

So I am requesting that this module be added to the repository so that my addon (PR 1172) can use it.

Thank you.

John

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

dagwieers:
@jmooremcc This module failed to work because it was packaged as `enum` and as I have repeatedly told you this breaks Python3 (because this enum34 fails to work on Python 3.5+). As a result any add-on that requires it (e.g. on Python 2) will fail to work on Python 3.

jmooremcc:
> @jmooremcc This module failed to work because it was packaged as `enum` and as I have repeatedly told you this breaks Python3 (because this enum34 fails to work on Python 3.5+). As a result any add-on that requires it (e.g. on Python 2) will fail to work on Python 3.

This module is only to be run under Python 2. It is not needed for Python 3.5+. And the module does work under Leia which still runs under Python 2.
The next version of my add-on will not require the module since enum is included in Python 3.5+.

dagwieers:
> This module is only to be run under Python 2. It is not needed for Python 3.5+.

@jmooremcc Sure, and how do you plan to make an add-on that works on both Python 2 and Python 3? Not the way you packaged it, that is why you need to package it as `enum34`. Else you break Python 3.

jmooremcc:
> > This module is only to be run under Python 2. It is not needed for Python 3.5+.
> 
> @jmooremcc Sure, and how do you plan to make an add-on that works on both Python 2 and Python 3? Not the way you packaged it, that is why you need to package it as `enum34`. Else you break Python 3.

For Python 3, the only thing I will need to do is remove the reference to the enum module from my addon.xml file. I won’t have to modify any other code in any other module because it will automatically access the version of enum in Python 3.

dagwieers:
@jmooremcc That means that the way you package this **forces** everyone that wants to use your enum34 package to have two different ZIP files, one for Python 2 and one for Python 3.5+.

That is completely ridiculous. Most (if not all) add-ons now work on both Python 2 and Python 3 without the need to distribute different ZIP files. And there is no easy way to distribute these add-ons through Kodi repositories, because the Python versions is not tied to a specific Kodi release.

Kodi Matrix could ship/run on an operating system that is Python 2 or Python 3.

So I do not understand why you would not make the proposed obvious change to accomodate Kodi add-ons, but insist on a more complex, unworkable and theoretical use case.

dagwieers:
I am afraid you lack the practical experience with add-ons that work on both Python 2 and Python 3. Or supporting Kodi add-ons that support Krypton, Leia and Matrix on top of all kinds of operating systems.

basrieter:
Regarding our discussion here: https://github.com/xbmc/repo-scripts/pull/1172

You still have the `lib/enum/` path as the main library path. That won't work. As Dagwieers suggested: make it `lib/enum/` so that users can do:

```python
try:  # Python 3
    from enum import Enum
except ImportError:
    from enum34 import Enum
```
So if the `enum`-module is not present in the Python version it will fall back to the Kodi module.

jmooremcc:
> Regarding our discussion here: #1172
> 
> You still have the `lib/enum/` path as the main library path. That won't work. As Dagwieers suggested: make it `lib/enum/` so that users can do:
> 
> ```python
> try:  # Python 3
>     from enum import Enum
> except ImportError:
>     from enum34 import Enum
> ```
> 
> So if the `enum`-module is not present in the Python version it will fall back to the Kodi module.

I'm confused. 
In your comment it's confusing when you said "You still have the `lib/enum/` path as the main library path. That won't work. As Dagwieers suggested: make it `lib/enum/` ". This line does not change anything. Can you be more specific about what you want me to do.
Thank you.

basrieter:
My bad: make it lib/enum34/

jmooremcc:
> My bad: make it `lib/enum34/`

I hate to appear so dense mentally but where exactly should I make that change?
I renamed the directory from enum to enum34. I checked the addon.xml file and the only thing I see is '<extension library="lib" point="xbmc.python.module" />'. 
I noticed that somehow a second directory named enum appeared under lib and I have deleted it.
Please give me specific directions on what to change and where to change it.
Thank you in advance for your patience and your help.

basrieter:
Did you push those changes? The code here seems OK on first sight, just the `lib/enum/` to `lib/enum34/` rename should happen. Commit (amend so it stays a single commit) and then force push the changes.

basrieter:
It could be you need to create a new PR for this, as I can't open this one anymore.

jmooremcc:
> It could be you need to create a new PR for this, as I can't open this one anymore.

New PR #1334

PR #1131, opened 2019-08-23
[script.module.dropbox] 8.4.2

micahg:
### Description
* Fix Windows import errors (@robweber)
* Make qrcode module an optional dependency (@robweber)
### Checklist:

### DIFF
--- a/script.module.dropbox/addon.xml
+++ b/script.module.dropbox/addon.xml
@@ -1,33 +1,36 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
 <addon id="script.module.dropbox"
        name="Dropbox API"
-       version="8.4.1"
-       provider-name="micah">
+       version="8.4.2"
+       provider-name="Dropbox">
   <requires>
     <import addon="xbmc.python" version="2.25.0"/>
     <import addon="script.module.requests" version="2.18.4"/>
-    <import addon="script.module.qrcode" version="5.2"/> <!-- For QR Code display for Dropbox API authentication -->    
+    <import addon="script.module.six" version="1.11.0" />
+    <import addon="script.module.qrcode" version="5.2" optional="true" /> <!-- For QR Code display for Dropbox API authentication -->
   </requires>
-  <extension point="xbmc.python.library" library="authenticate.py"> <!-- entry to start the authentification script from the settings dialog of other addons -->
-    <provides> </provides> <!-- Addon should not be shown anywhere -->
-  </extension>
+  <extension point="xbmc.python.library" library="authenticate.py" /> <!-- entry to start the authentification script from the settings dialog of other addons -->
   <extension point="xbmc.python.module" library="lib" /> <!-- entry to import this in other addons -->
   <extension point="xbmc.addon.metadata">
-    <summary lang="en">Dropbox API</summary>        
-    <description lang="en">Provide Dropbox Python API support</description>
-    <description lang="he">התוסף מספק תמיכה ב-API Python Dropbox</description>
-    <disclaimer lang="en"></disclaimer>     
+    <summary lang="en_GB">Dropbox API</summary>
+    <description lang="en_GB">Provide Dropbox Python API support</description>
+    <description lang="he_IL">התוסף מספק תמיכה ב-API Python Dropbox</description>
     <platform>all</platform>
     <language/>
     <license>Apache2</license>
-    <forum></forum>
-    <website></website>
-    <source>http://python-requests.org</source>
+    <website>https://www.dropbox.com/developers/documentation/python</website>
+    <source>https://github.com/dropbox/dropbox-sdk-python</source>
     <email>micahgalizia AT gmail.com</email>
-    <news>v8.4.1 (2017-11-28)
+    <news>
+v8.4.2 (2019-08-23)
+- Fix Windows import errors (github.com/robweber)
+- Make qrcode module an optional dependency (github.com/robweber)
+
+v8.4.1 (2017-11-28)
 - Update to API version 8.4.0
 - Switch strings to po format
 - Add Hebrew (github.com/Eng2Heb)
-- Fix for UI lib issues (github.com/SchapplM)</news>
+- Fix for UI lib issues (github.com/SchapplM)
+    </news>
   </extension>
 </addon>
--- a/script.module.dropbox/lib/dropbox/pkg_resources.py
+++ b/script.module.dropbox/lib/dropbox/pkg_resources.py
@@ -1,0 +1,3 @@
+
+def resource_filename(package_or_requirement, resource_name):
+    return resource_name
--- a/script.module.dropbox/lib/dropbox/dropbox.py
+++ b/script.module.dropbox/lib/dropbox/dropbox.py
@@ -6,7 +6,7 @@ __all__ = [
 
 # This should always be 0.0.0 in master. Only update this after tagging
 # before release.
-__version__ = '8.4.0'
+__version__ = '8.4.1'
 
 import contextlib
 import json
--- a/script.module.dropbox/lib/dropbox/session.py
+++ b/script.module.dropbox/lib/dropbox/session.py
@@ -41,7 +41,6 @@ class _SSLAdapter(HTTPAdapter):
             block=block,
             cert_reqs=ssl.CERT_REQUIRED,
             ca_certs=_TRUSTED_CERT_FILE,
-            ssl_version=ssl.PROTOCOL_TLSv1,
         )
 
 def pinned_session(pool_maxsize=8):


micahg:
Hi, I _think_ the errors picked out by @TravisBuddy are all in the dropbox source (eg: not my code). Are there any issues I _must_ fix to get this merged?

Thanks,

Rechi:
The error is reported for a non upstream file. Warnings and infos are just warnings and infos and do not fail the build.

micahg:
yeah sorry -- i was misreading @TravisBuddy

script.module.dropbox/addon.xml
```diff
@@ -1,33 +1,36 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
 <addon id="script.module.dropbox"
        name="Dropbox API"
-       version="8.4.1"
+       version="8.4.2"
        provider-name="micah">
```
Rechi: Official author of the Dropbox module is `Dropbox`. https://github.com/dropbox/dropbox-sdk-python/blob/v8.4.1/setup.py#L59.

script.module.dropbox/addon.xml
```diff
@@ -1,33 +1,36 @@
...
   <requires>
     <import addon="xbmc.python" version="2.25.0"/>
     <import addon="script.module.requests" version="2.18.4"/>
```
Rechi: upstream requires `requests >= 2.16.2`

script.module.dropbox/addon.xml
```diff
@@ -1,33 +1,36 @@
...
-    <import addon="script.module.qrcode" version="5.2"/> <!-- For QR Code display for Dropbox API authentication -->    
+    <import addon="script.module.six" version="1.11.0" />
```
Rechi: upstream requires `six >= 1.3.0`

script.module.dropbox/addon.xml
```diff
@@ -1,33 +1,36 @@
...
+    <import addon="script.module.qrcode" version="5.2" optional="true" /> <!-- For QR Code display for Dropbox API authentication -->
   </requires>
-  <extension point="xbmc.python.library" library="authenticate.py"> <!-- entry to start the authentification script from the settings dialog of other addons -->
-    <provides> </provides> <!-- Addon should not be shown anywhere -->
-  </extension>
+  <extension point="xbmc.python.library" library="authenticate.py" /> <!-- entry to start the authentification script from the settings dialog of other addons -->
   <extension point="xbmc.python.module" library="lib" /> <!-- entry to import this in other addons -->
   <extension point="xbmc.addon.metadata">
     <summary lang="en">Dropbox API</summary>
```
Rechi: `lang="en_GB"`

script.module.dropbox/addon.xml
```diff
@@ -1,33 +1,36 @@
...
+    <import addon="script.module.qrcode" version="5.2" optional="true" /> <!-- For QR Code display for Dropbox API authentication -->
   </requires>
-  <extension point="xbmc.python.library" library="authenticate.py"> <!-- entry to start the authentification script from the settings dialog of other addons -->
-    <provides> </provides> <!-- Addon should not be shown anywhere -->
-  </extension>
+  <extension point="xbmc.python.library" library="authenticate.py" /> <!-- entry to start the authentification script from the settings dialog of other addons -->
   <extension point="xbmc.python.module" library="lib" /> <!-- entry to import this in other addons -->
   <extension point="xbmc.addon.metadata">
     <summary lang="en">Dropbox API</summary>        
     <description lang="en">Provide Dropbox Python API support</description>
```
Rechi: `lang="en_GB"`

script.module.dropbox/addon.xml
```diff
@@ -1,33 +1,36 @@
...
     <description lang="he">התוסף מספק תמיכה ב-API Python Dropbox</description>
```
Rechi: `lang="en_IL"`

script.module.dropbox/addon.xml
```diff
@@ -1,33 +1,36 @@
...
-    <disclaimer lang="en"></disclaimer>     
     <platform>all</platform>
     <language/>
     <license>Apache2</license>
-    <forum></forum>
-    <website></website>
-    <source>http://python-requests.org</source>
+    <website>https://www.dropbox.com</website>
```
Rechi:
```suggestion
    <website>https://www.dropbox.com/developers/documentation/python</website>
```

script.module.dropbox/addon.xml
```diff
@@ -1,33 +1,36 @@
...
+    <source>https://www.dropbox.com/developers/documentation/python</source>
```
Rechi:
```suggestion
    <source>https://github.com/dropbox/dropbox-sdk-python</source>
```

script.module.dropbox/addon.xml
```diff
@@ -1,33 +1,36 @@
...
     <email>micahgalizia AT gmail.com</email>
```
Rechi: upstream email is `dev-platform@dropbox.com`

script.module.dropbox/lib/dropbox/pkg_resources.py
```diff
@@ -0,0 +1,3 @@
+
+def resource_filename(*args):
```
Rechi: If you want to provide `pkg_resources.resource_filename(...)`, this one has wrong arguments.
Correct is `def resource_filename(package_or_requirement, resource_name):`

script.module.dropbox/lib/dropbox/pkg_resources.py
```diff
@@ -0,0 +1,3 @@
...
+    return "trusted-certs.crt"
```
Rechi: Why hardcode the return value, instead of implementing it the correct way?

robweber: These changes came from me. These additions came from [some advice given in the forum](https://forum.kodi.tv/showthread.php?tid=143406&pid=1221401#pid1221401). The dropbox library throws an error when pkg_resources is not present and this was the suggested work around to giving the library what it wanted without a lot of extra work. This function, returning the cert location, is all it cares about and gets the library to load and function. I've been using this workaround as part of the Backup addon for years.

Rechi: I had the following in mind, which isn't a lot of extra code, but will also work if additional `pkg_resources.resource_filename` calls are added later.
```
def resource_filename(package_or_requirement, resource_name):
    return resource_name
```

robweber: Makes sense to me. @micahg - do you want me to make this change and do another PR on your repo so you can merge this in? I see there are other changes to consider from above as well. Just let me know if you need more from me.

micahg:
Hey,  I'm good to push up the requested changes -- just want some clarification on the versions above.

Thanks,

script.module.dropbox/addon.xml
Rechi: It's not a blocker, but I see no reason to not use the upstream requirements.

script.module.dropbox/addon.xml
Rechi: Usually I would set this to the upstream email, or do not provide the email tag. If you want to keep your own email here, it's fine for me.

micahg:
Hi @Rechi, if its ok with you, I'll leave the versions where they were. I think I've addressed everything else.

Thanks,

script.module.dropbox/addon.xml
Rechi: still needs to be changed

Rechi:
> Version 8.4.2 should contain upstream code of version 8.4.1 (https://github.com/dropbox/dropbox-sdk-python/releases/tag/v8.4.1). (There is no upstream 8.4.2 version).

still missing

> Each add-on submission should be a single commit with using the following style: [script.foo.bar] 1.0.0

please squash the commits

micahg:
Ah, right 8.4.2 -- how would you prefer me to version it? I can't rerelease 8.4.1 and I don't plan on moving to version 9.

Will squash once the version is sorted.

Rechi:
You can name it 8.4.2, but it has to contain upstream code 8.4.1. script.module.dropbox 8.4.1 also didn't contain 8.4.1 upstream code.

micahg:
Hello @Rechi -- apologies for taking so long to get back to this.

There were actually just a few minor changes between 8.4.0 and 8.4.1. They're merged in here now.

script.module.dropbox/addon.xml
Rechi: change is still missing

micahg:
OK, sorry I missed that he is now he_IL. Hopefully that's everything now.

micahg:
Thanks for sticking it out @Rechi. Also, thanks for the patches @robweber

PR #1134, opened 2019-08-27
[script.embuary.helper] 1.2.22

sualfred:
### Description

* copy skin nodes function
* get path stats function
* fixed unwanted logging

### Checklist:

Additional information :

### DIFF
--- a/script.embuary.helper/addon.xml
+++ b/script.embuary.helper/addon.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<addon id="script.embuary.helper" name="Embuary Helper" version="1.2.20" provider-name="sualfred">
+<addon id="script.embuary.helper" name="Embuary Helper" version="1.2.22" provider-name="sualfred">
 	<requires>
 		<import addon="xbmc.python" version="2.26.0"/>
 		<import addon="script.module.pil" version="1.1.7"/>
--- a/script.embuary.helper/resources/lib/helper.py
+++ b/script.embuary.helper/resources/lib/helper.py
@@ -38,6 +38,19 @@ def get_kodiversion():
     return int(build[:2])
 
 
+def addon_setting(skin,setting,save=False):
+    profile = xbmc.getInfoLabel('System.ProfileName')
+    setting_id = skin + '_' + profile + '_' + setting
+    skin_version = xbmcaddon.Addon(skin).getAddonInfo('version')
+
+    if not save:
+        if ADDON.getSetting(id=setting_id) == skin_version:
+            return True
+        return False
+    else:
+        ADDON.setSetting(id=setting_id, value=skin_version)
+
+
 def log(txt,loglevel=NOTICE,force=False):
     if ((loglevel == NOTICE or loglevel == WARNING) and ADDON.getSettingBool('log')) or loglevel == DEBUG or force:
 
@@ -60,6 +73,8 @@ def remove_quotes(label):
         label = label[1:-1]
         if label.startswith('"') and label.endswith('"') and len(label) > 2:
             label = label[1:-1]
+        elif label.startswith('&quot;') and label.endswith('&quot;'):
+            label = label[6:-6]
 
     return label
 
@@ -72,8 +87,8 @@ def get_date(date_time):
 
 
 def execute(cmd):
-    log('Execute: %s' % cmd)
-    xbmc.executebuiltin(cmd, DEBUG)
+    log('Execute: %s' % cmd, DEBUG)
+    xbmc.executebuiltin(cmd)
 
 
 def visible(condition):
--- a/script.embuary.helper/resources/lib/plugin_content.py
+++ b/script.embuary.helper/resources/lib/plugin_content.py
@@ -842,6 +842,66 @@ class PluginContent(object):
                         self.li.append((li_path, li_item, False))
 
 
+    def getpathstats(self):
+        path = remove_quotes(self.params.get('path'))
+        prop_prefix = self.params.get('prefix','Stats')
+
+        played = 0
+        numitems = 0
+        inprogress = 0
+        episodes = 0
+        watchedepisodes = 0
+        tvshowscount = 0
+        tvshows = []
+
+        if 'activatewindow' in path.lower() and '://' in path and ',' in path:
+            path = path.split(',')[1]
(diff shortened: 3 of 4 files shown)


script.embuary.helper/resources/lib/utils.py
```diff
@@ -127,6 +127,26 @@ def textviewer(params):
     DIALOG.textviewer(headertxt, bodytxt)
 
 
+def copyskinnodes(params):
```
enen92: Don't think this is allowed

sualfred: Why shouldn't this be allowed? The library node editor does exactly the same. And the wiki is showing a howto how to create own nodes.
The problem is, that .xml nodes aren't working outside of the library path.

sualfred: If it's a problem that I move them to
* %USERDATA%/library/%SKINNAME%/node.xml
instead of
* %USERDATA%/library/video/%SKINNAME%/node.xml
then I could change that. I only thought this would be an much more organized and clearer to users.

Background context why skinners are using direct xml nodes in same cases -> .xml nodes overwrite the window history. Otherwise a home button link "all tv show titles" for example could lead into a episode listing instead to the titles if MyVideoNav.xml was left without wallking the whole way back before. The same applies to all other nodes that are stacked.
By using a node you can directly and correctly point to a path without any hassle. 
I would use the default ones, but library://video/tvshows/titles.xml are also including empty shows and it's ignoring the Kodi setting "show empty TV shows".

ronie: > Why shouldn't this be allowed?

as mentioned in the wiki https://kodi.wiki/view/Add-on_rules#Requirements_for_scripts_and_plugins

`add-ons should store all their data in their own subfolder inside the addon_data directory. Access (read/write/delete) to any other files / folders is not allowed by default.`

if these custom nodes are used to overcome certain issues in kodi, as i gather from your reply, i'd much rather see some effort the address these issues in core instead of introducing skin/script workarounds.

sualfred: @ronie 
Sorry, but that would mean that the library node editor also has to be removed. It does exactly the same and stores nothing in the addon_data folder, because library nodes only work in library:// and not outside of the path. If that would be the case skins could store them directly in the skin directory.
This function just offers the feature that skins can ship premade nodes without pointing users to https://kodi.wiki/view/Video_nodes or the https://kodi.wiki/view/Add-on:Library_Node_Editor which are too complicated for most of the common users.  It simplify things. And creates advantages to script.skinshortcuts, because these nodes can be listed there automatically.

My example only showed one scenario that is helpful as workaround but it can do a lot more. And sorry, I know nearly nothing about c++. I'm a skinner who just learned Python. Don't expect that I'm able to learn Chinese too. I don't have any single time left for it. As much as I would do it, but my pregnant wife and my daughter would kill me.

enen92: We sometimes make exceptions. For instance, add-ons cannot write to the keymap.xml file but the keymap editor add-on does it and is the only add-on allowed to do it. It is not a question of removing add-ons that do the same...but more of a compromise to reduce the error footprint while not compromising functionality. If fixing core is not an option maybe call into the node editor add-on?

sualfred: Doesn't help and won't work. 

Don't get me wrong. I agree that addons should store everything ins it's own addon_data folder when it's possible. But things like your mentioned keymap, or a profile based custom node, which can the user create on his own if he understands what he is doing(!!!!), that should be allowed to be handled, or in this case extended, by an addon. I do not harm any default library node. I also force a skin folder name, so it's clear and structured.

C'mon guys... don't let me get frustrated again because of a rule that cannot fit all scenarios.  We are talking just about node.xmls's not about a hacky core thing. It makes it more userfriendly and skinners don't have to deal with annoyning questions of unexperienced users and low WAF.

ronie: in case of the library node editor / keymap editor:
- it is manually installed by the user
- the user knowingly and willingly uses it to modify those files.
that's a use-case we've made an exception for.

as for your script, it is auto-installed and modifications are made in the background without the user realizing changes are made to his system.
in addition, those copied library nodes will be left behind when a user decides to uninstall a skin (or your helper addon).

btw, it wasn't my intention to say that _you_ should be the one to take on the job of fixing kodi core. but i'd rather have addon devs create a bug report on github than seeing them implement such workarounds.

sualfred: It's not a bug. It's a common and mostly correct behaviour. If I back out to the home screen and then return to the video DB it's useful to have the window history there. There is just no way to override this. And there are scenarios where those things just don't fit. 
I also wouldn't call it a workaround. It's a feature in my eyes that can make skinners life easier for other scenarios like shipping own playlists, what already works.

Kodi is not really self maintaining. The textures stuff get huge after months of runnings, addon_data is full of stuff from old addons or deleted addons.. I don't think 1kb big files will harm anybody next to all the other dead leftovers. If it calms you, I can put in an cleanup option to delete all script created nodes (but I bet noone will ever use it :wink: )

And with your argument that users are knowingly willing to modify files -> I don't modify anything. I copy possible new filter ones. In Embuary exactly 2 nodes.  And the script does have options to control everything: Kodi settings, addon settings, but  I let users know and decide. I show very big YES/NO dialog what I want to setup everything correctly so everything would work right out of the box (see https://github.com/sualfred/script.embuary.helper/wiki/Script:-Change-global-Kodi-settings#attention-and-warning ). 
Simplified and just-working-out-the-box > complicated and manual setup with a higher frustating level and jump-off-rate. It's all about the user experience.

Long story short:
Without this Python function my TV show hub integration will "break" and people will get empty TV shows again. 

Anyway.. you have to power to approve it. Not me. But I hope you can understand that I cannot agree and understand this rule for this case.

sualfred:
@ronie @enen92 

I've remove the skinnodes function and replaced it with a tvshowsroot() for my own needs.

sualfred:
Found a way to force the reset of the path history via skin level. No need for any workaround via Python or skin provided nodes.

`ActivateWindow(videos,videodb://tvshows/titles/?force_root=$INFO[System.Time(mm)]$INFO[System.Time(ss)],return)`

enen92:
Awesome, thanks a lot for adjusting

PR #1140, opened 2019-08-31
[script.module.inputstreamhelper] 0.4.0

dagwieers:
### Description
This release includes various reliability improvements and bugfixes,
a new API for add-ons and a settings menu for troubleshooting.

This release is the first release with integration tests and test
coverage, and was a collaborative effort of a handful of people.

### Checklist:


dagwieers:
We would like to see this release as soon as possible, since it is blocking the release of other add-ons.
And the new functionality to reinstall/remove Widevine is very welcome for troubleshooting playback issues.

script.module.inputstreamhelper/addon.xml
```diff
@@ -1,25 +1,45 @@
...
+        <import addon="script.module.inputstreamhelper" optional="true"/>
         <import addon="script.module.kodi-six" version="0.0.4"/>
-        <import addon="script.module.requests" version="2.9.1"/>
+        <import addon="script.module.pysocks" version="1.6.8" optional="true"/>
+        <import addon="script.module.routing" version="0.2.0"/>
     </requires>
-    <extension library="lib" point="xbmc.python.module"/>
+    <extension point="xbmc.python.module" library="lib"/>
+    <!-- This is needed for running from RunPlugin() -->
+    <extension point="xbmc.python.pluginsource" library="addon.py"/>
```
enen92: ?

script.module.inputstreamhelper/changelog.txt
```diff
@@ -1,56 +1,83 @@
...
-+ Automatic Widevine CDM download on ARM devices
-+ Display Widevine EULA during installation procedure
-+ German translation (thanks to asciidisco)
-+ New, smaller and less ugly generic icon
-+ Better exception handling
-+ Code cleanup
-
-2017.09.13 v0.1.0
-+ Initial release
+## v0.4.0 (2019-09-01)
```
enen92: Changelog is no longer parsed by kodi, I'd be nice to not include it in the submissions

dagwieers: Ok, I can remove it.

IMO it is useful to have it included on the system.

script.module.inputstreamhelper/addon.xml
dagwieers: I tried various things to make this work. Have an API for add-ons as well as an icon that can bring users to the settings of the plugin. This is what eventually happened to work.

enen92: Why not turn the navigation into a dialog select? Using this add-on as a plugin doesn't really result into a seamless user experience

dagwieers: I don't know what you mean with "turn the navigation into a dialog select".
What we have now is a vast improvement of what was before and it works as designed.

enen92: I just realized you don't even have any navigation inside the plugin entrypoint. Plugins are VFS add-ons that display directory items to users and provide content to Kodi. In this case your plugin add-on entrypoint is a script and you are only using the plugin structure to allow external calls to some of the methods. I am afraid you'll have to do things properly and not abuse the plugin interface. Turn add-on.py into a script and use RunScript instead.

dagwieers: No, we don't want to have a myriad of scripts as interface to other add-ons.
https://github.com/emilsvennesson/script.module.inputstreamhelper/wiki/API

enen92: Nop sorry this is not a plugin and you can't use it like one. If you want an external API to other scripts the add-on already has a script module entrypoint, so other add-ons import it as a dependency and call into the appropriate methods. If you need to expose that as built-in functions make addon.py a script and not a plugin.

Read the difference between plugins and scripts here: https://kodi.wiki/view/About_Add-ons

dagwieers:
@emilsvennesson Back to you.

script.module.inputstreamhelper/addon.xml
enen92: Furthermore you don't need a myriad of scripts as interface to other asdons...add-on py is enough.
You parse the args just like a plugin and redirect to the appropriate method. You call it from other add-ons as `RunScript(script, args)`.

PR #1146, opened 2019-09-03
[script.module.inputstreamhelper] 0.4.2

mediaminister:
### Description

This release includes urgent fixes for two problems that unfortunately emerged after the 0.4.1 release:
- Fix unresponsive Kodi when opening add-on information pane
- Revert a commit that broke Chrome OS extracting for some platforms

This release also contains a new InputStream Helper information settings page and some small improvements regarding add-on structure and performance.

### Checklist:

Additional information :

### DIFF
--- a/script.module.inputstreamhelper/default.py
+++ b/script.module.inputstreamhelper/default.py
@@ -1,0 +1,10 @@
+# -*- coding: utf-8 -*-
+''' This is the actual InputStream Helper API script entry point '''
+
+from __future__ import absolute_import, division, unicode_literals
+import os
+import sys
+sys.path.append(os.path.join(os.path.dirname(os.path.realpath(__file__)), 'lib'))  # noqa: E402
+from inputstreamhelper.api import run
+
+run(sys.argv)
--- a/script.module.inputstreamhelper/README.md
+++ b/script.module.inputstreamhelper/README.md
@@ -61,3 +61,106 @@ Please report any issues or bug reports on the [GitHub Issues](https://github.co
 
 ## License ##
 This module is licensed under the **The MIT License**. Please see the [LICENSE.txt](LICENSE.txt) file for details.
+
+## Releases
+## v0.4.2 (2019-09-03)
+- Move release history to readme file (@mediaminister)
+- Clean up coverage/codecov config (@dagwieers)
+- Add InputStream Helper Information to settings page (@horstle, @dagwieers)
+- Make sure addon.xml meets all requirements (@mediaminister)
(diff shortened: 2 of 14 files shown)


dagwieers:
I approve of this new release.

script.module.inputstreamhelper/default.py
```diff
@@ -0,0 +1,10 @@
+# -*- coding: utf-8 -*-
+''' This is the actual InputStream Helper API script entry point '''
+
+from __future__ import absolute_import, division, unicode_literals
+import os
+import sys
```
enen92: Why is this needed?

enen92: I mean the sys.path.append

Lunatixz:
LGTM

script.module.inputstreamhelper/default.py
mediaminister: Because default.py (**xbmc.python.script** extension point) needs access to  to functions of the **xbmc.python.module** extension point` but is not in the same directory.
In 0.4.1 we imported script.module.inputstreamhelper into itself to solve this problem, but that caused the Kodi add-on information pane to freeze.

enen92: Hum, you guys are likely doing something wrong then. There's no need to append to sys.path inside the same add-on. I'd recommend to take a deeper look in the future to see which imports caused the circular dependency in the first place. For now I'm fine with it.

dagwieers: Or we have been bitten by an issue I have experienced myself various times before, which causes a just installed script.module to not be added to PYTHONPATH. We have seen this for different modules and add-ons, so I doubt it is an issue with one specific module or add-on.

Possibly related to Kodi using the repository metadata and not the package metadata (as reported by @mediaminister).

basrieter: @dagwieers I have also noticed this. Perhaps report it at https://github.com/xbmc/xbmc/issues/

dagwieers: @basrieter Reported issues get closed without debug log, so I don't bother if it's not easily reproduced. Feel free to report it if you can.

basrieter: Well, you cannot expect devs to look into stuff without the proper information. Just provide them the info and people will look into it.

dagwieers: It depends on how you get about with reported issues. But I am not expecting anything from volunteers, as a volunteer. But in this case I didn't bother. And you didn't either :wink:

PR #1190, opened 2019-10-21
[script.simkl] 2.2.0

ekleop:
### Checklist:

Additional information :


ekleop:
Hi @Rechi  is there an approximate timeline when this will be merged?

script.simkl/addon.xml
```diff
@@ -24,5 +24,9 @@
       <fanart>resources/fanart.jpg</fanart>
     </assets>
     <language>en</language>
+    <news>
```
enen92: the news tag is only available for krypton and later versions

script.simkl/crowdin.yml
```diff
@@ -0,0 +1,4 @@
+files:
```
enen92: please keep this file out of the submission

script.simkl/addon.xml
enen92: Although I am ok with keeping it in the addon.xml since the tag will be ignored by jarvis

enen92:
@ekleop please also drop all the .po files without translations. There's no point in having those since Kodi will always fallback to the en as the default lang

script.simkl/addon.xml
ekleop: Doesn't krypton and later versions cannot install the addon? so they will use this tag or am I wrong?

Rechi: This add-on can be installed in Krypton and newer versions. Those versions do not have a `changelog.txt` fallback, so if you want to see a changelog inside Kodi, you have to keep the news tag.

ekleop:
ok, I've removed yaml file + all empty language files.Everything should be fine now.

script.simkl/addon.xml
ekleop: yeap, that is what I wanted to know. We left it specifically for this. Thanks.

KarellenX:
@enen92 
Could you submit this if all is ok. It has been over 5 weeks since the PR was submitted.

Thanks

KarellenX:
Thanks @enen92

PR #1192, opened 2019-10-22
[script.module.trakt] 3.1.0

Razzeee:
Small update adding two features

### DIFF
--- a/script.module.trakt/addon.xml
+++ b/script.module.trakt/addon.xml
@@ -1,19 +1,19 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
 <addon id="script.module.trakt"
        name="trakt.py"
-       version="3.0.0"
-       provider-name="Fuzeman, Razzeee">
+       version="3.1.0"
+       provider-name="Dean Gardiner">
   <requires>
     <import addon="xbmc.python" version="2.14.0"/>
     <import addon="script.module.requests" version="2.12.4"/>
-    <import addon="script.module.six" version="1.9.0"/>
-    <import addon="script.module.arrow" version="0.6.0"/>
+    <import addon="script.module.six" version="1.10.0"/>
+    <import addon="script.module.arrow" version="0.10.0"/>
   </requires>
   <extension point="xbmc.python.module"
              library="lib" />
   <extension point="xbmc.addon.metadata">
-    <summary lang="en_GB">Python interface for the trakt.tv API</summary>
-    <description lang="en_GB">Packed for Kodi from https://github.com/fuzeman/trakt.py</description>
+    <summary lang="en">Python interface for the trakt.tv API</summary>
+    <description lang="en">Packed for Kodi from https://github.com/fuzeman/trakt.py</description>
     <platform>all</platform>
     <language></language>
     <license>MIT</license>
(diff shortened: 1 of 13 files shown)


script.module.trakt/addon.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
 <addon id="script.module.trakt"
        name="trakt.py"
-       version="3.0.0"
+       version="3.1.0"
        provider-name="Fuzeman, Razzeee">
```
Rechi: correct author is `Dean Gardiner`, https://github.com/fuzeman/trakt.py/blob/3.1.0/setup.cfg

Razzeee:
Updated

script.module.trakt/addon.xml
```diff
@@ -1,19 +1,19 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
 <addon id="script.module.trakt"
        name="trakt.py"
-       version="3.0.0"
-       provider-name="Fuzeman, Razzeee">
+       version="3.1.0"
+       provider-name="Dean Gardiner, Razzeee">
```
Rechi: author is only `Dean Gardiner`

script.module.trakt/addon.xml
```diff
@@ -1,19 +1,19 @@
...
   <requires>
     <import addon="xbmc.python" version="2.14.0"/>
-    <import addon="script.module.requests" version="2.12.4"/>
-    <import addon="script.module.six" version="1.9.0"/>
-    <import addon="script.module.arrow" version="0.6.0"/>
+    <import addon="script.module.requests" version="2.15.1"/>
```
Rechi: Upstream specifies `2.4.0` as minimum version. Why are you now bumping it to an even higher version then before?

script.module.trakt/addon.xml
```diff
@@ -1,19 +1,19 @@
...
+    <import addon="script.module.six" version="1.10.0"/>
+    <import addon="script.module.arrow" version="0.10.0"/>
   </requires>
   <extension point="xbmc.python.module"
              library="lib" />
   <extension point="xbmc.addon.metadata">
-    <summary lang="en_GB">Python interface for the trakt.tv API</summary>
-    <description lang="en_GB">Packed for Kodi from https://github.com/fuzeman/trakt.py</description>
+    <summary>Python interface for the trakt.tv API</summary>
+    <description>Packed for Kodi from https://github.com/fuzeman/trakt.py</description>
```
Rechi: `lang="en_GB"` isn't valid, correct is`lang="en"` for gotham

script.module.trakt/addon.xml
Razzeee: To get the latest version that we have in gotham

Rechi: So you don't know that version means minimum version and that the repo for a specific Kodi version contains only one version of every add-on.

Razzeee: Sure, but why not force it to get the latest with all possible bugfixes?

script.module.trakt/addon.xml
Razzeee: Changed

script.module.trakt/addon.xml
Razzeee: Changed

PR #1196, opened 2019-10-30
[screensaver.turnoff] v0.10.1

dagwieers:
### Description
- Add sanity tests, unit tests and coverage support
- Use JSON-RPC for all built-ins
- Improvements for Python 3
- Support Odroid-C2 display method
- Support RPi touchscreen display method
- Improve mute and unmuting audio using JSON-RPC
- Fix translations
- Fix an issue when stopping the screensaver

### Checklist:


screensaver.turnoff/addon.xml
```diff
@@ -1,13 +1,44 @@
...
+
+Next to managing your display, it can also manage your device, log your profile off or mute audio to avoid sounds through your A/V receiver.
+    </description>
+    <license>GNU General Public License, v2+</license>
     <email>dag@wieers.com</email>
+    <website>https://kodi.wiki/view/Add-on:Turn_Off</website>
+    <source>https://github.com/dagwieers/screensaver.turnoff</source>
+    <forum>https://forum.kodi.tv/showthread.php?tid=331076</forum>
+    <news>
+v0.10.1 (2019-10-30)
```
enen92: Please keep this to a minimum and only for the last version.

dagwieers: The other versions were never released. So this contains the condensed changelog since the previous release.

screensaver.turnoff/screensaver.py
```diff
@@ -1,164 +1,300 @@
...
+
+DEBUG_LOGGING = get_global_setting('debug.showloginfo')
+
+if __name__ == '__main__':
     # Do not start screensaver when command fails
-    screensaver = Screensaver('screensaver-turnoff.xml', addon_path, 'default')
-    screensaver.doModal()
-    del screensaver
+    TurnOffDialog('gui.xml', ADDON_PATH, 'default').doModal()
+    sys.modules.clear()
```
enen92: Leftover?

screensaver.turnoff/screensaver.py
```diff
@@ -1,164 +1,300 @@
+# -*- coding: utf-8 -*-
```
enen92: Normally recommended to keep the entrypoint as small as possible. Keep it in mind for a future update.

screensaver.turnoff/screensaver.py
dagwieers: A lot of screensavers are doing this, so I decided to include it as well.
I have had various issues with the implementation and this was one of the measures that made things a lot more reliable.

screensaver.turnoff/screensaver.py
dagwieers: Ok.

screensaver.turnoff/addon.xml
dagwieers: I removed the v0.9.1 part as that version was already released before.

Although I am in favor to keep a condensed changelog going back a few versions, so users can relate changes (in behaviour) to a (recent) update.

dagwieers:
@enen92 Is there anything blocking in this PR?

screensaver.turnoff/screensaver.py
dagwieers: This will be fixed in the next release.
https://github.com/dagwieers/screensaver.turnoff/pull/13

screensaver.turnoff/screensaver.py
dagwieers: After some more isolated testing, I found two issues and removed the `sys.modules.clear()` call.
dagwieers/screensaver.turnoff#13
dagwieers/screensaver.turnoff#14

I hope they weren't needed for Krypton :-/

dagwieers:
@enen92 @basrieter So what is needed to see this merged?

screensaver.turnoff/screensaver.py
```diff
@@ -1,164 +1,300 @@
...
+def log(level=1, msg='', **kwargs):
+    ''' Log info messages to Kodi '''
+    max_log_level = int(get_setting('max_log_level', 0))
+    if not DEBUG_LOGGING and not (level <= max_log_level and max_log_level != 0):
+        return
+    from string import Formatter
+    if kwargs:
+        msg = Formatter().vformat(msg, (), SafeDict(**kwargs))
+    msg = '[{addon}] {msg}'.format(addon=ADDON_ID, msg=msg)
+    xlog(from_unicode(msg), level % 3 if DEBUG_LOGGING else 2)
```
basrieter: Your logging is still weird. Add-ons should only log use the xbmc.LOGDEBUG (value 0) level. Seems that logging happens at all levels except 3 (because of the %3).

dagwieers: Only when you enable debug logging.

basrieter: If I don't have `DEBUG_LOGGING` on False, you always log with 2 = LOGNOTICE not LOGDEBUG (0)

dagwieers: So if you have `DEBUG_LOGGING` set to False, it depends on `max_log_level` whether anything gets logged. Specifically for **screensaver.turnoff** by default we log information regarding which display-method and/or system-method was attempted for suspend/resume or shutdown.

This is not different than how it used to work before. We need to log what action was performed in both cases even when `DEBUG_LOGGING` is disabled, because we offer about 10 different possible display methods and a handful of shutdown methods. We cannot know which method will work for the end-user as it not only depends on the system or platform, but also depends on the display device (projector, TV, display, smartphone, touchscreen, ...)

The user can disable logging, or enable more verbosity in the screensaver settings to help him troubleshoot his scenario. And the reason that this is required without `DEBUG_LOGGING` is because we need to know exactly what method was used when it failed, which could be at any time, even when Kodi is not being used. We cannot expect our users to have Kodi debug logging enabled all the time, or enable it before the screensaver kicks in.

dagwieers:
In the meantime we have released v0.10.2...

PR #1213, opened 2019-11-21
[screensaver.turnoff] v0.10.2

dagwieers:
### Description
- Create separate entrypoint
- Improve stability
- Add more unit tests
- Add sanity tests, unit tests and coverage support
- Use JSON-RPC for all built-ins
- Improvements for Python 3
- Support Odroid-C2 display method
- Support RPi touchscreen display method
- Improve mute and unmuting audio using JSON-RPC
- Fix translations
- Fix an issue when stopping the screensaver

### Checklist:


screensaver.turnoff/addon.xml
```diff
@@ -1,13 +1,49 @@
...
-    <description lang="en_GB">The Turn Off screensaver turns your TV, projector or monitor off, like any old fashioned screensaver is intended to do.</description>
-    <source>http://github.com/dagwieers/plugin.screensaver.off</source>
-    <license>GPL2+</license>
+    <summary lang="en_GB">Screensaver that turns your display off to save power</summary>
+    <description lang="en_GB">
+The Turn Off screensaver turns your TV, projector or screen off, like any old fashioned screensaver is intended to do.
+
+Next to managing your display, it can also manage your device power state, log your profile off or mute audio to avoid sounds through your A/V receiver.
+    </description>
+    <license>GNU General Public License, v2+</license>
```
enen92: Please use spdx

screensaver.turnoff/screensaver.py
```diff
@@ -1,164 +1,304 @@
...
+
+    def onScreensaverDeactivated(self):  # pylint: disable=invalid-name
+        ''' Perform cleanup function '''
+        self.action()
+
+
+def run():
+    ''' Runs the screensaver '''
+
+    # If player has media, avoid running
```
enen92: I assume this is because of audio playing?

dagwieers: Or when testing the screensaver when media is playing.

This was giving me issues, when I discovered this is not supposed to work in Kodi.
(The screensaver test button should be disabled when media is playing IMO)

screensaver.turnoff/addon.xml
dagwieers: I will fix this in the next release. Thanks.

*It doesn't seem to be a requirement, as it is missing from the docs and just recently other add-ons were merged that did not use SPDX for their license-tag.
https://github.com/xbmc/addon-check/issues/126*

enen92: I documented on the wiki some months ago: https://kodi.wiki/view/Addon.xml#.3Clicense.3E
 It's not mandatory but it is nice to start having some sort of standardization. Might be added to the addon checker if someone steps up.
Thanks for adjusting

dagwieers: Oh, I agree with having SPDX as a standard.

PR #1222, opened 2019-12-04
[script.module.xbmcswift2] 2.5.0

enenbot:
### Add-on details:

- **General**
  - Add-on name: xbmcswift2
  - Add-on ID: script.module.xbmcswift2
  - Version number: 2.5.0
  - Kodi/repository version: gotham

- **Code location**
  - URL: 
  
xbmcswift2 is a small framework to ease development of KODI addons.

### Checklist:

### DIFF
--- a/script.module.xbmcswift2/addon.xml
+++ b/script.module.xbmcswift2/addon.xml
@@ -1,10 +1,13 @@
-<addon id="script.module.xbmcswift2" name="xbmcswift2" provider-name="Jonathan Beluch (jbel)" version="2.4.0">
+<addon id="script.module.xbmcswift2" name="xbmcswift2" provider-name="Jonathan Beluch (jbel), enen92" version="2.5.0">
   <requires>
-    <import addon="xbmc.python" version="2.1.0" />
+    <import addon="xbmc.python" version="2.14.0" />
   </requires>
   <extension library="lib" point="xbmc.python.module" />
   <extension point="xbmc.addon.metadata">
-    <language />
     <platform>all</platform>
+    <source>https://github.com/XBMC-Addons/script.module.xbmcswift2</source>
+    <license>GPL-3.0</license>
+    <summary lang="en">xbmcswift2 is a small framework to ease development of KODI addons.</summary>
+    <description lang="en">xbmcswift2 is a small framework to ease development of KODI addons.</description>
   </extension>
 </addon>
--- a/script.module.xbmcswift2/lib/xbmcswift2/listitem.py
+++ b/script.module.xbmcswift2/lib/xbmcswift2/listitem.py
@@ -9,6 +9,7 @@
     :license: GPLv3, see LICENSE for more details.
 '''
 from xbmcswift2 import xbmcgui
+from xbmcswift2.logger import log
 
 
 class ListItem(object):
@@ -16,15 +17,13 @@ class ListItem(object):
     of any set properties that xbmcgui doesn't expose getters for.
     '''
     def __init__(self, label=None, label2=None, icon=None, thumbnail=None,
-                 path=None):
+                 path=None, fanart=None):
         '''Defaults are an emtpy string since xbmcgui.ListItem will not
         accept None.
         '''
         kwargs = {
             'label': label,
             'label2': label2,
-            'iconImage': icon,
-            'thumbnailImage': thumbnail,
             'path': path,
         }
         #kwargs = dict((key, val) for key, val in locals().items() if val is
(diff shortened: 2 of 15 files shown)


enen92:
@Rechi can I have your review here if you have some time? This script.module is the base of a lot of addons that are now broken. All this PR does is to ensure initial compatibility with python3 and the removal of some of the unauthorised/deprecated API methods.
Was also moved to https://github.com/XBMC-Addons since the author is inactive. I'll likely update the module as I port some of the unmaintained addons that depend on it.

script.module.xbmcswift2/addon.xml
```diff
@@ -1,10 +1,13 @@
-<addon id="script.module.xbmcswift2" name="xbmcswift2" provider-name="Jonathan Beluch (jbel)" version="2.4.0">
+<addon id="script.module.xbmcswift2" name="xbmcswift2" provider-name="Jonathan Beluch (jbel), Team Kodi" version="2.5.0">
```
Rechi: If you add Team Kodi as provider, Team Kodi should have access to the upstream repo.

script.module.xbmcswift2/lib/xbmcswift2/listitem.py
```diff
@@ -15,16 +15,14 @@ class ListItem(object):
     '''A wrapper for the xbmcgui.ListItem class. The class keeps track
     of any set properties that xbmcgui doesn't expose getters for.
     '''
-    def __init__(self, label=None, label2=None, icon=None, thumbnail=None,
+    def __init__(self, label=None, label2=None, icon=None, thumbnail=None, fanart=None,
```
Rechi: This is a breaking change.

script.module.xbmcswift2/lib/xbmcswift2/listitem.py
```diff
@@ -183,16 +190,16 @@ def as_xbmc_listitem(self):
         return self._listitem
 
     @classmethod
-    def from_dict(cls, label=None, label2=None, icon=None, thumbnail=None,
+    def from_dict(cls, label=None, label2=None, icon=None, thumbnail=None, fanart=None,
```
Rechi: This is a breaking change.

script.module.xbmcswift2/lib/xbmcswift2/listitem.py
```diff
@@ -183,16 +190,16 @@ def as_xbmc_listitem(self):
...
                   path=None, selected=None, info=None, properties=None,
-                  context_menu=None, replace_context_menu=False,
```
Rechi: This is a breaking change.

script.module.xbmcswift2/lib/xbmcswift2/listitem.py
```diff
@@ -33,36 +31,41 @@ def __init__(self, label=None, label2=None, icon=None, thumbnail=None,
...
     def __str__(self):
+        if PY3:
+            return ('%s (%s)' % (self.label, self.path))
         return ('%s (%s)' % (self.label, self.path)).encode('utf-8')
 
     def get_context_menu_items(self):
         '''Returns the list of currently set context_menu items.'''
         return self._context_menu_items
 
-    def add_context_menu_items(self, items, replace_items=False):
```
Rechi: This is a breaking change and now function documentation is also wrong.

script.module.xbmcswift2/addon.xml
enen92: There are others such as library.node editor that also follow the same logic: https://github.com/XBMC-Addons/plugin.library.node.editor/blob/master/addon.xml#L2

I moved the addon to https://github.com/XBMC-Addons isn't that enough?

script.module.xbmcswift2/lib/xbmcswift2/listitem.py
enen92: Hum, do you recommend to change the addon-id to xbmcswift3 or is it okay to keep it as is? The addons I've been adapting always specify the version in addon.xml for the module

script.module.xbmcswift2/lib/xbmcswift2/listitem.py
enen92: Will adjust the docs

script.module.xbmcswift2/addon.xml
Rechi: I know that there are others, but Team Kodis GitHub organisation is https://github.com/xbmc.

script.module.xbmcswift2/lib/xbmcswift2/listitem.py
Rechi: I don't care if you break add-ons, just pointing out that this could cause issues. If you move the new argument at the end, it is backwards compatible.

Version is minimum version, not exact or maximum version. Or are you talking about something else?

script.module.xbmcswift2/addon.xml
enen92: changed to enen92.

enen92:
@Rechi I appreciate your input. Now it should be backwards compatible with the previous versions.

PR #1223, opened 2019-12-05
[script.module.unidecode] 1.1.1

enen92:
Update to the last version stored in pypi - python3 compatible and a dependency of library node editor


Rechi:
This version doesn't support Python 2.6.

enen92:
@Rechi I'll take a look

PR #1227, opened 2019-12-06
script.module.dropbox 9.4.0

robweber:
### Description
This will update the Dropbox python library to the most current version tagged for Python 3, [9.4.0](https://github.com/dropbox/dropbox-sdk-python/tree/v9.4.0). 

There are a few changes to this from the source that make it work within the Kodi addon system. These are the ```pkg_resources.py``` file and a modification to the ```stone_serializers.py``` which fixes errors related to repeated uses of the library (known datetime error). This was tested on the script.xbmcbackup addon on a development build of Kodi v19. 

### Checklist:

Additional information :


robweber:
Forgot to note that there is only one addon using this addon as a dependency currently in lower Kodi repos that I could find using the Kodi Addon Checker. It was __service.watchedlist__. Since I did remove the ```authorize.py``` file to make this a pure version of the script.module files this will break that addon if it tries to update for Kodi 19. 

I believe there is time once this is merged in for the addon author to: 

1. Update their code to include the authorize file themselves
2. Repackage the authorization flow as it's own addon for submission

enen92:
@robweber isn't the module also compatible with python2? If so, why not updating the version already available? Are there any other add-ons that depend on this module?

robweber:
If I'm honest - laziness. There is only one addon in the Kodi repo using this as a dependency. I didn't want to take the time to make sure all the other libraries (requests, etc) this module requires were backwards compatible with Leia plus cross check the Kodi version compatibility; when the incremental benefit in the lower branches would only help one additional addon. The lower revisions do work (8.x I believe). The 8.x revisions are the same as bundled with the current script.xbmcbackup module and they work for a pretty good population of users with no issues. 

I'll go through the exercise if you want to see if it will work on Leia.

enen92:
It's just because 160.000 lines are quite a big change to review, and we have to do it

robweber:
I did not realize that, I guess I thought you'd do a diff off the vendor provided lib or something. If I can push to the lower branch (assuming it's all compatible) then you can just see the changes? I can do that if it will help out.

robweber:
@enen92 
I was looking in to updating this for the Krypton branch (where the most currently version is now) and ran into a problem. 

Mostly it's due to the custom code. In the current version ([8.4.2](https://github.com/xbmc/repo-scripts/tree/krypton/script.module.dropbox)) there is the custom code I previously mentioned. This should probably be split into it's own addon. The problem I'm having is that if I push an update to the krypton branch with the pure Dropbox code there is going to be issues with addons dependent on this (only 1 I could find in Kodi repo) that will cause breaking changes. My thoughts pushing only to the Matrix branch were that it wouldn't break any current addons since they would need to update for Matrix anyway. Here are the options as I see them:

1. Push the 9.4.0 version to the Krypton branch, breaking addons that use it currently due to the custom code being missing. 
2. Continue with this PR for Matrix and communicate the changes, giving authors time to update their addons and possibly push the custom code in another addon since Matrix is still in development. 
3. Work with michag to split custom code into it's own addon first, then update 9.4.0 into Krypton - this will take more time as he has indicated he doesn't have time for development right now. 
4. Do nothing. I really only care about this module as it impacts the Backup addon. In order to expedite the merging of that PR I'll just pull the Dropbox pieces from it and return to them at a later date when this is all sorted out. 

I don't see a simple solution here to update this given that the custom code is already intertwined in the Python 2 repos. If I could provide a diff of the changes between the 8.4.2 and 9.4.0 versions would that help you review this PR and get it merged in (option 2)? [This is the closed PR](https://github.com/micahg/script.module.dropbox/pull/14) for micahg's branch that I had open before doing my own. It shows the changes between the 8.4.2 and 9.4.0 versions. I honestly think abandoning the lower versions and giving authors time to adapt to the missing custom pieces (either within their own addons or within a new addon with this as a dependency) is the easiest here.

enen92:
Keep it as is, I'll review this one. Need to do a diff first.
Will probably get back to it tonight

robweber:
Thanks - I know it's a pain. I can see now why there is a policy on mixing the vendor code with additional.

PR #1233, opened 2019-12-12
[script.module.arrow] 0.12.1

Razzeee:
Updates arrow to the lastest version


Rechi:
This version isn't compatible with Python 2.6.

Razzeee:
Which branch is the one to target then?

Rechi:
For a non Python 2.6 compatible version, Krypton.
There is a newer version of this module, which is Python 2.6 compatible.

Razzeee:
Not sure what your hinting at, https://pypi.org/project/arrow/ states it's only 2.7 and above. And that's the newest tagged version.

Rechi:
Have a look at https://github.com/crsmithdev/arrow/releases and find the last version which supports Python 2.6.

Razzeee:
Ah, that's your point :) 
0.12.1 would be what we want

Razzeee:
Updated PR to that version. My branch name might be a bit confusing.

PR #1234, opened 2019-12-12
[script.module.arrow] 0.15.4

Razzeee:
An older version of this is included in the gotham branch


Razzeee:
I don't think we want to change the code of the module provided by the autor, so the problem is fine.

Rechi:
No it's not, have a look what's reported as error and what's reported as info.

Razzeee:
https://github.com/xbmc/repo-scripts/pull/1235 needs to be merged first

PR #1235, opened 2019-12-12
[script.module.dateutil] 2.8.1

Razzeee:
Update dateutil to the latest version


Razzeee:
I don't think we want to change the code of the module provided by the autor, so the problem is fine.

Rechi:
No it's not, have a look what's reported as error and what's reported as info.

Razzeee:
Good catch, we just need to merge https://github.com/xbmc/repo-scripts/pull/1232 before this

PR #1238, opened 2019-12-13
[script.artistslideshow] 3.0.0

pkscout:
### Description

Note: because this version requires skin updates and has support for the Kodi 18 music artist folder, v3.0.0 will be Kodi 18+ only. The 2.x branch will continue to be supported for Kodi 17 and older.

v.3.0.0
- Python2/3 compatible
- rewrite to use a single image control instead of multi-image control in skin
- AS can now store and show music in Kodi 18 music artist folder
- AS now shows the Kodi artist fanart (if available) before anything else (with option to disable)
- added option to include Kodi album fanart in image list (defaults to False)
- added routine in settings to move images to Kodi artist information folder
- added option to disable extranfanart folder
- added option to control time between slide transitions
- AS can now store images and information files in separate directory structure
- AS now deletes empty artist information and artwork folders
- converted language files to current format for forward compatibility
- removed option to have a pop up displayed when downloading artwork

### Checklist:

Additional information :


script.artistslideshow/.gitignore
```diff
@@ -0,0 +1,130 @@
+# Byte-compiled / optimized / DLL files
```
enen92: please drop from the submission

script.artistslideshow/addon.xml
```diff
@@ -0,0 +1,95 @@
...
+		<description lang="ru_RU">Дополнение для загрузки изображений и дополнительной информации о воспроизводимом исполнителе с ресурсов fanart.tv and theaudiodb.com. Загруженные изображения и доп. информация, наравне с локальными, могут быть использованы для создания слайд-шоу во время прослушивания композиций.</description>
+		<description lang="sl_SI">Dodatek za prenos slik in dodatnih informacij  iz fanart.tv in theaudiodb.com za trenutno predvajanega izvajalca. Slike, skupaj z lokalnimi slikami izvajalcev ter informacije, se lahko uporabljajo s preobleko za ustvarjanje diaprojekcije slik izvajalca, ki ga trenutno poslušate.</description>
+		<description lang="sv_SE">Tillägg för att ladda ner bilder och tilläggsinformation från fanart.tv och theaudiodb.com om artisten som spelas för tillfället. Bilderna, tillsammans med lokala artistbilder och information, kan användas av skinnet för att skapa ett bildspel av artisten man lyssnar på.</description>
+		<description lang="zh_CN">从fanart.tv和theaudiodb.com下载当前播放艺人的图片和附加信息的插件。这此图片可以和本地艺人图片及信息一起，被皮肤用来创建正收听的艺人的幻灯片。</description>
+		<language />
```
enen92: please drop language tag

script.artistslideshow/addon.xml
```diff
@@ -0,0 +1,95 @@
...
+		<platform>all</platform>
+		<license>GNU GENERAL PUBLIC LICENSE. Version 2, June 1991</license>
```
enen92: please use a spdx identifier if possible: https://spdx.org/licenses/

script.artistslideshow/addon.xml
```diff
@@ -0,0 +1,95 @@
...
+		<forum>https://forum.kodi.tv/showthread.php?t=124880</forum>
+		<website>https://kodi.wiki/index.php?title=Add-on:Artist_Slideshow</website>
+		<email>artistslideshow.xbmc@gmail.com</email>
```
enen92: please confirm this is still valid. If not please drop or change to a valid email

script.artistslideshow/changelog.txt
```diff
@@ -0,0 +1,354 @@
+v.3.0.0
```
enen92: please drop. Leia and up do not use the changelog in the interface but only the news tag. You're still free to keep it along your code in the upstream repo

script.artistslideshow/resources/common/fileops.py
```diff
@@ -0,0 +1,208 @@
...
+        for t in _range( timeout * 4 ):
+            time.sleep( 0.25 )
+            if p.poll() is not None:
+                return p.communicate(), log_lines
+        p.kill()
+        log_lines.append( 'script took too long to run, terminating' )
+        return False, log_lines
+    else:
+        log_lines.append( 'running command with os.system' )
+        os.system( command )
```
enen92: I'm not the biggest fan of this. Also the same for the mv and delete generics. Makes it harder to review since we need to take into account all the usages of the methods

script.artistslideshow/resources/common/fileops.py
```diff
@@ -0,0 +1,208 @@
...
+
+def naturalKeys( text ):
+    '''
+    alist.sort( key=naturalKeys ) sorts in human order
+    http://nedbatchelder.com/blog/200712/human_sorting.html
+    '''
+    return [ atoi( c ) for c in re.split( r'(\d+)', text ) ]
+
+
+def popenWithTimeout( command, timeout ):
```
enen92: looks like it isn't even used

script.artistslideshow/resources/common/fileops.py
```diff
@@ -0,0 +1,208 @@
...
+            log_lines.append( 'unknown error while reading data from ' + filename )
+            log_lines.append( e )
+            return log_lines, ''
+        return log_lines, data
+    else:
+        log_lines.append( '%s does not exist' % filename )
+        return log_lines, ''
+
+
+def renameFile ( src, dst ):
```
enen92: used?

script.artistslideshow/resources/plugins/fanarttv.py
```diff
@@ -0,0 +1,136 @@
...
+                self.loglines.extend( wloglines )
+        exists, cloglines = checkPath( filepath, False )
+        self.loglines.extend( cloglines )
+        if exists:
+            rloglines, rawdata = readFile( filepath )
+            self.loglines.extend( rloglines )
+            try:
+                json_data = _json.loads( rawdata )
+            except ValueError:
+                success, dloglines = deleteFile( filepath )
```
enen92: Not sure this is allowed

script.artistslideshow/resources/common/fileops.py
pkscout: I understand the review concern, but all the common libraries hold things that I use in many different projects (both Kodi addons and generic python scripts), and they allow me to include error checking that would otherwise have to be copied and pasted throughout the code (which would make it harder for me to manage).  I've been using these libraries for a number of years with Kodi submissions without issue, but if I have to get rid of them I will.  It'll just be awhile before I can resubmit since I'll have to rewrite a pretty substantial portion of AS.

script.artistslideshow/resources/common/fileops.py
pkscout: This is a shorter version of my comment above. The libraries in question are commone across all my projects, so they contain some things that aren't always used in a given project but that I use in more than one.  Assuming I'm allowed to keep them, I can delete any unused one, but then that creates new work for me to maintain different versions of the common libraries across my projects.  As before, I'll do that if I have to.

script.artistslideshow/resources/common/fileops.py
pkscout: See other comments above.

script.artistslideshow/resources/plugins/fanarttv.py
pkscout: I'm not sure to what you're referring.  This piece of code has existed in AS for quite some time and has previously been allowed in the repo.  But if you can be more specific I can change it.

script.artistslideshow/addon.xml
pkscout: Done.

script.artistslideshow/changelog.txt
pkscout: Will do.  Since I'm going to have to manually submit to delete the .gitignore file anyway, I can delete this one too.  I still greatly dislike the fact that the changelog interface was removed.  Having news in the addon is just double work for addon developers.  But I know that ship has sailed.

script.artistslideshow/.gitignore
pkscout: I'll have to manually submit this rather than using the tool I found for submitting from my master branch, as I can't find a way for GitHub to ignore my gitignore file (even if I add .gitignore to the ignore file).  This is apparently a "feature" of github know so that people sharing repos can share gitignore files.  But once the other comments on this are resolved I'll close this PR and resubmit without this file.

script.artistslideshow/resources/plugins/fanarttv.py
enen92: No problem for the script. It is just a concern that this creates an attack vector for other scripts to explore. They can just call the script with whatever file they want, or achieve arbitrary code execution.
And such things can be easily missed out in submissions. Not to talk about external scripts that are not in our repo and and miss our review process.

pkscout: I think I see a way I could add some validation to the filepath so that it's at least in the parent directory I'm expecting. I might be able to check the full path too, but I need to look at that more closely, as AS allows users to set an arbitrary directory where the folder if files is kept. I have that in the settings, I just need to see if I can get it from within this part of the script.  Would that help?

I think I'm about to show my level of Python experience (even after seven years of self taught Python) and understanding of Kodi addons, but how would another addon call an arbitrary function within my code?  If I understood the threat vector better I could take a look at all the AS code and try and address it in a future version.

enen92: In this script the problem does not reflect much as the addon calling into yours would need to add your file to the python path (e.g. sys.path.append). The problem is more evident in script modules that can just be set as a dependency of another addon. Either way, sometimes it is just a question of common-sense and avoid creating the possible hole in the first place. For instance, in this case, the `os.system` call belongs to a method that doesn't even get called. So why keep it in the first place?
I understand you reuse a lot of the same methods across addons and that's the reason the method is there. Running a linter against your code will detect all those unused methods.
IIRC some time ago team kodi was suggesting addon developers to use `codacity` to perform static analysis on the code automatically.

pkscout: Thanks for helping me better understand all this.  Given the larger security concerns with subprocess and os.system, I went ahead and removed that code.  I'll have to go back and look, but it might only be used in one addon anyway.

Thanks also for the Codacy recommendation.  I had setup codacy sometime ago but forgot about it.  I just ran AS through that and fixed a bunch of stuff (including SO MUCH TRAILING WHITESPACE).  Codacy identified some other security issues with ElementTree, so I moved over to defusedxml, and I had some unverified HTTPS connections with requests that I fixed too.  Some of the method versus function stuff I decided to ignore, as there are readability reasons to have some of the code as methods inside a class even though they could be functions.  I also have one except clause that is unbounded, but I can't find any docs on what kind of error messages xmbc.GetInfoLabel returns, so I had to leave it.   Anyway, Codacy now rates the AS code as an "A."  I'll close this request and submit a new pull request with all the updates (and exclude the two files you mentioned earlier).

pkscout:
As noted above, closing this pull request so I can do another one with updated code.

PR #1264, opened 2020-01-07
[script.artistslideshow] 3.1.0

pkscout:
### Description
v.3.1.0
- In daemon mode you can now change AS settings without having to restart Kodi
- added back option to show notifications when downloading new images
- updated service plugins to use https for API calls
- added options to change sleep time for main and slideshow threads
- set slideshow thread to close if no music playing
- added option to search stream titles more aggressively for an artist name
- fix for crash if artist image name has an extended character in it
- fix for crash when trying to get last fanart number from Kodi Artist Information Folder
- fix for artist images not changing when using a radio stream
- fix for theaudiodb.com plugin not downloading all three images for an artist
- fix for unicode errors causing cache trimming not to work
- fix for non-images getting added to image list and showing a black screen
- fix for high CPU usage when using AS in daemon mode and no music playing
- fix for issue where slideshow might not start when in daemon mode
- fix for runaway loop if there is only one image in an artist's slideshow
- fix for AS taking too long to quit on Kodi exit when in daemon mode
- fix for Setting Type Error in Kodi logs

### Checklist:

Additional information :


pkscout:
Just one quick note.  On line 240 I'm using xmbc.sleep even though I know you all would rather people use xbmc.Monitor().waitForAbort(). During testing it appeared to me that if I used waitForAbort in the slideshow thread that the main thread never got any waitForAbort responses, and that made it nearly impossible to track abort requests.  Using xbmc.sleep in the secondary thread resolved that.  While xbmc.sleep does take a variable, in the settings it's a slider that can only be a max of 3 seconds.  In testing AS never took more than 4 seconds to quit on an abort request, and usually it was much faster.

script.artistslideshow/resources/lib/artistslideshow.py
```diff
@@ -0,0 +1,1240 @@
...
+          self.WINDOW.setProperty( property_name, value )
+          lw.log( ['%s set to %s' % (property_name, value)] )
+        except Exception as e:
+          lw.log( ['Exception: Could not set property %s to value %s' % (property_name, value), e] )
+
+
+    def _wait( self, wait_time=1, sleep_time=1 ):
+        waited = 0
+        while waited < wait_time:
+            xbmc.sleep( sleep_time*1000 )
```
enen92: when dealing with seconds please always try to use `xbmc.Monitor().waitForAbort()`. This method  argument can be configured in the settings of your addon so it is highly likely the user doesn't know what he's/she's doing. Kodi will hang on exit if abort is requested while your addon is sleeping in the said conditions.

script.artistslideshow/resources/lib/artistslideshow.py
```diff
@@ -0,0 +1,1240 @@
...
+            else:
+                lw.log( ['first song started'] )
+                if not xbmc.Monitor().waitForAbort( 1 ): # it may take some time for Kodi to read the tag info after playback started
+                    self._slideshow_thread_start()
+                    self._use_correct_artwork()
+                    self._trim_cache()
+                else:
+                    self._set_property( 'ArtistSlideshowRunning' )
+            sleeping = False
+            while not xbmc.Monitor().abortRequested() and self._get_infolabel( self.ARTISTSLIDESHOWRUNNING ) == 'True':
```
enen92: probably a good ideia do create monitor and player objects and define them as member variables on init. Otherwise you're creating them all the time (which is expensive).

script.artistslideshow/resources/lib/artistslideshow.py
```diff
@@ -0,0 +1,1240 @@
...
+            if artist_name:
+                artists_info.append( (py2_encode( artist_name ), self._get_musicbrainz_id( py2_encode( artist_name ), mbid )) )
+        return artists_info
+
+
+    def _get_current_artists_info( self ):
+        featured_artists = ''
+        artist_names = []
+        mbids = []
+        if xbmc.Player().isPlayingAudio():
```
enen92: see, creating an instance again

script.artistslideshow/resources/lib/artistslideshow.py
```diff
@@ -0,0 +1,1240 @@
...
+
+    def _get_playing_item( self, item ):
+        got_item = False
+        playing_item = ''
+        max_trys = 3
+        num_trys = 1
+        while not got_item:
+            try:
+                if item == 'album':
+                    playing_item = xbmc.Player().getMusicInfoTag().getAlbum()
```
enen92: again and below again

script.artistslideshow/resources/lib/artistslideshow.py
enen92: Just seen your comment above. What if you check `xbmc.Monitor().abortRequested()` after the waitForAbort?

pkscout: I'll try that suggestion and see what happens.  I think I might have to do that in both threads to keep AS from hanging Kodi on quit.  I knew you wouldn't like the xbmc.sleep().  '-)

pkscout: It looks like changing the xbmc.sleep() to self.MONITOR.waitForAbort( time ) is OK now.  So maybe it was something else in the code that I cleaned up at some point that was causing the hang.  Interestingly, the slideshow thread will likely never see the abort request because it looks like when Kodi quits it stops the player before sending abort requests to scripts. When the player stops the main thread sends a quit command to the slideshow thread and stops it, so there's a really narrow window that will likely never happen in the real world where an abort request gets to the script before it realizes the player has stopped.

pkscout:
I'm going to close this PR and submit a new one with the updates.

PR #1286, opened 2020-02-05
[Matrix] set xbmc.python to 3.0.0

ronie:
the addon repo currently lists both python2 and python 3 addons in Kodi Matrix.
all python 2 addons will ofcourse not work in Matrix, and shouldn't be listed.

this change will make sure only addons with xbmc.python set to 3.0.0 will be listed in the addon manager.

this should be merged once https://github.com/xbmc/xbmc/pull/17188 gets merged.

### DIFF
--- a/targets.cfg
+++ b/targets.cfg
@@ -77,10 +77,10 @@ minversions =
     xbmc.addon:12.0.0
 
 [matrix]
-branches = gotham,helix,isengard,jarvis,krypton,leia,matrix
+branches = matrix
 minversions =
     xbmc.gui:5.14.0,
-    xbmc.python:2.1.0,
+    xbmc.python:3.0.0,
     xbmc.json:6.0.0,
     xbmc.metadata:1.0,
     xbmc.addon:12.0.0


targets.cfg
```diff
@@ -80,7 +80,7 @@ minversions =
 branches = gotham,helix,isengard,jarvis,krypton,leia,matrix
```
anxdpanic: Would it make sense to remove gotham to leia as well, since a minimum xbmc.python version 3.0.0 will only be satisfied by matrix+ ?

ronie: yeah, makes perfect sense.
thx!

enen92:
Did we reach any decision regarding the timeline for this? IMO if this is the way to go (I personally think it is) the change should not be delayed for a long time to give addon-devs a bit of time to start moving the addons. So that when Matrix is out, at least we have a decent list of addons there already.

ronie:
yup, see: https://github.com/xbmc/xbmc/issues/17326#issuecomment-586521056

i plan to merge this PR in ~10 days, to give devs a bit of time to get their addon in matrix before we make the switch.

targets.cfg
ronie: turned out it didn't make sense..
we do want to pull skins/scrapers/resources from lower branches.

i'll revert it.

odilon1962:
> the addon repo currently lists both python2 and python 3 addons in Kodi Matrix.
> all python 2 addons will ofcourse not work in Matrix, and shouldn't be listed.
> 
> this change will make sure only addons with xbmc.python set to 3.0.0 will be listed in the addon manager.
> 
> this should be merged once [xbmc/xbmc#17188](https://github.com/xbmc/xbmc/pull/17188) gets merged.

PR #1252, opened 2019-12-22
[script.tag-generator] 0.7.0

wellspokenman:
### Description
Leia support and bugfixes
### Checklist:

- [X] My code follows the [add-on rules](http://kodi.wiki/view/Add-on_rules) and [piracy stance](http://kodi.wiki/view/Official:Forum_rules#Piracy_Policy) of this project. 
- [X] I have read the [CONTRIBUTING](https://github.com/xbmc/repo-scripts/blob/master/CONTRIBUTING.md) document
- [X] Each add-on submission should be a single commit with using the following style: [script.foo.bar] 1.0.0

Additional information :
- Submitting your add-on to this specific branch makes it available to any Kodi version equal or higher than the branch name with the applicable Kodi dependencies limits.
- [add-on development](http://kodi.wiki/view/Add-on_development) wiki page.
- Kodi [pydocs](http://kodi.wiki/view/PyDocs) provide information about the Python API
- [PEP8](https://www.python.org/dev/peps/pep-0008/) codingstyle which is considered best practise but not mandatory.
- This add-on repository has automated code guideline check which could help you improve your coding. You can find the results of these check at [Codacy](https://www.codacy.com/app/Kodi/repo-scripts/dashboard). You can create your own account as well to continuously monitor your python coding before submitting to repo.
- Development questions can be asked in the [add-on development](http://forum.kodi.tv/forumdisplay.php?fid=26) section on the Kodi forum.

PR #1299, opened 2020-02-10
[script.tag-generator] Fix broken leia repo generator

enen92:
Leia repo generator has been broken for 2 months...
Tag generator includes a clearlogo and points to a jpeg file that doesn't exist. Correct file is has png extension. 
Wondering why addon-checker let this one pass. @Razzeee for reference (I'll look later)

Thanks to @kib for the server support.

Cross-reference: https://github.com/xbmc/repo-scripts/pull/1252

### DIFF
--- a/script.tag-generator/addon.xml
+++ b/script.tag-generator/addon.xml
@@ -28,7 +28,7 @@
     <assets>
         <icon>resources/icon.png</icon>
         <fanart>resources/fanart.jpg</fanart>
-        <clearlogo>resources/clearlogo.jpg</clearlogo>
+        <clearlogo>resources/clearlogo.png</clearlogo>
     </assets>
   </extension>
 </addon>


anxdpanic:
Should clearlogo be added to https://kodi.wiki/view/Addon.xml#.3Cassets.3E and addon-check? Or should it be reported as an error in addon-check if an unsupported asset is included?
It looks like it's just skipped over in addon-check atm

enen92:
@anxdpanic yeah I though the addon-checker was checking for this already but haven't looked into the code yet.

Razzeee:
There also seems to be `banner` in addition to `cleanart`? https://github.com/xbmc/xbmc/pull/12244/files

Are these used? Or is it just an option for third party skins to use these? Nothing of this seems to be documented?

enen92:
They are not used by estuary as far as I know, but they belong to the art map of the addon listitem, so any skin can take advantage of that. We should add them to the wiki and add the checks on addon-checker to avoid a similar issue in the future

Razzeee:
Well, they are not mandatory right? So we want a check for any tags, that are not supported at all I guess?

PR #1319, opened 2020-02-29
[service.xbmc.versioncheck] 0.5.4

anxdpanic:
### Description
add Kodi 18.6 stable release

### Checklist:

Additional information :

### DIFF
--- a/service.xbmc.versioncheck/resources/lib/version_check/distro/distro.py
+++ b/service.xbmc.versioncheck/resources/lib/version_check/distro/distro.py
@@ -1,0 +1,1216 @@
+# Copyright 2015,2016,2017 Nir Cohen
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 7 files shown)


service.xbmc.versioncheck/resources/lib/version_check/distro/distro.py
```diff
@@ -0,0 +1,1216 @@
+# Copyright 2015,2016,2017 Nir Cohen
```
basrieter: I assume that this file comes from https://github.com/nir0s/distro/blob/master/distro.py? 

Assuming that more people might need this functionality (espcially in Python 3.8+) it would make sense to please submit a proper `script.module.distro` add-on? @ronie  you suggested to just include it. What would be the best?

enen92: I think this add-on is included by default in Kodi, so moving it to a script.module will just increase complexity for no reason. Since it is a single file, it's say to keep it bundled

ronie: i'm not sure what the best practice is here..

for instance, if we want to ship kodi with the new python based scrapers, most of them have several dependencies as well which we would have to include in kodi.

anxdpanic: I'm happy to proceed with either method once it's decided.

PR #1321, opened 2020-02-29
[script.module.inputstreamhelper] 0.4.4

mediaminister:
### Description

 - Added option to restore a previously installed Widevine version
- Improve progress bar when extracting Widevine on ARM devices
- Improve Widevine library version detection
- Improve InputStream Helper information
- Increase download reliability for Chrome OS on ARM devices
- Added Japanese, Korean, Croatian and Hungarian translations
- Updated existing translations
- Various small bugfixes for Widevine installation on ARM devices

### Checklist:

Additional information :


script.module.inputstreamhelper/default.py
```diff
@@ -2,18 +2,18 @@
...
+from xbmcaddon import Addon
 
 
 def to_unicode(text, encoding='utf-8'):
     ''' Force text to unicode '''
     return text.decode(encoding) if isinstance(text, bytes) else text
 
 
-sys.path.append(os.path.join(to_unicode(xbmc.translatePath(xbmcaddon.Addon().getAddonInfo('path'))), 'lib'))
+sys.path.append(os.path.join(to_unicode(translatePath(Addon().getAddonInfo('path'))), 'lib'))
```
basrieter: Is it not possible to make it a proper package structure and not have a sys.path.append?

enen92: Same concern as bas. Why can't a proper module be used? I thought this was refactored already

BubbeleM:
I LOVE you

Am 01.03.2020 um 13:46 schrieb enen92 <notifications@github.com>:

﻿

@enen92 commented on this pull request.

Same concern as bas

________________________________

In script.module.inputstreamhelper/default.py<https://github.com/xbmc/repo-scripts/pull/1321#discussion_r386104971>:

>

 def to_unicode(text, encoding='utf-8'):
     ''' Force text to unicode '''
     return text.decode(encoding) if isinstance(text, bytes) else text


-sys.path.append(os.path.join(to_unicode(xbmc.translatePath(xbmcaddon.Addon().getAddonInfo('path'))), 'lib'))
+sys.path.append(os.path.join(to_unicode(translatePath(Addon().getAddonInfo('path'))), 'lib'))


Same concern as bas. Why can't a proper module be used? I thought this was refactored already

script.module.inputstreamhelper/default.py
mediaminister: Yes, I refactored the addon structure.

mediaminister: I refactored the addon structure.

dagwieers: And I think we should not have done that refactor. I think the reviewers are not understanding why we had to import the package ourselves like this.

And as a result we are now exposing *default.py* from PYTHONPATH potentially breaking add-ons that have their own default.py.

The reason why we had to change the python library path is because Kodi does not add our own package to the PYTHONPATH because the add-on does not require itself.

But since this was originally designed as a **script.module** add-on, but we still want to have the ability to have settings, we needed to do it this way. The refactoring may make *default.py* appear better, but in reality it makes things worse for the wrong reasons.

cc @basrieter @enen92

dagwieers: We need to split *default.py* (our settings) from the package and python library path.
And a **script.module** add-on should offer its package from *lib/* and not from *resources/lib/*.

dagwieers: @basrieter It is a proper package structure. But the *default.py* is not part of the package (or the library path, for good reasons) and Kodi does not load our package library path because the add-on cannot require itself.

So our own default.py (script) needs a work-around to find its own package. It may look ugly, and if you know of a prettier way to do this, fine, but this is the best we could come up with.

dagwieers: In the previous review this was already explained: https://github.com/xbmc/repo-scripts/pull/1163#pullrequestreview-308834407

dagwieers: Let me add a comment in the code to explain this, so we don't have to do that next time :wink:

basrieter: Could you not let the addon.xml point to:

```xml
<extension point="xbmc.python.script" library="lib/default.py">
</extension>
```

Never tried it myself, but then they are at the same level and then you won't have the conflict with imports in your `script` and `library`?

dagwieers: Same problem, default.py will be exposed in the python library path.

basrieter: But it does need to be called `default.py` it could be something else so it won't interfere?

dagwieers: The point is that this file doesn't need to be exposed in the python library path (which it will), and it could always conflict with something of the same name.

Kodi was simply not designed for this, and I don't think we need to change what we currently have. It may not look nice, but it works as designed and it is better than any alternative. And we discussed this before. So what exactly do you want to fix here?

dagwieers: Your original comment was:

> Is it not possible to make it a proper package structure and not have a sys.path.append?

It is a proper package structure, the main problem here is that Kodi doesn't include it in the library path for its own included scripts. That is why we have to do this ourselves in that line.

enen92: @dagwieers @mediaminister what's the problem with:

```
diff --git a/default.py b/default.py
index 66cf30e..28ac868 100644
--- a/default.py
+++ b/default.py
@@ -2,10 +2,9 @@
 ''' This is the actual InputStream Helper API script entry point '''

 from __future__ import absolute_import, division, unicode_literals
-import sys
-import os.path
 from xbmc import translatePath
 from xbmcaddon import Addon
+from lib.inputstreamhelper.api import run


 def to_unicode(text, encoding='utf-8'):
@@ -13,8 +12,4 @@ def to_unicode(text, encoding='utf-8'):
     return text.decode(encoding) if isinstance(text, bytes) else text


-# NOTE: This is required so our RunScript interface can import our own inputstreamhelper package
-sys.path.append(os.path.join(to_unicode(translatePath(Addon().getAddonInfo('path'))), 'lib'))
-from inputstreamhelper.api import run  # noqa: E402
-
 run(sys.argv)
diff --git a/lib/inputstreamhelper/__init__.py b/lib/inputstreamhelper/__init__.py
index b3c6ee4..fa14ea6 100644
--- a/lib/inputstreamhelper/__init__.py
+++ b/lib/inputstreamhelper/__init__.py
@@ -4,7 +4,7 @@

 from __future__ import absolute_import, division, unicode_literals
 import os
-from inputstreamhelper import config
+from . import config
 from .kodiutils import (addon_profile, addon_version, browsesingle, get_proxies, get_setting, get_setting_bool,
                         get_setting_float, get_setting_int, jsonrpc, kodi_to_ascii, kodi_version, localize, log, notification,
                         ok_dialog, progress_dialog, select_dialog, set_setting, set_setting_bool, textviewer,
diff --git a/lib/inputstreamhelper/api.py b/lib/inputstreamhelper/api.py
index 5fafa10..cabc18c 100644
--- a/lib/inputstreamhelper/api.py
+++ b/lib/inputstreamhelper/api.py
@@ -3,7 +3,7 @@
 """This is the actual InputStream Helper API script"""

 from __future__ import absolute_import, division, unicode_literals
-from inputstreamhelper import Helper
+from . import Helper
 from .kodiutils import ADDON, log
```

dagwieers: May be something to consider for v0.5.0, if it works (IIRC using relative imports were problematic for unit tests). Not something we should change to get past a review on a whim (while it passed reviews in the past).

enen92: We're trying to help you doing this right.

This is not a whim and has been pointed out in past reviews (since 0.4.2 at least - accepted temporarily as that update was labeled as an "urgent fix"). This release is not.

I spent 2 minutes implementing the change and testing both the script (opening the settings window) as well as a consumer (retrospect in this case). So please test wether this breaks anything or if it works as expected. If it does, there's no reason not to implement it rather than stubborness. Unit tests are not executed by kodi and can always be worked out (although I doubt they'll break).

I think I don't need to remind you but **team kodi** is the one in charge of the repository and its inclusion rules.

mediaminister: @enen92 Many thanks for figuring this solution out. I tested it and it just works fine. https://github.com/emilsvennesson/script.module.inputstreamhelper/pull/287

Once this gets merged into our master branch, I will update this pull request.

enen92: @mediaminister we're fine accepting this as is as long as if you guys really work out on removing the "hack" in the future and don't defer it to a future version that will never happen. The main problem here is that other devs literelly copy code from other addons as a base and assume this is the way to do it. Furthermore we create unwanted exceptions in the repo which are later hard to enforce to others (why can addon x do it and I can't?).

Kodi in this case doesn't do anything wrong. The script has access to the whole addon folder in the path, so `lib` is there. The script.module has access to the module folder (`lib/inputstreamhelper`) so it shouldn't import itself as if it was an external module. It's literally a 3 line fix.

mediaminister: I agree, I personally forgot about this "hack" and tried to fix it quickly but in a wrong manner.
So, I'm happy that we can move forward now.

horstle: @enen92 I think you're right with blocking it until this is merged. I also don't understand @dagwieers problem, since there are already relative imports (like .kodiutils).

mediaminister:
I think this can be merged now.

enen92:
thanks for adjusting

script.module.inputstreamhelper/default.py
dagwieers: > Kodi in this case doesn't do anything wrong.

Kodi *is* doing something wrong, it is not adding a perfectly fine package to the library path of its own scripts. The fact that we have to modify the package, that works fine for every add-on except for itself, is a workaround as well.

Using the package relatively is as much as a hack as modifying the library path IMO and marginally better. Functionally it is the same. But the reason we did it was because Kodi is *not* doing the right thing.

I am not in favor to make a modification like this just before a release where it did not get the same testing (especially since we had issues before with relative imports), but that is @mediaminister's call.

PS The irony here is that this futile thing is causing an add-on to be blocked during review, whereas Kodi devs happily accept something like xbmc/xbmc#17188 to be merged.

dagwieers: BTW I am not protesting the fix in itself, what I do find annoying is that this is blocking *this* release for a futility. When investigating this I found exactly one add-on in the exact same situation exposing its script in the package. Which was the original suggestion during review and was the wrong approach.

Mind you, there was no alternative workaround until @enen92 proposed this one.

PR #1362, opened 2020-03-29
[script.module.beautifulsoup4] 4.8.2+matrix.1

ronie:
we're currently quite a few versions behind, so let's update beautifulsoup4 to the latest version.

this needs a new dependency: [script.module.soupsieve](https://github.com/xbmc/repo-scripts/pull/1363).


ronie:
the source file is the py2 version.

i used the py3 wheel file from the downloads page

PR #1396, opened 2020-04-08
[script.module.livestreamer] 1.12.2+matrix.2

enen92:
The direct push to matrix of several script.modules made it skip the addon-checker checks. This is a round of pull requests to all the addons that are triggering errors in the matrix branch with a minor bump on the revision. Goal is to finally have a green branch (since matrix is pretty much starting from the beginning).

This is good to merge as long as travis does not complain.


L2501:
obsolete module, replaced by streamlink.

PR #1426, opened 2020-04-09
[script.tubecast] 1.4.2

enen92:
### Add-on details:

- **General**
  - Add-on name: TubeCast
  - Add-on ID: script.tubecast
  - Version number: 1.4.2
  - Kodi/repository version: krypton

- **Code location**
  - URL: 
  
An implementation of the Cast V1 protocol in Kodi to act as a player for the Youtube mobile application

### Description of changes:

            [ci] automated submissions to matrix
            [fix] adapted xbmcgui methods for matrix
            [fix] shrink screenshots
        

### Checklist:


script.tubecast/resources/lib/tubecast/youtube/app.py
```diff
@@ -430,7 +430,7 @@ def report_state_change(self, status_code, current_time, duration):  # type: (in
     def __post_bind(self, sc, postdata):  # type: (str, dict) -> None
         self.ofs += 1
         post_data = {"count": "1", "ofs": str(self.ofs), "req0__sc": sc}
-        for key in postdata.keys():
+        for key in list(postdata.keys()):
```
basrieter, line 433: I don't think it is needed to use `list` here, as `.keys()` returns an iterator, which is needed in the `for`. But it won't break either.

enen92: this was to silence 2to3 in addon checker, performance is likely lower and this is a false positive as I'm iterating (https://stackoverflow.com/questions/27476079/why-does-2to3-change-mydict-keys-to-listmydict-keys). I prefer to keep the list though, its not a big issue

PR #1432, opened 2020-04-10
[script.module.archive_tool] 2.0

zach-morris:
For Matrix
https://github.com/zach-morris/script.module.archive_tool

### Description

archive_tool is a set of common python functions to work with the Kodi archive virtual file system (vfs) binary addons such as vfs.libarchive and vfs.rar

### Checklist:

Additional information :


enen92:
Hey @zach-morris is there any addon using this module?

enen92:
If the code is exacly identical to the one in Leia (with just the addon.xml change) you probably want to version this differently and use PEP440 local version identifiers. 

in leia: `1.13.0`
in matrix: `1.13.0+matrix.1`

That way we know this is the same version

zach-morris:
OK, thanks.  I'll resolve these

enen92:
Policy has always been: script.modules to be in the repo must be a dependency of any addon that actually uses it. So, probably not a bad idea to submit your addon to the official repo. If your addon is the only one using this dependency and is shared in an external repo, maybe the place for the module is also in the external repo.

Ah it seems to be @highlandr. Maybe he can already start to use it as a dependency in his submissions?

zach-morris:
Yes probably @highlander would be the reason to submit it at this point then.  I can't submit IAGL since it contains shell scripts (which I think are against the rules) for external launching of programs.  I can put this module in my external repository for now.

highlandr:
Zach do whatever you want atm, you can delay, I put your name on the "news" :)
When you release this I will use it. I don't use vfs.libarchive just vfs.rar.
You can check the code in my git.

zach-morris:
OK, sounds like close this and put it in my repository then for now.  Thanks

PR #1454, opened 2020-04-13
[script.transmission] 1.0.0

basilgello:
### Description
The control panel add-on for Transmission Bittorrent client for Kodi 19.0 "Matrix" and newer versions

This repository is a rework of original Transmission-XBMC addon by Correl J. Roush for Kodi 19.0 "Matrix".

Since version 19.0 "Matrix", the Python 2 support will be dropped and only Python 3 addons will be supported.

Also, the original addon featured the 'Search Torrent' functionality using pirate sites. This violates the rules of official Kodl respository, so this addon has only 'Add Torrent File' functionality.
### Checklist:
- [x] My code follows the [add-on rules](http://kodi.wiki/view/Add-on_rules) and [piracy stance](http://kodi.wiki/view/Official:Forum_rules#Piracy_Policy) of this project. 
- [x] I have read the [CONTRIBUTING](https://github.com/xbmc/repo-scripts/blob/master/CONTRIBUTING.md) document
- [x] Each add-on submission should be a single commit with using the following style: [script.foo.bar] 1.0.0

Additional information :
- Submitting your add-on to this specific branch makes it available to any Kodi version equal or higher than the branch name with the applicable Kodi dependencies limits.
- [add-on development](http://kodi.wiki/view/Add-on_development) wiki page.
- Kodi [pydocs](http://kodi.wiki/view/PyDocs) provide information about the Python API
- [PEP8](https://www.python.org/dev/peps/pep-0008/) codingstyle which is considered best practise but not mandatory.
- This add-on repository has automated code guideline check which could help you improve your coding. You can find the results of these check at [Codacy](https://www.codacy.com/app/Kodi/repo-scripts/dashboard). You can create your own account as well to continuously monitor your python coding before submitting to repo.
- Development questions can be asked in the [add-on development](http://forum.kodi.tv/forumdisplay.php?fid=26) section on the Kodi forum.

PR #1455, opened 2020-04-13
[script.transmission] broken

enen92:
Addon is violating our piracy policy due to the inclusion of search functionality on trackers known to contain copyrighted material.

@basilgello you're free to submit an update which removes the search functionality. 
Unfortunately this slipped between previous reviewers fingers, long long time ago.


basilgello:
@enen92 just removing search.py without removing gui part will break UI. See my last force-push of #1454 or if you want me to do complete the cleanup here, I can do it in the morning.

PR #1458, opened 2020-04-13
[script.module.cryptography] removal

ronie:
as mentioned here: https://github.com/xbmc/repo-scripts/pull/1388#issuecomment-611432710
this addon has a C extention and is not suitable for our repo.

my bad, mea culpa.


basrieter:
Should we not include one with a higher version and mark it as broken? Or do we need to fix this in a different way?

ronie:
i think the changes are very small that someone has this module installed...
there are up to this point no addons in the matrix repo that require it (or script.module.requests_oauthlib or script.module.pymysql)

PR #1466, opened 2020-04-15
[weather.ozweather] 1.0.2

bossanova808:
### Description

Minor update to deal with upstream scraper changes.

There is some whitespace changes in here too - sorry, it's the beginnings of some Python 3 / Leia work.

(I would appreciate if this could just get merged as is to keep folks getting their weather - this is a Python 2 version, for Krypton and below.... I will prepare a Python 3 / Leia version separately in due course). 

### Checklist:


enen92:
@bossanova808 gotham-leia addons won't be visible in matrix. Take your time in the conversion phase, no rush :)

PR #1488, opened 2020-04-25
script.skinvariables 0.0.3

jurialmunkey:
### Description

Helper addon for Skinners to construct variables for multiple containers and/or listitem positions based on a simple json template.

### Checklist:

Additional information :


script.skinvariables/.gitignore
```diff
@@ -0,0 +1,63 @@
+# Created by .gitignore support plugin (hsz.mobi)
```
enen92, line 1: remove

script.skinvariables/addon.xml
```diff
@@ -0,0 +1,21 @@
...
+provider-name="jurialmunkey">
+<requires>
+  <import addon="xbmc.python" version="2.26.0"/>
+</requires>
+<extension point="xbmc.python.script" library="script.py" />
+<extension point="xbmc.addon.metadata">
+  <summary lang="en_GB">Skin Variables helps skinners to construct variables for multiple containers and listitems from a template</summary>
+  <description lang="en_GB">Skin Variables helps skinners to construct variables for multiple containers and listitems from a template</description>
+  <disclaimer lang="en_GB">For skinners</disclaimer>
+  <license>GPL-3.0</license>
```
enen92, line 14: proper spdx identifier

script.skinvariables/resources/lib/script.py
```diff
@@ -0,0 +1,129 @@
+# -*- coding: utf-8 -*-
```
enen92, line 1: you should really try to make this file as smaller as possible (the only file that is not compiled to bytecode as this is an entrypoint). As this is a skin helper script it is called all the type.

Can you refactor to make something like:

```python
import sys
from resources.lib import myscriptlogic

myscriptlogic.run(sys.argv)
```

script.skinvariables/resources/lib/script.py
```diff
@@ -0,0 +1,129 @@
...
+                arg_split = arg.split('=', 1)
+                if arg_split[0] and arg_split[1]:
+                    key, value = arg_split
+                    self.params.setdefault(key, value)
+            else:
+                self.params.setdefault(arg, True)
+
+    def make_variables(self):
+        p_dialog = xbmcgui.DialogProgressBG()
+        p_dialog.create('Skin Variables', 'Constructing Variables...')
```
enen92, line 33: non localized strings are not allowed in the ui

script.skinvariables/resources/lib/script.py
jurialmunkey: I'm confused... isn't this what the other script.py file does:
https://github.com/xbmc/repo-scripts/blob/b81e1700f8b982e34fb4a69e3aee44c5c28f260f/script.skinvariables/script.py

There's two files called `script.py` - one in the basedir and one in `resources/lib/`
The one in the basedir is super simple and is literally just:
```python
from resources.lib.script import Script

if __name__ == '__main__':
    TMDbScript = Script()
    TMDbScript.get_params()
    TMDbScript.router()
```

script.skinvariables/resources/lib/script.py
jurialmunkey: I'm happy to localize. It just felt like overkill to localize something that no end user should ever see -- only skinners should be running this script.

Usage pattern is:
- Skinner runs the script to generate a `script-skinvariables-includes.xml` file
- Skinner commits the generated file to their repo

There is no reason for this script to run on an end-user machine. The generated file is the same for every user and the script doesn't do anything once the file has been generated.

I only made it an addon to simplify things a bit for skinners because it means you can add something like
`<onload condition="System.HasAddon(script.skinvariables)">Runscript(script.skinvariables)</onload>` to the Home.xml and then it will auto rebuild whenever the template file changes. Since the end-user has no reason to install the script, it will never run for them.

I'm not sure if only being for skinners and not end-users means that it maybe shouldn't be on the repo?

script.skinvariables/resources/lib/script.py
enen92: Disresgard, I though this was the entrypoint. All good

script.skinvariables/resources/lib/script.py
enen92: ok all good

PR #1495, opened 2020-04-26
[script.module.qrcode] 6.1.0+matrix.3

SchapplM:
This addon was imported in #1289 and #1407 to matrix from previous Kodi versions (#1171 in isengard). From my assumption for matrix never tested by clean install.
The dependency script.module.pil does not exist in matrix, but is not needed anyway, since it is included in this addon (verified on Ubuntu 18.04). No current maintainer is known to me, therefore the explanation here.


enen92:
Are you sure pil is not used by the module? Or is it just optional? In case it is optional, please make pil "optional"="true" instead of simply dropping it.

PR #1519, opened 2020-04-30
[script.module.actionhandler] 1.0.4

enen92:
The fact that this addon (no longer maintained) does not define xbmc.python as a requirement makes it available in the matrix branch. This makes the addon checker to fail while validating the krypton branch because this one has a lower version (and users can't migrate). This PR makes sure it won't be available in matrix.

Found with https://github.com/xbmc/addon-check/pull/234


enen92:
Merging as the error is the "false positive" we want to fix.

PR #1583, opened 2020-06-16
[weather.ozweather@matrix] 1.2.0

bossanova808:
### Add-on details:

- **General**
  - Add-on name: Oz Weather
  - Add-on ID: weather.ozweather
  - Version number: 1.2.0

- **Code location**
  - URL: https://github.com/bossanova808/weather.ozweather
  
Weather forecast data scraped from the Australian Bureau of Meteorology via WeatherZone - www.bom.gov.au and www.weatherzone.com.au.  For full features (e.g. radar) make sure you install the replacement skin files found via the addon wiki (https://kodi.wiki/index.php?title=Add-on:Oz_Weather).

### Checklist:


bossanova808:
@enen92 Don't bother reviewing this one yet, I'm doing some more comprehensive clean-up based on your recent reviews, so will push some updates soon...

enen92:
Np, just tag me when it's ready

bossanova808:
@enen92 Ok this should be good for review now.  Thanks yet again!

weather.ozweather/resources/lib/abcvideo.py
```diff
@@ -0,0 +1,74 @@
...
+# *  http://www.gnu.org/copyleft/gpl.html
+# *
+
+import requests
+import re
+
+try:
+    from .common import log as log
+except ImportError:
+    print("\nKodi is not available -> probably unit testing")
```
enen92, line 25: Should this be used after the function declaration (it is right below)? Use log instead of print as you are "aliasing" in the method below.

weather.ozweather/resources/lib/bom.py
```diff
@@ -0,0 +1,275 @@
...
+import time
+import urllib.error
+import urllib.parse
+import urllib.request
+import urllib3
+
+try:
+    from .common import log as log
+except ImportError:
+    print("\nKodi is not available -> probably unit testing")
```
enen92, line 33: Same I'd say

weather.ozweather/resources/lib/bom.py
```diff
@@ -0,0 +1,275 @@
...
+        if filetime < twoHoursAgo:
+            log("Deleted " + str(os.path.basename(file)))
+            os.remove(file)
+
+    # rename the currently kept radar backgrounds to prevent Kodi caching issues
+    currentFiles = glob.glob(overlayLoopPath + "/*.png")
+    for file in currentFiles:
+        os.rename(file, os.path.dirname(file) + "/" + timeNow + "." + os.path.basename(file)[13:])
+
+    # We need make the directories to store stuff if they don't exist
```
enen92, line 147: I don't see any call to rmtree

weather.ozweather/resources/lib/bom.py
```diff
@@ -0,0 +1,275 @@
...
+        attempts = 0
+        success = False
+        while not success and (attempts < 20):
+            try:
+                os.makedirs(backgroundsPath)
+                success = True
+                log("Successfully created " + backgroundsPath)
+            except Exception:
+                attempts += 1
+                time.sleep(0.1)
```
enen92, line 161: Never use time, use xbmc.sleep for small values

weather.ozweather/resources/lib/bom.py
```diff
@@ -0,0 +1,275 @@
...
+        if not success:
+            log("ERROR: Failed to create directory for radar background images!")
+            return
+
+    if not os.path.exists(overlayLoopPath):
+        attempts = 0
+        success = False
+        while not success and (attempts < 20):
```
enen92, line 169: I'd probably check for Kodi abort request here

weather.ozweather/resources/lib/bom.py
```diff
@@ -0,0 +1,275 @@
...
+            try:
+                os.makedirs(overlayLoopPath)
+                success = True
+                log("Successfully created " + overlayLoopPath)
+            except Exception:
+                attempts += 1
+                time.sleep(0.1)
```
enen92, line 176: Xbmc.sleep

weather.ozweather/resources/lib/bom.py
```diff
@@ -0,0 +1,275 @@
...
+        if not os.path.isfile(overlayLoopPath + timeNow + "." + f):
+            # ignore the composite gif...
+            if f[-3:] == "png":
+                imageToRetrieve = FTPSTUB + f
+                outputFile = timeNow + "." + f
+                log("Retrieving new radar image: " + imageToRetrieve)
+                log("Output to file: " + outputFile)
+                try:
+                    radarImage = urllib.request.urlopen(imageToRetrieve)
+                    fh = open(overlayLoopPath + "/" + outputFile, "wb")
```
enen92, line 230: Use python with statement

weather.ozweather/default.py
```diff
@@ -0,0 +1,303 @@
...
+# *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+# *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# *  GNU General Public License for more details.
+# *
+# *  You should have received a copy of the GNU General Public License
+# *  along with XBMC; see the file COPYING. If not, write to
+# *  the Free Software Foundation, 675 Mass Ave, Cambridge, MA 02139, USA.
+# *  http://www.gnu.org/copyleft/gpl.html
+# *
+
```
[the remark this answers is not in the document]
bossanova808: Noted...will move other addons to that strategy over time as well..thanks for explaining the why, gives the needed motivation :)

weather.ozweather/addon.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<addon id="weather.ozweather" name="Oz Weather" version="1.2.0" provider-name="Bossanova808">
+	<requires>
+		<import addon="xbmc.python" version="3.0.0"/>
+		<import addon="script.module.pil" version="1.1.7"/>
```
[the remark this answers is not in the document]
bossanova808: Well, I can't see it the module here - https://github.com/xbmc/repo-scripts/tree/matrix ?  That's where it would be, right?  (That's where I have been grabbing version strings from, as honestly I'm a bit confused by it all, but that seems to work).

Although I thought I'd read something about PIL matrix somewhere too...accoriding to this it is 'fixed' - https://forum.kodi.tv/showthread.php?tid=348400 - but I don't know what version string I should be using..

(I can say the above import works, for now, at least)

weather.ozweather/resources/lib/abcvideo.py
bossanova808: So this is the only thing I am a bit stuck on, I think.

I really want to be able to keep unit testing these sub-modules (abc, bom, weatherzone .py) - because it makes dealing with required upstream changes (i.e. scraping changes) _very_ much easier and quicker.  Hence this little bit of code basically will raise the exception if the module is run outside of Kodi and then, in only that case, defines a log function (that is just print).

For the moment I'll add an explanatory comment above this bit of code, but more than happy to accept suggestions for a better approach that lets me maintain the unit testing capability but is better for Kodi and/or more pythonic

(In case you haven't guessed, I am a Python hacker at best!)

bossanova808:
Have done most of that, couple questions above, though, if you wouldn't mind.

weather.ozweather/addon.xml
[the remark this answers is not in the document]
enen92: PIL is a python binary module so it is in kodi source: https://github.com/xbmc/xbmc/pull/17794

weather.ozweather/resources/lib/abcvideo.py
enen92: that's fine

weather.ozweather/addon.xml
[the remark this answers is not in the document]
enen92: It's fine as is since kodi addons require a minimum or higher version than x.x.x when you define them as dependencies. Keep it in mind for the future, if you addon works well with the current `7.1.2` and this version is available, it is a good practice to lock the import to that specific module. That way we know which version is supposed to work or not :)

PR #1589, opened 2020-06-16
[script.module.exifread] 2.1.2+matrix.0

ronie:
python library to read exif tags from pictures.
needed by: https://github.com/xbmc/repo-scripts/pull/1588


enen92:
Should have been matrix.1 for consistency but it is fine as is

PR #1614, opened 2020-07-04
[script.service.janitor] 4.4.1

Anthirian:
### Description
- [FIX] Improved unicode support
- [FIX] Correctly match excluded paths that contain user credentials
### Checklist:

Additional information :

### DIFF
--- a/script.service.janitor/addon.xml
+++ b/script.service.janitor/addon.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<addon id="script.service.janitor" name="Janitor" version="4.4.0" provider-name="Anthirian, drewzh">
+<addon id="script.service.janitor" name="Janitor" version="4.4.1" provider-name="Anthirian, drewzh">
     <requires>
         <import addon="xbmc.python" version="2.26.0" />
         <import addon="xbmc.json" version="9.7.2" />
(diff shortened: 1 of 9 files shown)


script.service.janitor/addon.xml
```diff
@@ -39,15 +39,16 @@ Cet addon à la faculté de supprimer entièrement votre collection vidéo! Assu
         <disclaimer lang="nl_NL">Het is verstandig om videos pas permanent te verwijderen als je bekend bent met alle instellingen.</disclaimer>
         <disclaimer lang="fr_FR">Il est recommandé de ne pas supprimer physiquement de vidéos avant de bien maitriser les réglages.</disclaimer>
         <platform>all</platform>
+        <language>en nl fr</language>
```
enen92, line 42: the addon does not provide content, please remove the language tag

script.service.janitor/addon.xml
```diff
@@ -39,15 +39,16 @@ Cet addon à la faculté de supprimer entièrement votre collection vidéo! Assu
...
         <license>GNU GENERAL PUBLIC LICENSE. Version 3, 29 June 2007</license>
```
enen92, line 43: would be nice if you could use the spdx identifier for the license instead (GPL-3.0-only or GPL-3.0-or-later depending on your license

Anthirian:
I have fixed all the issues with the .po files that I could, but still the CI bot errors out. Currently there is no way for me to avoid this construct in my code:

```
# Progress dialog translations. Keep the text between brackets as is! All dialogs have three lines we need to fill.
msgctxt "#32615"
msgid " "
msgstr "" # One space is required to remove line from previous dialog```

This is due to the way the progress dialogs work in Leia. Unfortunately if you supply a multi line text as multiple arguments and then refresh the contents of the dialog, you are forced to write at least _something_. An empty string will not cause a redraw of a line, which means I am forced to write a space in such a case. This single space is not parsed correctly by the CI bot unfortunately.

**Comment by anxdpanic** [comment]
@Anthirian 
The current issue is
```
msgstr "" # One space is required to remove line from previous dialog
```
Move the comment to above msgctxt
`"` is expected to be end of line for these cases

Anthirian:
> 
> 
> @Anthirian
> The current issue is
> 
> ```
> msgstr "" # One space is required to remove line from previous dialog
> ```
> 
> Move the comment to above msgctxt
> `"` is expected to be end of line for these cases

Thanks, that helped actually. Are there any guidelines for what the bot expects with respect to this? I didn't run into these issues last time I pushed an update and nothing changed in terms of translations this time.

enen92:
@Anthirian it is a known issue in addon-checker/polib regarding translator comments. We're tracking it here: https://github.com/xbmc/addon-check/issues/237

Unfortunately no fix yet. Thanks for adjusting though (to avoid the error)

PR #1617, opened 2020-07-06
[script.kodi.windows.update] v.1.3.2

Lunatixz:
### Checklist:

Additional information :


script.kodi.windows.update/addon.xml
```diff
@@ -0,0 +1,31 @@
...
+        <import addon="script.module.kodi-six" version="0.1.3.1"/>
+        <import addon="script.module.simplecache" version="1.0.0"/>
+        <import addon="script.module.beautifulsoup4" version="4.3.2"/>
+    </requires>
+    <extension point="xbmc.python.script" library="default.py"/>
+    <extension point="xbmc.service" library="service.py"/>
+    <extension point="xbmc.addon.metadata">
+        <summary lang="en_GB">Download and Install Kodi packages.</summary>
+        <description lang="en_GB">Download and Install Kodi window executables; select from  Nightlies, Releases, Snapshots and Test-builds.</description>
+        <license>GNU GENERAL PUBLIC LICENSE. Version 3, June 2007</license>
```
enen92, line 15: SPDX license identifier if possible (GPL-3.0-or-later or GPL-3.0-only depending on your interest)

script.kodi.windows.update/addon.xml
```diff
@@ -0,0 +1,31 @@
...
+        <platform>windows</platform>
+        <website>http://kodi.tv</website>
+        <source>https://github.com/Lunatixz/KODI_Addons/tree/master/script.kodi.windows.update</source>
+        <forum>https://forum.kodi.tv/showthread.php?tid=325226</forum>
+        <disclaimer lang="en_GB">UWP not supported! Not intended for Kodi installation through the Windows Store.</disclaimer>
```
enen92, line 20: can't this be filtered in the platform tag? I updated the values a while ago based on the code: https://kodi.wiki/view/Addon.xml#.3Cplatform.3E

script.kodi.windows.update/default.py
```diff
@@ -0,0 +1,249 @@
+#     Copyright (C) 2020 Team-Kodi
```
enen92, line 1: spdx headers would be nice, look at the kodi source code for examples

script.kodi.windows.update/default.py
```diff
@@ -0,0 +1,249 @@
...
+            percent = min(numblocks * blocksize * 100 / filesize, 100) 
+            currently_downloaded = float(numblocks) * blocksize / (1024 * 1024) 
+            kbps_speed = numblocks * blocksize / (time.time() - start_time) 
+            if kbps_speed > 0: eta = (filesize - numblocks * blocksize) / kbps_speed 
+            else: eta = 0 
+            kbps_speed = kbps_speed / 1024 
+            if eta < 0: eta = divmod(0, 60)
+            else: eta = divmod(eta, 60)
+            total   = (float(filesize) / (1024 * 1024))
+            label   = '[B]Downloading:[/B] %s'%SETTINGS_LOC
```
enen92, line 227: needs to be localized

script.kodi.windows.update/default.py
```diff
@@ -0,0 +1,249 @@
...
+            label2  = '%.02f MB of %.02f MB'%(currently_downloaded,total)
+            label2 += ' | [B]Speed:[/B] %.02f Kb/s'%kbps_speed
```
enen92, line 229: same

script.kodi.windows.update/default.py
```diff
@@ -0,0 +1,249 @@
...
+            label2 += ' | [B]ETA:[/B] %02d:%02d'%eta
```
enen92, line 230: same

script.kodi.windows.update/default.py
```diff
@@ -0,0 +1,249 @@
...
+            dia.update(int(percent), '%s\n%s\n%s'%(label,fle,label2))
+        except Exception as e: 
+            log("pbhook failed! %s" + str(e), xbmc.LOGERROR)
+            dia.update(100)
+        if dia.iscanceled(): raise Exception
+
+
+    def installEXE(self, exefile):
+        if not xbmcvfs.exists(exefile): return
+        xbmc.executebuiltin('XBMC.AlarmClock(shutdowntimer,XBMC.Quit(),2.0,true)')
```
enen92, line 240: please drop the XBMC. prefix, it has been removed in matrix. Use only AlarmClock

script.kodi.windows.update/default.py
```diff
@@ -0,0 +1,249 @@
...
+        self.killKodi.start()
+        subprocess.call(exefile, shell=True)
+        
+        
+    def killME(self):
+        subprocess.call('taskkill /f /im kodi.exe')
+        
+        
+if __name__ == '__main__': Installer()
```
enen92, line 249: missing line at the end of the file

enen92, line 249: would be nice if you could move the installer somewhere else to make the plugin faster

script.kodi.windows.update/service.py
```diff
@@ -0,0 +1,91 @@
...
+        count = 0
+        try:
+            while not self.myMonitor.abortRequested() and count < 15:
+                count += 1 
+                if self.myMonitor.waitForAbort(1): return
+                build = (xbmc.getInfoLabel('System.OSVersionInfo') or 'busy')
+                if build.lower() != 'busy': return REAL_SETTINGS.setSetting("Version",str(build))
+        except Exception as e: log("getVersion Failed! " + str(e), xbmc.LOGERROR)
+              
+if __name__ == '__main__': Service()
```
enen92, line 91: missing new line at the end of the file

enen92, line 91: same, would be nice if the Service() could be moved to the resources/lib folder

Lunatixz: THX for the suggestion; I have to wonder why this would be needed? Kodi'spython interpreter will initialize all modules regardless to its folder placement.

Lunatixz: is this is a false positive? linter isn't understanding the single line if statement?

script.kodi.windows.update/default.py
Lunatixz: Thanks, I read about this change... however, I forgot to adopt it; good eye!

script.kodi.windows.update/default.py
Lunatixz: Can you send me a link with the available spdx headers?

script.kodi.windows.update/addon.xml
Lunatixz: Those new? I guess 'windows-x86_64' would apply here. I'll make the change.

enen92: Better to test, I don't have a windows machine. Those are not new, just not documented :)

script.kodi.windows.update/default.py
enen92: Any CPP file will do, see for example https://github.com/xbmc/xbmc/blob/master/xbmc/AppInboundProtocol.h#L1-L7
Make sure you adapt to your LICENSE path instead of a direct copy-paste (`LICENSES/README.md`)

script.kodi.windows.update/service.py
enen92: github shows a sign whenever this happens. It's no big deal though

enen92: python interpreter will not compile the entrypoint files to bytecode, only those that are imported. Means that after the first execution of the addon, subsquent executions will be faster.

Lunatixz:
@enen92 Looks like the linter needs updating?
```
failed validating 'windows-x86_64' with XsdEnumerationFacets("['linux', 'osx', 'osx64', 'osx32', 'ios', ...]"):

Reason: invalid value 'windows-x86_64', it must be one of ['linux', 'osx', 'osx64', 'osx32', 'ios', 'windx', 'windows', 'windowsstore', 'android', 'freebsd', 'all']

Schema:

  <xs:enumeration xmlns:xs="http://www.w3.org/2001/XMLSchema" value="linux" />

Instance:

  <platform>windows-x86_64</platform>

Path: /extension/platform
```

script.kodi.windows.update/addon.xml
enen92: I think what you're looking for is windx. This means windows without uwp (windowsstore). But yeah those values need to be updated on the checker

enen92:
@lunatixz answered in line (I think you can use windx althought the other values should be also added to the checker)

Lunatixz:
> @Lunatixz answered in line (I think you can use windx althought the other values should be also added to the checker)

Ahh haven't tested 'windx' can we merge as is; and I'll look into the platform another time. THX

.gitignore
```diff
@@ -0,0 +1 @@
+*.db
```
enen92, line 1: Remove from submission

Lunatixz: THX for the review; I understand its under-appreciated, and time-consuming.

I don't mind making changes however, there is way too many rule inconsistency to keep up with. I can't micromanage every commit; can we generate a single "RULES" list... THX

https://github.com/xbmc/repo-scripts/pull/660#discussion_r151667266


I'm leaving this commit as is, merge or don't:

enen92: Sorry lunatixz I understand the frustration. Martijn doesn't review addons since ~ 2017 and other people have picked up reviews and the repo curation and opinions differ.

Furthermore, your addon has "Team Kodi" as author, so we can't be sending the wrong message and should play by the book. To be honest, an addon that is authored by team kodi normally means it should be available from github.com/xbmc organization, but I'm not even considering it here as you are a team member.

You shouldn't be micromanaging commits, you've now options to submit the addon automatically and exclude files from the submission using gitattributes (however it is designed for single repo addons - https://github.com/xbmc/action-kodi-addon-submitter)

Lunatixz:
@enen92 sorry for the delay... I completely forgot about these PR's. I'm getting too old! LOL

PR #1630, opened 2020-07-13
[script.harmony.control] 0.2.0

pkscout:
### Description
This addon allows a person to create actions that control a Harmony Hub by running Harmony Activities and remote commands. Without arguments, the script displays a menu of actions. With arguments, the script will run a specific action.

### Checklist:

- [x] My code follows the [add-on rules](http://kodi.wiki/view/Add-on_rules) and [piracy stance](http://kodi.wiki/view/Official:Forum_rules#Piracy_Policy) of this project. 
- [x] I have read the [CONTRIBUTING](https://github.com/xbmc/repo-scripts/blob/master/CONTRIBUTING.md) document
- [x] Each add-on submission should be a single commit with using the following style: [script.foo.bar] 1.0.0

Additional information :
- Submitting your add-on to this specific branch makes it available to any Kodi version equal or higher than the branch name with the applicable Kodi dependencies limits.
- [add-on development](http://kodi.wiki/view/Add-on_development) wiki page.
- Kodi [pydocs](http://kodi.wiki/view/PyDocs) provide information about the Python API
- [PEP8](https://www.python.org/dev/peps/pep-0008/) codingstyle which is considered best practise but not mandatory.
- This add-on repository has automated code guideline check which could help you improve your coding. You can find the results of these check at [Codacy](https://www.codacy.com/app/Kodi/repo-scripts/dashboard). You can create your own account as well to continuously monitor your python coding before submitting to repo.
- Development questions can be asked in the [add-on development](http://forum.kodi.tv/forumdisplay.php?fid=26) section on the Kodi forum.

PR #1631, opened 2020-07-13
[script.harmony.control] 0.2.0+matrix.1

pkscout:
### Description
This addon allows a person to create actions that control a Harmony Hub by running Harmony Activities and remote commands. Without arguments, the script displays a menu of actions. With arguments, the script will run a specific action.

### Checklist:

Additional information :


script.harmony.control/addon.xml
```diff
@@ -0,0 +1,25 @@
+<addon id="script.harmony.control" name="Harmony Hub Controller" provider-name="pkscout" version="0.1.105">
```
basrieter, line 1: Why not use 0.1.5+matrix.1 and keep it in line with https://github.com/xbmc/repo-scripts/pull/1630. That is the preferred way.

pkscout: If it's OK, I can change the numbering for the next release. It's possible someone grabbed this from my beta repo, and if I change now they won't get updated (unless I bump the minor too, but then I also need to update the PR for the Leia version so they're in sync).

basrieter: In that case I would suggest you to use `0.1.105` for Leia and `0.1.105+matrix.1` for Matrix in these two PR's?

basrieter:
I noticed there are some `__init__.py` files missing compared to the Leia version? In `lib` and `lib/api`? Is that on purpose?

pkscout:
> I noticed there are some `__init__.py` files missing compared to the Leia version? In `lib` and `lib/api`? Is that on purpose?

Yes, that's on purpose. Since in Python3 you don't need empty __init.py__ files to have folders treated as packages, my Matrix stuff I only includes them if I actually need to do something in them.

PR #1709, opened 2020-10-07
[script.cu.lrclyrics] 5.8.3

ronie:
### Description

changes:
- add new syair scraper
- fix darklyrics scraper
- remove broken lyricwiki scraper

### Checklist:

Additional information :


script.cu.lrclyrics/resources/lib/gui.py
```diff
@@ -480,7 +480,7 @@ def get_page_lines(self):
         while xbmc.getInfoLabel('Container(110).NumPages') and (int(xbmc.getInfoLabel('Container(110).NumPages')) < 2) and (not self.Monitor.abortRequested()):
             listitem = xbmcgui.ListItem(offscreen=True)
             self.text.addItem(listitem)
-            xbmc.sleep(50)
+            xbmc.sleep(5)
```
pkscout, line 483: We've been asking folks to use `xbmc.Monitor().waitForAbort(x)` instead of `xbmc.sleep(x)` where possible.

pkscout (changes requested):

script.cu.lrclyrics/resources/lib/gui.py
ronie: why?

enen92: @pkscout that advice only applies to long sleep times (e.g. t > 5 seconds - the time kodi waits before force-killing the script ) or in the context of loops... not exactly for 50 msecs :)

pkscout (approved):

script.cu.lrclyrics/resources/lib/gui.py
enen92: btw, @ronie any reason for such a decrease on the sleep time? I assume you only loop here for a small number of iterations (and as such this is fine) otherwise you could be increasing the cpu usage too much

ronie: @enen92 it's calculating how many lines of text would fit on the screen, and it depends on the skin you're using.
but ~ 10 to 15 loops would be a good guess.
ideally i wouldn't need to use any sleep at all, but for whatever it takes a tiny bit of time before the infolabel gets updated.

the reason for the decrease is to speed things up a bit... the previous value was a bit too conservative.

PR #1732, opened 2020-10-30
[service.upnext@matrix] 1.1.3+matrix.1

im85288:
### Add-on details:

- **General**
  - Add-on name: Up Next
  - Add-on ID: service.upnext
  - Version number: 1.1.3+matrix.1

- **Code location**
  - URL: https://github.com/im85288/service.upnext
  
A service add-on that shows a Netflix-style notification for watching the next episode. After a few automatic iterations it asks the user if he is still there watching.

A lot of existing add-ons already integrate with this service out-of-the-box.

### v1.1.3 (2020-10-14)
- Enable customPlayTime by default
- Do not seek to the end before playing next episode
- Add Kodi v17 compatibility
- Fix logging on Kodi v19 (Matrix) after recent breakage
- Enqueue the next episode in the playlist

### v1.1.2 (2020-06-22)
- Bug fix for change from sleep to waitForAbort using seconds and not ms

### v1.1.1 (2020-06-21)
- Avoid conflict with external players
- Restore "Ignore Playlist" option
- Fix a known Kodi bug related to displaying hours
- Improvements to endtime visualization
- New translations for Hindi and Romanian
- Translation updates to Hungarian and Spanish

### v1.1.0 (2020-01-17)
- Add notification_offset for Netflix add-on
- Fix various runtime exceptions
- Implement new settings
- Implement new developer mode
- Show current time and next endtime in notification
- New translations for Brazilian, Czech, Greek, Japanese, Korean
- New translations for Russian, Slovak, Spanish, Swedish
- Translation updates to Croatian, French, German, Hungarian, Italian, Polish
        

### Checklist:

### DIFF
--- a/service.upnext/resources/lib/script.py
+++ b/service.upnext/resources/lib/script.py
@@ -5,7 +5,7 @@
 from __future__ import absolute_import, division, unicode_literals
 from datetime import datetime, timedelta
 from math import ceil
-from xbmc import Monitor, sleep
+from xbmc import Monitor
 from xbmcgui import WindowXMLDialog
 from statichelper import from_unicode
 from utils import addon_path, get_setting_bool, localize, localize_time
@@ -88,12 +88,13 @@ def test_popup(window):
     popup.show()
     step = 0
     wait = 100
+    wait_s = wait / 1000
     timeout = 10000
     monitor = Monitor()
     while popup and step < timeout and not monitor.abortRequested():
         if popup.pause:
             continue
-        sleep(wait)
+        monitor.waitForAbort(wait_s)
         popup.update_progress_control(timeout, wait)
         step += wait
 
--- a/service.upnext/README.md
+++ b/service.upnext/README.md
@@ -15,7 +15,7 @@ The add-on has various settings to fine-tune the experience, however the default
 
   * Simple or fancy mode (defaults to fancy mode, but a more simple interface is possible)
   * The notification time can be adjusted (defaults to 30 seconds before the end)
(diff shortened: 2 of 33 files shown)


service.upnext/resources/lib/script.py
```diff
@@ -88,12 +88,13 @@ def test_popup(window):
...
     step = 0
     wait = 100
+    wait_s = wait / 1000
     timeout = 10000
     monitor = Monitor()
     while popup and step < timeout and not monitor.abortRequested():
         if popup.pause:
             continue
-        sleep(wait)
+        monitor.waitForAbort(wait_s)
```
enen92, line 97: Any reason for this change? Sleeping for 0.1 seconds doesn't necessarily require you to use waitforabort

enen92 (approved):

service.upnext/resources/lib/script.py
dagwieers: cc @MoojMidge

MoojMidge: It was an attempt to prevent Kodi from hanging when the test popup was created and the addon was restarted or Kodi was shutdown. 

Couldn't figure out why a 0.1s sleep would cause the hang, but decided to try waitForAbort in case there was some interaction between sleep and abortRequested that was causing the problem. 

Turns out the actual problem was that Kodi wasn't stopping scripts on shutdown and so the abortRequested call wasn't actually doing anything, which is fixed in https://github.com/xbmc/xbmc/pull/18305

Retained the call to waitForAbort as it seemed slightly more idiomatic to only use the Monitor methods going forward.

PR #1764, opened 2020-11-27
[service.subtitles.supersubtitles@matrix] 0.0.25

fape:
### Description
- Update python dependency to 3.0.0. 
- remove python 2.x support, dependency lib too 
### Checklist:


service.subtitles.supersubtitles/service.py
```diff
@@ -0,0 +1,533 @@
...
+    "szlovák": "Slovak",
+    "szlovén": "Slovenian",
+    "spanyol": "Spanish",
+    "svéd": "Swedish",
+    "török": "Turkish",
+}
+
+EPISODE_REGEXP = re.compile(r'S?(?P<season>\d+)([x_-]|\.)*E?(?P<episode>\d+)', re.IGNORECASE)
+
+def recreate_dir(path):
```
enen92, line 128: Since this is only used for the temp dir can you re-write this in a way that it receives no arguments and specifically processes that folder only?

enen92 (commented):
only one "minor" I'd like to see changed, and it's good to go. Will you still support the version in leia?

enen92 (approved):

fape:
Thanks. I change to no args.
[Yes leia support reamins too](https://github.com/xbmc/repo-scripts/tree/leia/service.subtitles.supersubtitles). But there was no change in logic. I just bumped python dependency and removed script.module.kodi-six (python2 and python3 support lib) dependency .

enen92:
Thanks for the update @fape

fape:
thanks

PR #1793, opened 2020-12-22
[script.module.kodiswift] 0.0.9

tingjhenjiang:
### Description

1. enable script.module.kodiswift to support python 2 and python 3 by incorporating the six package.
2. add supports on subtitles.

### Checklist:

### DIFF
--- a/script.module.kodiswift/addon.xml
+++ b/script.module.kodiswift/addon.xml
@@ -1,18 +1,15 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<addon id="script.module.kodiswift" name="kodiswift" provider-name="Aaron Frase (Sinap)" version="0.0.8">
+<addon id="script.module.kodiswift" name="kodiswift" provider-name="Aaron Frase (Sinap); TJ Jiang" version="0.0.9">
   <requires>
     <import addon="xbmc.python" version="2.24.0"/>
   </requires>
   <extension library="lib" point="xbmc.python.module"/>
   <extension point="xbmc.addon.metadata">
     <summary lang="en">A fork of the xbmcswift2 project</summary>
-    <description lang="en">A micro framework to enable rapid development of Kodi plugins.</description>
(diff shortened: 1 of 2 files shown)


script.module.kodiswift/addon.xml
```diff
@@ -11,8 +11,5 @@
     <license>GNU General Public License v3 (GPL-3)</license>
```
enen92, line 11: GPL-3.0-only or GPL-3.0-or-later (spdx identifiers)

script.module.kodiswift/addon.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<addon id="script.module.kodiswift" name="kodiswift" provider-name="Aaron Frase (Sinap)" version="0.0.8">
+<addon id="script.module.kodiswift" name="kodiswift" provider-name="Aaron Frase (Sinap)" version="0.0.9">
```
enen92, line 2: Looks like you're not the author of the addon...

tingjhenjiang:
OK, I've made corresponding modification in terms of [GNU 3.0](https://www.gnu.org/licenses/gpl-3.0.html):
1. add my own name to the provider-name in addon.xml.
2. describe the modifications is subject to me in addon.xml
3. add my email  in addon.xml.

> To “modify” a work means to copy from or adapt all or part of the work in a fashion requiring copyright permission, other than the making of an exact copy. The resulting work is called a “modified version” of the earlier work or a work “based on” the earlier work.

> For the developers' and authors' protection, the GPL clearly explains that there is no warranty for this free software. For both users' and authors' sake, _the GPL requires that modified versions be marked as changed, so that their problems will not be attributed erroneously to authors of previous versions._

> Notwithstanding any other provision of this License, for material you add to a covered work, you may... c) Prohibiting misrepresentation of the origin of that material, or requiring that modified versions of such material be marked in reasonable ways as different from the original version; or

2.

enen92 (approved):

PR #1829, opened 2021-01-29
[script.tvmaze.scrobbler@matrix] 1.1.0+matrix.1

romanvm:
### Add-on details:

- **General**
  - Add-on name: TVmaze Scrobbler/Tracker
  - Add-on ID: script.tvmaze.scrobbler
  - Version number: 1.1.0+matrix.1

- **Code location**
  - URL: https://github.com/romanvm/kodi.tvmaze.scrobbler
  
Automatically track all TV episodes you are watching to TVmaze. A must have if you want to sync your watch history between various applications.
Sign up for a free account at https://tvmaze.com for more features.

This Kodi TV episode Tracker uses the TVmaze.com user API's scrobbler endpoints(https://static.tvmaze.com/apidoc/)

What this addon does:
- Performs an initial sync between Kodi and TVmaze for the episodes you've watched.
- If you add an episode to the Kodi library it marks it as "acquired" on TVmaze.
- If you watch an episode in Kodi it marks it as "watched" in TVmaze.
- If you mark an episode as watched in TVmaze it marks it as watched in Kodi.

### Description of changes:

1.1.0:
- Added periodic pulling of episode watched statuses from TVmaze.
- Improved error logging.

1.0.1:
- Fixed crashes because of malformed JSON responses from TVmaze.
- Fixed crashes because of invalid SSL certificates.

### Checklist:

### DIFF
--- a/script.tvmaze.scrobbler/service.py
+++ b/script.tvmaze.scrobbler/service.py
@@ -18,10 +18,13 @@
 
 from __future__ import absolute_import, unicode_literals
 
-from libs.kodi_monitor import UpdateMonitor
+from libs.exception_logger import log_exception
+from libs.kodi_monitor import KodiMonitor
 from libs.kodi_service import logger
+from libs.scheduled_tasks import periodic_pull
 
-MONITOR = UpdateMonitor()
-logger.info('Service started')
-MONITOR.waitForAbort()
-logger.info('Service stopped')
+with log_exception():
+    monitor = KodiMonitor()
+    while not monitor.waitForAbort(3.0):
+        periodic_pull()
+    logger.info('Service stopped')
--- a/script.tvmaze.scrobbler/addon.xml
+++ b/script.tvmaze.scrobbler/addon.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon id="script.tvmaze.scrobbler"
    name="TVmaze Scrobbler/Tracker"
-   version="1.0.1+matrix.1"
+   version="1.1.0+matrix.1"
    provider-name="Roman V.M.">
   <requires>
     <import addon="xbmc.python" version="3.0.0" />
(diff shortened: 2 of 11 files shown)


script.tvmaze.scrobbler/service.py
```diff
@@ -18,10 +18,18 @@
...
-logger.info('Service started')
-MONITOR.waitForAbort()
-logger.info('Service stopped')
+with log_exception():
+    xbmc.sleep(3000)
+    monitor = KodiMonitor()
+    logger.info('Service started')
+    while not monitor.abortRequested():
+        periodic_pull()
+        xbmc.sleep(60000)
```
enen92, line 34: monitor.waitForAbort(60)

script.tvmaze.scrobbler/service.py
```diff
@@ -18,10 +18,18 @@
...
+from libs.kodi_monitor import KodiMonitor
 from libs.kodi_service import logger
+from libs.scheduled_tasks import periodic_pull
 
-MONITOR = UpdateMonitor()
-logger.info('Service started')
-MONITOR.waitForAbort()
-logger.info('Service stopped')
+with log_exception():
+    xbmc.sleep(3000)
```
enen92, line 29: Maybe you can also consider changing this one (although the pause time is small)

script.tvmaze.scrobbler/service.py
romanvm: The idea is to do a pull in 3s after Kodi start if a pull interval has passed. With your variant a pull will be executed in 60s after Kodi start. If you launch Kodi, quickly start some video and then immediately close Kodi after watching (as I often do on my Android tablet) a pull may never be executed at all (unless you enable "Pull during playback").

Lunatixz: `if monitor.waitForAbort(60): break`
Requires a break in the while loop...

romanvm: For some reason I was sure that `xbmc.sleep()` checks abort flag and exits prematurely if it is set. I don't know where I got that idea from. I have checked the Kodi code and this is not the case. So I'll refactor this in a different way.

romanvm:
Updated the code.

PR #1857, opened 2021-02-21
[service.subsmangler] 2.0.0

bkiziuk:
### Description
Kodi Media Center service plugin aiming to be a comprehensive subtitle helper.
New add-on version for Kodi Matrix.

### Checklist:
- [ X ] My code follows the [add-on rules](http://kodi.wiki/view/Add-on_rules) and [piracy stance](http://kodi.wiki/view/Official:Forum_rules#Piracy_Policy) of this project. 
- [ X ] I have read the [CONTRIBUTING](https://github.com/xbmc/repo-scripts/blob/master/CONTRIBUTING.md) document
- [ X ] Each add-on submission should be a single commit with using the following style: [script.foo.bar] 1.0.0


service.subsmangler/.gitignore
```diff
@@ -0,0 +1,74 @@
+# Covers JetBrains IDEs: IntelliJ, RubyMine, PhpStorm, AppCode, PyCharm, CLion, Android Studio and WebStorm
```
enen92, line 1: do not include this in the submission

service.subsmangler/addon.xml
```diff
@@ -0,0 +1,46 @@
...
+        <disclaimer lang="de_DE">Ein umfassendes Tool für Untertitel</disclaimer>
+        <disclaimer lang="en_GB">A comprehensive subtitle companion</disclaimer>
+        <disclaimer lang="hu_HU">Egy teljeskörű felirat asszisztens</disclaimer>
+        <disclaimer lang="pl_PL">Wszechstronny asystent napisów</disclaimer>
+        <disclaimer lang="sk_SK">Kompletný sprievodca titukly</disclaimer>
+        <news>v2.0.0 (2021-02-22)
+- first release for Kodi 19 Matrix
+</news>
+        <platform>all</platform>
+        <license>GNU General Public License v3.0</license>
```
enen92, line 38: GPL-3.0-only or GPL-3.0-or-later

service.subsmangler/resources/lib/__init__.py
```diff
@@ -0,0 +1 @@
+# Dummy file to make this directory a package.
```
enen92, line 1: you probably know this, but python3 doesn't require those init placeholder files anymore

service.subsmangler/resources/lib/pysubs2/__main__.py
```diff
@@ -0,0 +1,7 @@
+import sys
```
enen92, line 1: Please do not bundle python modules inside your addon. Can you submit pysubs2 as a script.module? Is it plain upstream or have you patched the module?

enen92 (changes requested):
See comments

service.subsmangler/.gitignore
bkiziuk: Got it.

service.subsmangler/resources/lib/pysubs2/__main__.py
bkiziuk: At the moment it is a plain upstream. This module is borrowed from https://github.com/tkarabela/pysubs2
At the moment there is no Kodi pysubs2 module I know of. As I'm not the author of this module, I wanted to avoid maintaining it as there can be other plugins can be based on it in the future.

service.subsmangler/resources/lib/__init__.py
bkiziuk: Frankly speaking, I wasn't aware of this.
But as I look in Python 3.8 documentation [https://docs.python.org/3.8/tutorial/modules.html#packages](url) 
it still mentions this usage.
Let me look into it. Looks like more reading is needed on PEP420.

bkiziuk: Ok. I've removed __init__.py from resources/ and resources/lib

service.subsmangler/resources/lib/pysubs2/__main__.py
enen92: you don't need to maintain it (if it is plain upstream) but we do request to have it as a separate module

enen92 (approved):

PR #1890, opened 2021-02-28
[script.myepisodes] 3.0.0

maximeh:
### Description
Switch MyEpisodes to Python3 and adapt to support new changes in Matrix
### Checklist:


script.myepisodes/addon.xml
```diff
@@ -0,0 +1,26 @@
...
+    <import addon="xbmc.python" version="3.0.0"/>
+    <import addon="script.module.beautifulsoup4" version="4.6.3"/>
+    <import addon="script.module.requests" version="2.22.0"/>
+  </requires>
+  <extension point="xbmc.service" library="default.py"/>
+  <extension point="xbmc.addon.metadata">
+    <summary lang="en_GB">MyEpisodes</summary>
+    <description lang="en_GB">Automatically set watched episodes on MyEpisodes.com</description>
+    <platform>all</platform>
+    <language>en fr it pt es</language>
```
enen92, line 16: please drop, this only applies to actual content (media files)

script.myepisodes/addon.xml
```diff
@@ -0,0 +1,26 @@
...
+    <website>https://www.myepisodes.com</website>
+    <license>GNU GENERAL PUBLIC LICENSE. Version 2, June 1991</license>
```
enen92, line 18:
```suggestion
    <license>GPL-2.0-only</license>
```

or

```suggestion
    <license>GPL-2.0-or-later</license>
```

(SPDX identifiers)

script.myepisodes/default.py
```diff
@@ -0,0 +1,234 @@
...
+        filename_full_path = self.getPlayingFile()
+        # We don't want to take care of any URL because we can't really gain
+        # information from it.
+        self.is_excluded = False
+        if utils.is_excluded(filename_full_path):
+            self.is_excluded = True
+            self.tearDown()
+            return
+
+        # Try to find the title with the help of XBMC (Theses came from
```
enen92, line 152:
```suggestion
        # Try to find the title with the help of Kodi (Theses came from
```

:)

script.myepisodes/default.py
```diff
@@ -0,0 +1,234 @@
...
+        # XBMC.Subtitles add-ons)
```
enen92, line 153:
```suggestion
        # Kodi.Subtitles add-ons)
```

script.myepisodes/resources/__init__.py
```diff
@@ -0,0 +1 @@
+# Dummy file to make this directory a package.
```
enen92, line 1: shoudn't be needed anymore, however I understand if this is necessary because of leia and below.

script.myepisodes/utils.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def getSettingAsInt(setting):
+    try:
+        return int(getSetting(setting))
+    except ValueError:
+        return 0
+
+
+def notif(msg, time=5000):
```
enen92, line 31: Use `xbmc.Dialog().notification()` instead of a builtin

enen92 (changes requested):
See comments

script.myepisodes/resources/__init__.py
maximeh: I've removed it, this is for Matrix and up anyway.

script.myepisodes/addon.xml
maximeh: What do you mean by this, I'm not sure what you're referring to there?

maximeh: oh okay the language key, is that right?

enen92: Yes the language element

maximeh:
> See comments

I think it should be all good now. I'll let you review again.

enen92 (approved):

PR #1900, opened 2021-03-04
[service.subtitles.napisy24pl] 3.0.4

CaTzil:
### Description
Fix for extraction of zip files on Windows

### Checklist:

### DIFF
--- a/service.subtitles.napisy24pl/resources/lib/NapisyUtils.py
+++ b/service.subtitles.napisy24pl/resources/lib/NapisyUtils.py
@@ -145,7 +145,7 @@ class NapisyHelper:
         for file in files:
             if os.path.splitext(file)[1] in exts:
                 filename_dest = os.path.join(__temp__, file)
-                flag = xbmcvfs.copy(os.path.join(zip_filepath, file), filename_dest)
+                flag = xbmcvfs.copy(zip_filepath + '/' + file, filename_dest)
                 if flag:
                     subtitle_list.append(filename_dest)
 
--- a/service.subtitles.napisy24pl/addon.xml
+++ b/service.subtitles.napisy24pl/addon.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<addon id="service.subtitles.napisy24pl" name="Napisy24.pl" version="3.0.3" provider-name="CaTz">
+<addon id="service.subtitles.napisy24pl" name="Napisy24.pl" version="3.0.4" provider-name="CaTz">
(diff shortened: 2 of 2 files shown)


service.subtitles.napisy24pl/resources/lib/NapisyUtils.py
```diff
@@ -145,7 +145,7 @@ def download(self, id):
         for file in files:
             if os.path.splitext(file)[1] in exts:
                 filename_dest = os.path.join(__temp__, file)
-                flag = xbmcvfs.copy(os.path.join(zip_filepath, file), filename_dest)
+                flag = xbmcvfs.copy(zip_filepath + '/' + file, filename_dest)
```
enen92, line 148: why does this fix anything?

CaTzil: It was test on Windows machine by a user (I don't have a PC).
Since this line tries to extract a file from zip file (via `zip://` protocol), probably it requires forward slash

enen92: Hum.. maybe you want to have a look at: https://codedocs.xyz/xbmc/xbmc/group__python__xbmcvfs.html#ga7d325903d7bc797f993c9df791d22c44

CaTzil: How this method helps me? 
When coping from zip protocol, the slash needs to be `/`, no reason to "validate" it

enen92: The name is probably not the best choice ever, but it seems to return a validated path not exactly a bool. I'm fine accepting this as is, just seems a hack to me...that's all.

CaTzil: Hehe, bad name!
I will ask the user to check if it works for him, if so, i will update

ciukacz: for some reason zip:// path requires a forward slash after archive name (and before the archive contents).
os.path.join() adds backslash on windows and that causes the extraction to fail

CaTzil: @enen92 can you merge it as is?

enen92, line 148: consider adding a comment here to avoid the same request in the future (free to add on future version)

enen92 (approved):

PR #1904, opened 2021-03-05
[service.subtitles.napiprojekt] 2.0.0

CaTzil:
### Description
Rewrite `service.subtitles.napiprojekt` to support Kodi 19

### Checklist:


service.subtitles.napiprojekt/service.py
```diff
@@ -0,0 +1,199 @@
...
+__scriptname__ = __addon__.getAddonInfo('name')
+__version__ = __addon__.getAddonInfo('version')
+__language__ = __addon__.getLocalizedString
+
+__cwd__ = xbmcvfs.translatePath(__addon__.getAddonInfo('path'))
+__profile__ = xbmcvfs.translatePath(__addon__.getAddonInfo('profile'))
+__resource__ = xbmcvfs.translatePath(os.path.join(__cwd__, 'resources', 'lib'))
+__temp__ = xbmcvfs.translatePath(os.path.join(__profile__, 'temp', ''))
+
+sys.path.append(__resource__)
```
enen92, line 26: Please use relative module import:
from XX import yyy

service.subtitles.napiprojekt/service.py
```diff
@@ -0,0 +1,199 @@
...
+    d = md5()
+    qpath = quote_plus(path, safe='')
+    if rar:
+        path = """rar://""" + qpath + '/'
+        for file in xbmcvfs.listdir(path)[1]:
+            if file.lower().endswith(('.avi', '.mkv', '.mp4')):
+                path = path + file
+                break
+
+    file = xbmcvfs.File(path, "rb")
```
enen92, line 61: with statement?

with xbmcvfs.... as f:
    f.do_something,()

enen92 (changes requested):
Sorry for the delay providing the review

service.subtitles.napiprojekt/service.py
CaTzil: Why? it is a matter of tasted, I prefer the code to be expressive (without trying to guess what is this local var `f`).

enen92: The point is the with statement that will ensure the handler is always closed, not the variable name. Although using file as a variable is also a bad practice since file is a python builtin function.

CaTzil: Ha, thanx for pointing it out, I missed the `with` statement :]

CaTzil:
@enen92, Thank you for your code review, I've refactored the code as requested.

service.subtitles.napiprojekt/service.py
```diff
@@ -0,0 +1,199 @@
...
+__scriptname__ = __addon__.getAddonInfo('name')
+__version__ = __addon__.getAddonInfo('version')
+__language__ = __addon__.getLocalizedString
+
+__cwd__ = xbmcvfs.translatePath(__addon__.getAddonInfo('path'))
+__profile__ = xbmcvfs.translatePath(__addon__.getAddonInfo('profile'))
+__resource__ = xbmcvfs.translatePath(path.join(__cwd__, 'resources', 'lib'))
+__temp__ = xbmcvfs.translatePath(path.join(__profile__, 'temp', ''))
+
+sys_path.append(__resource__)
```
enen92, line 26: This is still here...

enen92 (changes requested):

service.subtitles.napiprojekt/service.py
CaTzil: what is the issue with it?

enen92: It's an ugly hack and a bad practice, probably something you got from some legacy addon. Other devs get inspiration from the addons in repo so I prefer not to have addons doing it in the repo.
Plus, for some addon types kodi manipulates the sys.path itself (e.g. in context menu type addons, a listitem object is added to the sys.path), so addons should avoid messing with it.

Use:

```python
from resources.lib import NapiProjekt

NapiProjekt.whatever_method()
```

CaTzil: I've tried it, it doesn't works :|

enen92 (approved):

PR #1924, opened 2021-03-16
[script.motr] 19.0.3

large:
### Description
Kodi client addon for connecting to a "Movie On The Run" server, for short: MOTR.

Server can be downloaded here: https://github.com/large/MOTRd/releases
Server is based on .Net and can only be installed in Windows, clients are platform independent.
Server is a "wrapper" around handbreak where clients can queue up conversion of your movies.
Server also serve your files for download or streaming for the clients.

Kodi client uses websocket to communicate with the server, websocket ws4py is included in the package.

Test-server for MOTR can be reached here:
Server: test.motr.pw
Port: 880
PortAlternative: 4444 (with SSL)
Username: test
Password: welcome

If you somehow don't get connection, please do a browser check
http://test.motr.pw:880/
https://test.motr.pw:4444/
(SSL is only self signed, I don't have a valid cert, short on $$$)

I versioned the script to match which kodi release it is for, since it is Matrix only (python 3 only).

Please note that this is my first python script ever, so be kind :)
Also I don't do much PR in Github, so if I made a mistake, sorry in advance!

### Checklist:

Additional information :


script.motr/addon.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<addon id="script.motr" name="Movies On The Run" version="19.0.3" provider-name="Lars Werner">
+    <requires>
+        <import addon="xbmc.python" version="3.0.0"/>
+    </requires>
+    <extension point="xbmc.python.script" library="main.py">
+        <provides>video</provides>
+    </extension>
+    <extension point="xbmc.addon.metadata">
+        <summary lang="en_GB">Updated for Kodi Matrix!
```
enen92, line 10: you have summary and description swapped :)

script.motr/resources/lib/kodiutils.py
```diff
@@ -0,0 +1,69 @@
...
+def show_settings():
+    ADDON.openSettings()
+
+def get_setting(setting):
+    return ADDON.getSetting(setting).strip()
+
+def set_setting(setting, value):
+    ADDON.setSetting(setting, str(value))
+
+def get_setting_as_bool(setting):
```
enen92, line 52: all these methods don't make any sense anymore. We have getSettingBool, getSettingInt, etc on the python api:
https://codedocs.xyz/xbmc/xbmc/group__python__xbmcaddon.html#gaf0205dd5b96354f30a39fbf35599e681

script.motr/resources/lib/ws4py/LICENSE
```diff
@@ -0,0 +1,29 @@
+NOTICE: The version of ws4py included with kaithem has been modified.
```
enen92, line 1: Sorry this is not allowed, please package ws4py dependency as a script.module and submit as part of another PR.

enen92 (changes requested):

script.motr/resources/lib/ws4py/LICENSE
large: OK, that was new to me :)
So I'll wrap this as an own module and create a second PR for that then?

Second how do I handle a module import in Kodi, is it the same?
```
from .ws4py import *
from .ws4py.client.threadedclient import WebSocketClient
```

enen92: Exactly just the same, just make sure you add it as a dependency in addon.xml

script.motr/addon.xml
large: :) Didn't noticed, fixed now, thanks.

script.motr/resources/lib/kodiutils.py
large: Depending on how you see it. By doing so you have single point for set / get settings.
Makes it much simpler if you want to add some kind of logic later on. Like the get_setting always strip the value it gets.

In my taste for example get_string(id) function is better than each time calling xbmcaddon.Addon().getLocalizedString(id)
And it basically the same thing been called, since ADDON = xbmcaddon.Addon(), then get_string(id) calls ADDON.getLocalizedString(string_id).

I guess this isn't directly wrong so I would like to have it like that, if it is OK for you?

enen92: This was just a recommendation, not mandatory to change :)

large:
Well my git skill sucks, I tried to squash it into one PR but failed hard.
Any advice for correct cleanup?

enen92:
Yeah it'll merge when the repo syncs. Please remove the ws4py code from this add-on

script.motr/resources/lib/script.py
```diff
@@ -0,0 +1,732 @@
...
+import os, sys, re, socket, urllib, unicodedata, threading, time, traceback, ssl
+from urllib.request import urlopen
+from resources.lib import kodiutils
+
+import logging
+import xbmc, xbmcgui, xbmcaddon, xbmcvfs
+import datetime
+import gc
+
+try:
```
enen92, line 15: This can be removed, use only Json. Simplejson was only needed in helix :)

script.motr/resources/lib/script.py
```diff
@@ -0,0 +1,732 @@
...
+__pluginname__      = 'script.motr'
+__addon__           = xbmcaddon.Addon(id=__pluginname__)
+__addonid__         = __addon__.getAddonInfo('id')
+__addonname__       = __addon__.getAddonInfo('name')
+__cwd__             = __addon__.getAddonInfo('path')
+__author__          = __addon__.getAddonInfo('author')
+__version__         = __addon__.getAddonInfo('version')
+__language__        = __addon__.getLocalizedString
+__internaldebug__   = False
+__loggerlevel__     = xbmc.LOGINFO
```
enen92, line 32: This has to be changed to logdebug

script.motr/resources/skins/Default/720p/convertdialog.xml
```diff
@@ -0,0 +1,269 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
enen92, line 1: Can you rename the windows to script-motr-xxxx. Reason is to make sure there is no clash with skin dialogs

script.motr/resources/skins/Default/720p/motrwindow.xml
```diff
@@ -0,0 +1,707 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
enen92, line 1: Same

enen92 (changes requested):
After these changes and repo sync (~2hours) should be good to go. Thanks for adjusting.

large:
Thank you for the feedback, I have updated as discussed.
Have a nice evening :)

enen92 (approved):

PR #2042, opened 2021-06-12
[script.service.hue] 1.0.0

zim514:
### Description

This add-on adds Phillips Hue light support to Kodi, and to configure pre-set light scenes to trigger during playback actions such as play, pause and stop. It also includes support for Ambilight, but this is not supported by some hardware rendering options. (Android, ARM HEVC/4K)

This add-on has been stable for several months, and available in the forums for a couple of years, so should be pretty solid. 

The forum thread is here: [https://forum.kodi.tv/showthread.php?tid=344886](https://forum.kodi.tv/showthread.php?tid=344886)

### Checklist:

Additional information :


script.service.hue/resources/lib/kodiHue.py
```diff
@@ -0,0 +1,498 @@
...
+    ADDON.setSettingString("bridgeIP", "")
+    ADDON.setSettingString("bridgeUser", "")
+    settings_storage['connected'] = False
+
+    progressBar = xbmcgui.DialogProgress()
+    progressBar.create(_('Searching for bridge...'))
+    progressBar.update(5, _("Discovery started"))
+
+    complete = False
+    while not progressBar.iscanceled() and not complete:
```
ronie, line 119: please include a `not monitor.abortRequested()` check

script.service.hue/resources/lib/kodilogging.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def emit(self, record):
+        levels = {
+            logging.CRITICAL: xbmc.LOGFATAL,
+            logging.ERROR: xbmc.LOGERROR,
+            logging.WARNING: xbmc.LOGWARNING,
+            logging.INFO: xbmc.LOGINFO,
+            logging.DEBUG: xbmc.LOGDEBUG,
+            logging.NOTSET: xbmc.LOGNONE,
+        }
```
ronie, line 30: please only use the xbmc.LOGDEBUG level.
we really would like to keep the regular kodi.log file as clean (and small) as possible.

script.service.hue/resources/lib/reporting.py
```diff
@@ -0,0 +1,43 @@
...
+    data = {
+        'machine': platform.machine(),
+        'platform': platform.system(),
+        'kodi': KODIVERSION,
+        #'kodi': _kodi_version(),
+        'url': url,
+    }
+    rollbar.init(ROLLBAR_API_KEY, capture_ip="anonymize", code_version=ADDONVERSION, root=ADDONPATH, scrub_fields='bridgeUser')
+    rollbar.report_exc_info(sys.exc_info(), extra_data=data)
+
```
ronie, line 39: this really needs to be opt-in!

i would suggest to add a setting which a user can enable if they wish to submit error data.

ronie (changes requested):

script.service.hue/resources/lib/kodiHue.py
zim514: Good catch, will be corrected

script.service.hue/resources/lib/reporting.py
zim514: When there's an unhandled exception, an error message with the exception summary offers both 'cancel' and 'send report'. I preferred to ask every time rather than have an option people might forget. The Rollbar API is only initialized after they choose to send the report. 

The logic for this is in reporting.py, the dialogue box is on line 21 and check is on line 42.

script.service.hue/resources/lib/kodilogging.py
zim514: I will change it throughout. I wasn't aware of this guideline, but just saw it on the wiki. Please be aware that [this post](https://forum.kodi.tv/showthread.php?tid=353818) implies we can use other log levels, it may be helpful to update it with this information.

zim514:
Thanks for taking the time to review my code. I'll address these changes and update the PR after some testing.

script.service.hue/resources/lib/reporting.py
ronie: cool, thanx for the explanation. that way is certainly acceptable.

script.service.hue/resources/lib/kodilogging.py
ronie: cheers, i've added a note to the post as you've suggested.

ronie (approved):
looks good to me, thx!

PR #2043, opened 2021-06-13
[script.xbmc.unpausejumpback@matrix] 3.0.2

bossanova808:
### Add-on details:

- **General**
  - Add-on name: Unpause Jumpback
  - Add-on ID: script.xbmc.unpausejumpback
  - Version number: 3.0.2

- **Code location**
  - URL: https://github.com/bossanova808/script.xbmc.unpausejumpback/
  

        This addon will jump back a number of seconds whenever a video is un-paused - to make sure you don't miss anything.
        You can set the amount of seconds to jump back, and the minimum duration of the pause before the addon will trigger a jump back.
        It also allows to jump back or forward after rewind or fast-forward of a specified amount of seconds to compensate for over-shooting.
        (Originally created by Memphiz, up-keep taken over by bossanova808 in 2020 for Kodi Matrix and beyond).
    

### Description of changes:

v3.0.2
        [fix] "Kodi is not playing any media file" errors

Basically if a user skips past the end of the file, trap the error and cancel the unpause alarm.
    

### Checklist:

### DIFF
--- a/script.xbmc.unpausejumpback/resources/lib/common.py
+++ b/script.xbmc.unpausejumpback/resources/lib/common.py
@@ -1,17 +1,18 @@
 # -*- coding: utf-8 -*-
-
-# Handy utility functions for Kodi Addons
-# By bossanova808
-# Free in all senses....
-# VERSION 0.1.4 2020-10-18
-# (For Kodi Matrix & later)
+"""
+Handy utility functions for Kodi Addons
+By bossanova808
+Free in all senses....
+VERSION 0.2.1 2021-06-06
+(For Kodi Matrix & later)
+"""
+import sys
+import traceback
 
 import xbmc
 import xbmcvfs
 import xbmcgui
 import xbmcaddon
-import sys
-import traceback
 
 ADDON = xbmcaddon.Addon()
 ADDON_NAME = ADDON.getAddonInfo('name')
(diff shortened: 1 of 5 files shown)


script.xbmc.unpausejumpback/resources/lib/common.py
```diff
@@ -1,117 +1,199 @@
...
-    :param name: Required.  Name of the property.
-    :param value: Optional (defaults to "").  Set the property to this value.  An empty string clears the property.
-    """
-    window.setProperty(name, value)
-    # if value and value != False and value != 'na.png':
-    #     log(f'Set window property: [{name}] - value: [{value}]')
+        :param message: required, the message to log
+        :param exception_instance: optional, an instance of some Exception
+        """
+        log(message, exception_instance, level=xbmc.LOGINFO)
```
enen92, line 97: default log level should be debug

enen92 (changes requested):
One little change required before going in

script.xbmc.unpausejumpback/resources/lib/common.py
bossanova808: > default log level should be debug

...it is.  This is a line in a function specific to log at the info level (rarely used) - the default log function just above is the one used almost always by my addons, and it default to only log at the debug levl.  This is just for circumstances where its' _really_ useful to see something being logged. 

Is that ok?   We've actually had this chat before and it was ok then, but I guess I could move _everything_ to debug log, but I'd prefer not to.

enen92: Yeah it is fine. The root cause of myself bringing this all the time is the way the function is written it just implies info is the default level when you already have the other function where you can easily override the log level when needed. I advise you to drop this method in the future and call log with level=info in the caller's if possible to avoid bringing the same comment again :)

enen92 (approved):

bossanova808:
Good point, will do for next time.  Thanks!

Sent on the go... please excuse any typos!

________________________________

PR #2174, opened 2021-12-28
[service.subtitles.legendasdivx] v1.0.5 - Code fix. Login and Regex.

highlandr:
Fix Login
Fix Regex

### Checklist:

- [x ] My code follows the [add-on rules](http://kodi.wiki/view/Add-on_rules) and [piracy stance](http://kodi.wiki/view/Official:Forum_rules#Piracy_Policy) of this project. 
- [ x] I have read the [CONTRIBUTING](https://github.com/xbmc/repo-scripts/blob/master/CONTRIBUTING.md) document
- [ x] Each add-on submission should be a single commit with using the following style: [script.foo.bar] 1.0.0

Additional information :


highlandr:
@enen92 Can you check this out? Been a while since I didn't add new code to this plugin :) at the moment is broken and needs this update. Thanks. Cheers!

service.subtitles.legendasdivx/service.py
```diff
@@ -293,293 +308,288 @@ def append_subtitle(item):
...
\ No newline at end of file
+    def Search(item):
+        """Called when searching for subtitles from KODI."""
+        #### Do what's needed to get the list of subtitles from service site
+        #### use item["some_property"] that was set earlier
+        #### once done, set xbmcgui.ListItem() below and pass it to xbmcplugin.addDirectoryItem()
+        #### CHECKING FOR ANYTHING IN THE USERNAME AND PASSWORD, IF NULL IT STOPS THE SCRIPT WITH A WARNING
+        username = _addon.getSetting( 'LDuser' )
+        password = _addon.getSetting( 'LDpass' )
+        if username == '' or password == '':
```
enen92, line 319: You could make this a bit more pythonic by doing something like:

`if not username or not password:`

service.subtitles.legendasdivx/service.py
```diff
@@ -293,293 +308,288 @@ def append_subtitle(item):
...
+            xbmcplugin.endOfDirectory(int(sys.argv[1]))
+            if username == '' and password != '': xbmc.executebuiltin(('Notification(%s,%s,%d)' % (_scriptname , _language(32016).encode('utf-8'),5000)))
```
enen92, line 321: Can you please avoid the built-in function here? We have access to this on the python API:
xbmcgui.Dialog().notification()

enen92 (changes requested):
All good execpt for the notification usage as bultins. We should avoid that as much as possible if the same functions are available on the python interface.
It is required to avoid the sys.path.append() usage and use relative imports instead:

`from resources.lib import x`

However, as this was not introduced in the scope of this PR you can do it in a later update.

service.subtitles.legendasdivx/service.py
highlandr: Done @enen92

service.subtitles.legendasdivx/service.py
highlandr: Done.

enen92 (approved):

PR #2189, opened 2022-01-12
[script.bluetooth.delay] 1.0.0

MHR-official:
### Description
This addon makes it easy to use Kodi with bluetooth headphones.

With this addon you can set up Audio Delay for one or two devices, and toggle between them via a shortcut on your home screen, or a shortcut on your keyboard.
### Checklist:

New pull request, same files.
(I uploaded the files again, now with a folder).


MHR-official:
`Schema validation failed for the following points: xbmc.servicexbmc.addon.metadata`
What it means?

enen92:
means all these elements are wrong:

```
<license></license>
<language></language>
<forum>http://</forum>
<source>https://</source>
<website></website>
<email></email>
```

- License must be defined and use a spdx identifier (in your case `GPL-2.0-only` or `GPL-2.0-or-later` depending on your interest)
-  Language should be dropped as it doesn't make sense on a program addon (it refers to content, e.g. audio or video)
-  Forum should be a support thread on kodi.tv (or remove the element if none)
-  Website should be valid (or remove the element if none)
-  Email should be valid (or remove the element if none)

However a quick glance at the code reveals major issues. You're touching/creating keymap files and that is not allowed on the repo. As we don't have any API for that, there's no way to do it in a clean and not error prone way. The only addon allowed to modify keymaps is the "Keymap Editor" addon that is provided by the team.

MHR-official:
Any comments on the English translation in strings.po or README.md?
Thanks

MHR-official:
Any comments?
Thanks

script.bluetooth.delay/resources/settings.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="utf-8" standalone="yes"?>
+
+<settings>
```
enen92, line 3: can you use the new style settings? https://kodi.wiki/view/Add-on_settings_conversion

You can do this automatically see first post of https://forum.kodi.tv/showthread.php?tid=350135

script.bluetooth.delay/addon.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+d1 = Addon.getSetting('Device1')
+d2 = Addon.getSetting('Device2')
+script = xbmc.translatePath(os.path.join('special://home/addons/script.bluetooth.delay/AudioDelay.py'))
+
+if d2 == d1:
+	xbmcaddon.Addon().openSettings()
+else:
+	xbmc.executebuiltin("RunScript("+script+")", wait=True)
```
enen92, line 27: I really want to deprecate raw RunScript execution in the near future. Why don't you really only import and execute the code here?

script.bluetooth.delay/Device2.py
```diff
@@ -0,0 +1,168 @@
...
+integer_types = (int)
+
+Addon = xbmcaddon.Addon('script.bluetooth.delay')
+
+line1 = Addon.getSetting('line1')
+line2 = Addon.getSetting('line2')
+t = 1000
+
+if (xbmc.getCondVisibility('Player.HasMedia') == False):
+    xbmc.executebuiltin('Notification(%s, %s, %d)'%("",Addon.getLocalizedString(30015), t))
```
enen92, line 22: this should be xbmcgui.Dialog().notification()
We have support for this on the python api no need for builtins

script.bluetooth.delay/resources/settings.xml
MHR-official: With the new style settings the addon gets stuck (in the debug log nothing appears).
it's necessary?

script.bluetooth.delay/addon.py
MHR-official: After import, how do I execute the code?

MHR-official:
There is an error in the log with sys.exit(). This error occurs  only in kodi 19.
I need to check.

script.bluetooth.delay/AudioDelay.py
```diff
@@ -0,0 +1,189 @@
...
+    xbmc.executebuiltin("Action(select)")
+    xbmc.executebuiltin('SetFocus(11)')
+    xbmc.executebuiltin("Action(select)", wait=True)
+
+    time2 = os.path.getmtime(check)
+    while time2 == time1:
+        time.sleep(0.01)
+        time2 = os.path.getmtime(check)
+
+    sourcesXML = minidom.parse(xbmc.translatePath('special://profile/guisettings.xml'))
```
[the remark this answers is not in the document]
MHR-official: The setting in the guisettings.xml file is as follows:
`<audiodelay> 0.000000 </audiodelay>`
Without setting id.
See here [https://forum.kodi.tv/showthread.php?tid=353852&pid=2998739#pid2998739](https://forum.kodi.tv/showthread.php?tid=353852&pid=2998739#pid2998739).
I do not change the xml file, just read it.

MHR-official:
@enen92 
Saving audio delay value for Bluetooth is a basic thing for media player software.
The problem is that even via advancedsettings.xml it is not possible to do this.
See here https://forum.kodi.tv/showthread.php?tid=366506
If you still think it is not possible with minidom.parse, you can close the PR if you want. I will post the addon in the forum instead.
Thanks anyway for your reviews.

PR #2231, opened 2022-02-21
[script.bluetooth.delay] 1.1.0

MHR-official:
### Description
Update from v1.0.0 to v1.1.0

- Now you can set a delay even when the video is playing
- The audio delay setting does not reset the subtitle delay value
### Checklist:


MHR-official:
please update. The previous version may reset the delay value of the subtitles.
small question:
This addon (lite) and the [regular addon](https://github.com/MHR-official/repo/tree/master/zips/script.bluetooth.delay) in my repo, are with the same id. Could there be a conflict when there is an update, and is there a difference between kodi 18 and 19?
Thank you

script.bluetooth.delay/addon.py
```diff
@@ -99,24 +79,32 @@ def speakers():
 def main():
 
     if arg == "reset":
-        xbmc.executebuiltin('PlayerControl(stop)')
-        Addon.setSettingBool('state', 1)
-        reset1()
+        if (xbmc.getCondVisibility('Player.HasVideo') == False):
+            xbmc.executebuiltin("PlayMedia("+path+"reset.mp4,1)")
```
lrusak, line 83: I'm not sure if we allow including media like this.

script.bluetooth.delay/addon.py
```diff
@@ -99,24 +79,32 @@ def speakers():
...
 
-    elif (xbmc.getCondVisibility('Player.HasMedia') == False):
-        Addon.setSettingBool('state', 1)
-        reset1()
-        skin1()
-        xbmcgui.Dialog().notification("",line1, "",t)
-        xbmcgui.Dialog().notification("",Addon.getLocalizedString(30009), "",t)
+    elif (xbmc.getCondVisibility('Player.HasVideo') == False):
+        if state == "true":
+            xbmcgui.Dialog().notification("",line1 + " " + Addon.getLocalizedString(30009), "",t*3)
```
lrusak, line 105: You should use proper python formatting. https://docs.python.org/3/tutorial/inputoutput.html#the-string-format-method

script.bluetooth.delay/addon.py
```diff
@@ -99,24 +79,32 @@ def speakers():
...
 
     if arg == "reset":
-        xbmc.executebuiltin('PlayerControl(stop)')
-        Addon.setSettingBool('state', 1)
-        reset1()
+        if (xbmc.getCondVisibility('Player.HasVideo') == False):
+            xbmc.executebuiltin("PlayMedia("+path+"reset.mp4,1)")
+        xbmc.executebuiltin('ActivateWindow(busydialognocancel)')
+        time.sleep(2)
+        for x in range (800):
```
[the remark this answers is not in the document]
MHR-official: To change the delay value of a video - you must play a video.
Otherwise you do not have access to osdaudiosettings.
The idea is to reset the delay whenever a new value is set.

script.bluetooth.delay/addon.py
MHR-official: Please explain to me what exactly to change?

script.bluetooth.delay/addon.py
MHR-official: See [here](https://github.com/xbmc/repo-scripts/pull/2231#discussion_r813231704). This is necessary for resetting the delay value.

script.bluetooth.delay/addon.py
```diff
@@ -1,8 +1,13 @@
 import xbmcaddon
 import xbmcgui
+import time
+import sys
 
 Addon = xbmcaddon.Addon('script.bluetooth.delay')
 
+path = Addon.getAddonInfo('path')
+if sys.version < '3':
```
enen92, line 9: Unneeded, matrix is guaranteed to use python3

script.bluetooth.delay/addon.py
```diff
@@ -99,24 +79,32 @@ def speakers():
 def main():
 
     if arg == "reset":
-        xbmc.executebuiltin('PlayerControl(stop)')
-        Addon.setSettingBool('state', 1)
-        reset1()
+        if (xbmc.getCondVisibility('Player.HasVideo') == False):
```
enen92, line 82: if not X:

script.bluetooth.delay/addon.py
```diff
@@ -99,24 +79,32 @@ def speakers():
...
+            xbmc.executebuiltin("PlayMedia("+path+"reset.mp4,1)")
+        xbmc.executebuiltin('ActivateWindow(busydialognocancel)')
+        time.sleep(2)
```
enen92, line 85: Use xbmc.sleep() for small values, use XBMC.Monitor().waitForAbort() for longer values (usually > 4secs)

script.bluetooth.delay/addon.py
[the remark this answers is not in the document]
enen92: This really sounds like core limitations that should be fixed/attempted there

script.bluetooth.delay/addon.py
```diff
@@ -99,24 +79,32 @@ def speakers():
...
+        for x in range (800):
+            xbmc.executebuiltin("Action(AudioDelayPlus)")
+        for x in range (400):
+            xbmc.executebuiltin("Action(AudioDelayMinus)")
+        time.sleep(1)
```
enen92, line 90: No time usage, use Kodi API methods

script.bluetooth.delay/addon.py
```diff
@@ -99,24 +79,32 @@ def speakers():
...
         skin1()
-        xbmcgui.Dialog().notification("",Addon.getLocalizedString(30006), "",t*2)
+        time.sleep(1)
```
enen92, line 92: Same

script.bluetooth.delay/addon.py
```diff
@@ -99,24 +79,32 @@ def speakers():
...
+        xbmc.executebuiltin('Dialog.Close(busydialognocancel)')
+        Addon.setSettingBool('state', 1)
```
enen92, line 94: Why are you storing state in settings? Isn't it scoped to the running script? State should be a class variable or similar

script.bluetooth.delay/addon.py
enen92: Google python3 fstrings for example, or str format()

script.bluetooth.delay/addon.py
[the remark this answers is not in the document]
enen92: But yeap, shipping videos (or any binary file) is not allowed sorry

enen92 (changes requested):

script.bluetooth.delay/addon.py
MHR-official: Hello enen92
Need the state in the settings, so that it is saved even after restarting kodi.

MHR-official:
@enen92 
please update.
It is now without a binary file.
Thanks

MHR-official:
Hello
Who should I contact to update the addon?
The previous version may reset the delay value of the subtitles, and now it's okay.

PR #2261, opened 2022-04-21
[script.sublissimo] 2.0.0

Allertj:
- Major refactor of the existing code. Lots of clean up, splitting into separate files, more readable etc. (All functionality stays the same)
- Added a few fine tuning options.

### Checklist:

- [X] My code follows the [add-on rules](http://kodi.wiki/view/Add-on_rules) and [piracy stance](http://kodi.wiki/view/Official:Forum_rules#Piracy_Policy) of this project. 
- [X] I have read the [CONTRIBUTING](https://github.com/xbmc/repo-scripts/blob/master/CONTRIBUTING.md) document
- [X] Each add-on submission should be a single commit with using the following style: [script.foo.bar] 1.0.0

Additional information :
- Submitting your add-on to this specific branch makes it available to any Kodi version equal or higher than the branch name with the applicable Kodi dependencies limits.
- [add-on development](http://kodi.wiki/view/Add-on_development) wiki page.
- Kodi [pydocs](http://kodi.wiki/view/PyDocs) provide information about the Python API
- [PEP8](https://www.python.org/dev/peps/pep-0008/) codingstyle which is considered best practise but not mandatory.
- This add-on repository has automated code guideline check which could help you improve your coding. You can find the results of these check at [Codacy](https://www.codacy.com/app/Kodi/repo-scripts/dashboard). You can create your own account as well to continuously monitor your python coding before submitting to repo.
- Development questions can be asked in the [add-on development](http://forum.kodi.tv/forumdisplay.php?fid=26) section on the Kodi forum.

PR #2266, opened 2022-05-01
[script.sublissimo] 2.0.1

Allertj:
- Small bugfix

### Checklist:

Additional information :

### DIFF
--- a/script.sublissimo/resources/lib/menu_items/syncing.py
+++ b/script.sublissimo/resources/lib/menu_items/syncing.py
@@ -44,30 +44,22 @@ def check_player_instances(subtitle=None):
     if subtitle:
         subtitle.delete_temp_file()
 
-def experiment(typeofinstance, subtitle):
-    current_player = xbmc.Player()
-    if current_player.isPlayingVideo():
-        if current_player.getPlayingFile() == subtitle.videofilename:
-            player = PlayerInstance().request(typeofinstance)
-            player.add(subtitle)
-            player.start()
-            xbmc.Monitor().waitForAbort(1)
-    else:
-        return False
-
 def create_player(typeofinstance, subtitle):
-    if not experiment(typeofinstance, subtitle):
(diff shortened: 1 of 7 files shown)


Allertj:
@enen92 The fix in https://github.com/xbmc/repo-scripts/pull/2261 (xbmc.Monitor().waitForAbort(1)) caused some unforeseen problems. Just a small fix for them.

script.sublissimo/resources/lib/menu_items/syncing.py
```diff
@@ -44,30 +44,18 @@ def check_player_instances(subtitle=None):
...
+    location = retrieve_video(subtitle)
+    player.add(subtitle)
+    player.play(location)
+    player.start()
+    # -----Possibly needed if video is unavailable-----
+    # xbmc.sleep(1000)
+    # if player.isPlaying() == False:
+    #     show_dialog(subtitle)
+    # -------------------------------------------------
+    xbmc.Monitor().waitForAbort(1000)
```
enen92, line 58: this is 1000 seconds of pause. Why do you need such a long pause?
Most likely a clear sign something is wrong with your logic... Do you want to keep something alive while the player is active or so?

Allertj: For the sync option the user needs time to select a begin and an endpoint to sync the subtitle with. In the older versions of this script, I always used xbmc.sleep(1000). Most users (well I at least) select these two point within 5 minutes, and then exit the player. Without it, the menu's simply don't appear.

enen92: Do have access to the maximum time interval? Something like this should work:

```
while not monitor.abortRequested():
   if something:
       break
   monitor.waitForAbort(1)
```

Allertj: Well that seems to work! Is this solution okay?

enen92 (approved):
Thanks

PR #2272, opened 2022-05-09
[service.cronxbmc] 0.1.6

robweber:
### Description

This is an addon I've maintained for years but never submitted to the official repo before, hence the reference to xbmc in the addon id. I've removed all other references and renamed to them to Kodi but kept the addon id name for compatibility purposes. 

This addon add a basic cron scheduler for recurring tasks. These can be scheduled via a GUI using built in functions or JSON calls. It can also be imported as a dependency within another addon and used to schedule tasks so other addons don't need to create their own schedulers. Syntax for that can be [found in the documentation](https://github.com/robweber/cronxbmc#import-into-another-addon). 

### Checklist:

Additional information :


service.cronxbmc/addon.xml
```diff
@@ -0,0 +1,29 @@
...
+  <extension point="xbmc.python.pluginsource" library="gui.py">
+     <provides>executable</provides>
+  </extension>
+  <extension point="xbmc.service" library="service.py" />
+  <extension point="xbmc.python.module" library="resources/lib" />
+  <extension point="xbmc.addon.metadata">
+    <summary lang="en_GB">A customizable Cron timer for Kodi. </summary>
+    <description lang="en_GB">Create cron timers for Kodi functions. This addon can schedule and run any built-in or JSON Kodi functions. No more reliance on add-on developers to allow for "timer" functionality, no more scripts running on other machines to send commands to your HTPC. All of your timers can now be handled directly within Kodi. It can also display notifications when your tasks run if configured to do so.</description>
+    <platform>all</platform>
+    <language></language>
```
enen92, line 17: MIT

enen92, line 17: Drop please the empty tag

service.cronxbmc/gui.py
```diff
@@ -0,0 +1,217 @@
...
+import sys
+from urllib.parse import parse_qsl
+from resources.lib.cron import CronManager, CronJob
+from resources.lib.cron_utils import utils
+
+
+class CronGUI:
+    params = {}
+    context_url = "%s?%s"
+    plugin_url = 'Xbmc.RunPlugin(%s?%s)'
```
enen92, line 13: Won't work, XBMC prefix have been dropped. Rename to RunPlugin only please

service.cronxbmc/resources/lib/__init__.py
```diff
@@ -0,0 +1,5 @@
+# For relative imports to work from Python 3.6
+import os
+import sys
+
+sys.path.append(os.path.dirname(os.path.realpath(__file__)))
```
enen92, line 5: Please drop, no need to dummy init files, also please avoid syspath append and just use relative module import

service.cronxbmc/resources/lib/_release.py
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+__author__ = "Matsumoto Taichi"
```
enen92, line 3: Looks it doesn't do nothing

service.cronxbmc/resources/lib/cron.py
```diff
@@ -0,0 +1,282 @@
...
+                utils.log(tempJob.name + " " + tempJob.expression + " loaded")
+                adv_jobs[tempJob.id] = tempJob
+
+        except IOError:
+            # the file doesn't exist, return empty array
+            doc = xml.dom.minidom.Document()
+            rootNode = doc.createElement("cron")
+            doc.appendChild(rootNode)
+            # write the file
+            f = xbmcvfs.File(xbmcvfs.translatePath(self.CRONFILE), "w")
```
enen92, line 185: Would be nice to use pythons with keyword to avoid closing the handler manually

service.cronxbmc/resources/lib/cron.py
```diff
@@ -0,0 +1,282 @@
...
+                newChild.setAttribute("command", aJob.command)
+                newChild.setAttribute("show_notification", aJob.show_notification)
+                newChild.setAttribute("addon", aJob.addon)
+                newChild.setAttribute("last_run", str(aJob.last_run))
+                newChild.setAttribute('run_if_skipped', aJob.run_if_skipped)
+
+                rootNode.appendChild(newChild)
+
+            # write the file
+            f = xbmcvfs.File(xbmcvfs.translatePath(self.CRONFILE), "w")
```
enen92, line 216: Same

service.cronxbmc/resources/lib/cron_utils/__init__.py
```diff
@@ -0,0 +1 @@
+# Dummy file to make this directory a package.
```
enen92, line 1: Not needed in py3

service.cronxbmc/resources/lib/__init__.py
TermeHansen: I added this to make it possible to import cron from another addon. without this I get:

```
                                                   Error Type: <class 'ModuleNotFoundError'>
                                                   Error Contents: No module named 'resources.lib.croniter'
                                                   Traceback (most recent call last):
                                                     File "/home/termo/.kodi/addons/plugin.video.drnu/default.py", line 2, in <module>
                                                       from resources.lib import addon
                                                     File "/home/termo/.kodi/addons/plugin.video.drnu/resources/lib/addon.py", line 28, in <module>
                                                       from cron import CronManager, CronJob
                                                     File "/home/termo/.kodi/addons/service.cronxbmc/resources/lib/cron.py", line 6, in <module>
                                                       from resources.lib.croniter import croniter
                                                   ModuleNotFoundError: No module named 'resources.lib.croniter'

```

service.cronxbmc/gui.py
robweber: Is that like a planned deprecation? It does still work in Kodi Matrix.

TermeHansen:
> I think everything is fixed up with the exception of the one @TermeHansen commented on above. The import into other addons doesn't seem to work without it. Is there a better way to make sure the required files are in the path?

I looked at pandas last night and it should be possible to do with __init__.py file without adding to path. I'll try tonight and let you know.

robweber:
Updated with some fixes for the relative imports. There is still an `__init__.py` file but it facilitates the use of this addon within others as an import.

TermeHansen:
@enen92 I think this is ready now :)

service.cronxbmc/gui.py
```diff
@@ -0,0 +1,216 @@
...
+
+        # get the name, command, expression and notification setting
+        name = xbmcgui.Dialog().input(heading=utils.getString(30002))
+
+        if(name == ""):
+            return
+        else:
+            newJob.name = name
+
+        type = xbmcgui.Dialog().select(utils.getString(30067), ["Built-in Function", "JSON Command"], preselect=0)
```
enen92, line 31: Any items shown on the UI need to be localized/translatable

service.cronxbmc/gui.py
```diff
@@ -0,0 +1,216 @@
...
+        else:
+            newJob.show_notification = "false"
+
+        if(xbmcgui.Dialog().yesno(utils.getString(30005), utils.getString(30066))):
+            newJob.run_if_skipped = "true"
+        else:
+            newJob.run_if_skipped = "false"
+
+        if(not self.cron.addJob(newJob)):
+            xbmcgui.Dialog().ok(utils.getString(30000), 'Job not added, cron expression error')
```
enen92, line 63: Same

service.cronxbmc/gui.py
```diff
@@ -0,0 +1,216 @@
...
+            self.cron.addJob(aJob)
+
+        elif(command == 5):
+            # update the expression
+            aJob = self.cron.getJob(int(self.params['job']))
+
+            aJob.expression = xbmcgui.Dialog().input(utils.getString(30006) + " " + utils.getString(30004), aJob.expression)
+
+            if(not self.cron.addJob(aJob)):
+                xbmcgui.Dialog().ok(utils.getString(30000), 'Job not added, cron expression error')
```
enen92, line 100: same

enen92 (changes requested):
Just some minors, after those fixes it can go in

robweber:
Just wanted to bump this as the changes requested are all done.

pkscout (approved):

PR #2280, opened 2022-05-22
[script.bluetooth.delay] 1.1.0

MAR-official:
### Description
Update from v1.0.0 to v1.1.0

- Set the delay value when the video is playing
- The audio delay setting does not reset the subtitle delay value
### Checklist:


razzeee:
Please change your commit message to also be `[script.bluetooth.delay] 1.1.0`

script.bluetooth.delay/addon.py
```diff
@@ -34,89 +39,70 @@ def skin2():
...
-        xbmc.executebuiltin('PlayerControl(stop)')
-        Addon.setSettingBool('state', 1)
-        reset1()
-        skin1()
-        xbmcgui.Dialog().notification("",Addon.getLocalizedString(30006), "",t*2)
+        if not xbmc.getCondVisibility('Player.HasVideo'):
+            xbmcgui.Dialog().notification("",Addon.getLocalizedString(30006), "",t)
+        else:
+            if firstRun == "false":
+                if xbmcgui.Dialog().ok(Addon.getLocalizedString(30007), Addon.getLocalizedString(30008)) == True:
```
romanvm, line 85: `is True`

script.bluetooth.delay/addon.xml
```diff
@@ -1,18 +1,19 @@
...
 	<requires>
 		<import addon="xbmc.python" version="3.0.0"/>
-	</requires>	
+	</requires>
 	<extension point="xbmc.python.script" library="addon.py">
 		<provides>executable</provides>
 	</extension>
 	<extension point="xbmc.addon.metadata">
 		<platform>all</platform>
 		<summary lang="en">Toggle audio delay</summary>
```
romanvm, line 11: Please use 4-letter language codes: `en_GB`

romanvm (changes requested):
Please correct language codes in `addon.xml`. Other comments are not mandatory.

script.bluetooth.delay/addon.py
MAR-official: Do you mean `if xbmcgui.Dialog().ok(Addon.getLocalizedString(30007), Addon.getLocalizedString(30008)) is True:` ?
Because it does not work properly.

romanvm (approved):

PR #2331, opened 2022-10-21
[script.skipit] 1.0.3

shartrec:
### Description
Added a service to add a keymap file to bind "Right Arrow" to the plugin when in full screen video mode.
### Checklist:

- [ x] My code follows the [add-on rules](http://kodi.wiki/view/Add-on_rules) and [piracy stance](http://kodi.wiki/view/Official:Forum_rules#Piracy_Policy) of this project. 
- [ x] I have read the [CONTRIBUTING](https://github.com/xbmc/repo-scripts/blob/master/CONTRIBUTING.md) document
- [ x] Each add-on submission should be a single commit with using the following style: [script.foo.bar] 1.0.0

Additional information :

### DIFF
--- a/script.skipit/resources/lib/buildkeymap.py
+++ b/script.skipit/resources/lib/buildkeymap.py
@@ -1,0 +1,35 @@
+import os
+import xbmcvfs
+import xbmc
+
+
+def setup_keymap_folder():
+    if not os.path.exists(userdata):
+        os.makedirs(userdata)
+
+def write_keymap(file):
+    
+    setup_keymap_folder()
+    # if the skipit keymapfile already exists do nothing
+    if os.path.exists(file):
+        exit
+
+    keymapxml = open(file,"w")
+    keymapxml.write("<keymap><fullscreenvideo><keyboard>")
+    keymapxml.write("<key id=\"61571\">runaddon(script.skipit)</key>")
+    keymapxml.write("</keyboard></fullscreenvideo></keymap>")
+    
+    keymapxml.close()
+    
+monitor = xbmc.Monitor()
+
+if monitor.abortRequested():
+    exit
+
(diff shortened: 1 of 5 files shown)


script.skipit/resources/lib/buildkeymap.py
```diff
@@ -0,0 +1,35 @@
...
+        os.makedirs(userdata)
+
+def write_keymap(file):
+    
+    setup_keymap_folder()
+    # if the skipit keymapfile already exists do nothing
+    if os.path.exists(file):
+        exit
+
+    keymapxml = open(file,"w")
```
basrieter, line 17: It's not allowed to edit/create files outside of your add-on's user_data folder (see https://kodi.wiki/view/Add-on_rules:

> Add-ons should store all their data in their own subfolder inside the addon_data directory. Access (read/write/delete) to any other files / folders is not allowed by default.

So unfortunately you cannot create keymaps like this.

basrieter (changes requested):

script.skipit/resources/lib/buildkeymap.py
shartrec: OK, and I understand the reason for such a rule, but it does beg the question, why have a userdata/keymaps directory if not for adding just this sort of thing.

Anyway, do you know of any other way to provide this sort of functionality, i.e. add a key binding for some addon functionality.

Currently this requires to user to use the keymap editor plugin (which does write to this directory) to map a key, which I feel is a lot less user friendly than it might be.

basrieter: They could upload a keymap.xml themselves. But it is not allowed to be done from an add-on.

shartrec:
Because of restriction on creating files highlighted by @basrieter I will close this PR

PR #2438, opened 2023-04-15
[script.module.chompjs@krypton] 1.0.0

joaopa00:
### Description
Add chomps as a new module for Kodi. Ease the conversion of javascript object into python object.
### Checklist:

Additional information :
Chompjs will be requested by the next version of Catchuptvandmore plugin.


basrieter:
We are no longer accepting new Python 2 based add-ons. Please submit a Python 3 version for your add-on to either the Matrix, Nexus or Omega branch. I would suggest the Matrix one, as it will be available in all three.

joaopa00:
Thanks for the feedback. I opened a new PR. A note: many users of Kodi use it on old low system  (rasperri 2 or 3) that support python 2 only. It would be nice these users can profit from improvements of plugins that need python 2 only.

PR #2478, opened 2023-06-30
[script.timers] 3.7.0

Heckie75:
### Description
v3.7.0 (2023-06-30)
- If you stop explicitly playback while a start-stop-timer is running there won't be another stop action anymore when this timer runs out.
- Added workaround that streamed video (probably mpeg-dash) immediately stops after timer has started (only happened if 'seek to correct time if timer starts belatedly' is activated)

### Checklist:


romanvm (approved):
LGTM

PR #2585, opened 2024-01-26
[service.iptv.manager] v0.2.5

pkscout:
fix for bad commit from someone who wasn't author of addon


pkscout:
I'm merging this even though I submitted it because I'm the one that merged the previous thing (that had been reviewed and approved).  So this is to fix that mistake.

Issue #2225, opened 2022-02-20
MultiWeather: Weather Error since 0.0.9 update


darkenvy:
I get a generic error every time I open Kodi. After checking the logs, I get the following output from MultiWeather:

```
EXCEPTION Thrown (PythonToCppException) : -->Python callback/script returned the following error<--
                                                    - NOTE: IGNORING THIS CAN LEAD TO MEMORY LEAKS!
                                                   Error Type: <class 'KeyError'>
                                                   Error Contents: 0
                                                   Traceback (most recent call last):
                                                     File "/storage/emulated/0/Android/data/org.xbmc.kodi/files/.kodi/addons/weather.multi/default.py", line 5, in <module>
                                                       weather.MAIN(mode=sys.argv[1])
                                                     File "/storage/emulated/0/Android/data/org.xbmc.kodi/files/.kodi/addons/weather.multi/lib/weather.py", line 35, in __init__
                                                       self.get_forecast(location, locationid, locationlat, locationlon, ycookie, ycrumb)
                                                     File "/storage/emulated/0/Android/data/org.xbmc.kodi/files/.kodi/addons/weather.multi/lib/weather.py", line 157, in get_forecast
                                                       yahoo.Weather.get_weather(data, loc, locid)
                                                     File "/storage/emulated/0/Android/data/org.xbmc.kodi/files/.kodi/addons/weather.multi/lib/providers/yahoo.py", line 8, in get_weather
                                                       data = response['weathers'][0]
                                                   KeyError: 0
                                                   -->End of Python script error report<--
```

lrusak:
I think it's known, https://forum.kodi.tv/showthread.php?tid=352028&pid=3086785#pid3086785

PR #2626, opened 2024-06-16
[weather.multi] 0.0.27

ronie:
### Description
the addon no longer works on new installs due to changes at Yahoo.
while it continues to work on existing installs (for now...), 
mark it as broken to prevent users from installing this addon.

### Checklist:

Additional information :


romanvm (approved):

romanvm (approved):

Martii:
@ronie,

Would you please explain the "new user" bit here? 

I've already had this addon installed for a long time and got the notice that it was being disabled even though I'm an anonymous user i.e. I don't have an API key... I  usually get the temperature in my US region. After it was abruptly "recalled" by disabling I reenable it and it works just fine, anonymously. Even in the Weather "tab" it is still showing my extended forecast.

Thank you for taking some time to clarify why it was necessary to disable anonymous users of Multi Weather from being able to see their forecasts.

ronie:
the addon will continue to work on existing installs,
but it will fail if you install it now.

Martii:
@ronie 

> but it will fail if you install it now.

Are you claiming I can add it into an existing Kodi (fresh or otherwise) then it will fail?

Martii:
@ronie

Confirmed that the question *(Kodi is unable to add it in)* is self-answered **because this commit removed it**.

Again I reask "Thank you for taking some time to clarify why it was necessary to disable anonymous users of Multi Weather from being able to see their forecasts.". 

This is not applicable as an anonymous user without an API key. i.e. the Weather App has been disabled by Kodi itself even though the usual primary usage of it still works.

Ref(s) found:
* https://github.com/xbmc/repo-scripts/issues/2225
* https://forum.kodi.tv/showthread.php?tid=352028&pid=3086785#pid3086785

ronie:
@Martii please use the forum for discussions, github isn't the place for it.
(it sends out notifications to too many kodi devs with every message posted here).

Martii:
n/m

> ronie correctly marked the add-on as broken with useful information that states that only users with old+valid cookie data will still be able to fetch weather data.
Not doing this would not be a better solution because other users might ask why the add-on is not marked as broken when it clearly is in some way.

Ca. https://forum.kodi.tv/showthread.php?tid=352028&pid=3203275#pid3203275 

i.e. cookie based requests are broken so the whole thing is marked as broken. Kodi add-on structure should really refine this down for those who use a portion of this who only utilize anonymous.


Too bad you couldn't have put this in here in the PR. Reputation withstanding affected.

Martii:
@ronie 

> (it sends out notifications to too many kodi devs with every message posted here).

I'm aware of that. As I've already had a bad experience on the forum I choose here instead where the problem lies. GH is my preferred way of communicating with the devs as there is less b.s.

Thank you for the suggestion though.

PR #2758, opened 2025-08-30
[script.tag-generator] 0.9.1

wellspokenman:
### Description
Kodi 21 Support, Fixed IMDb parsing of Oscars, Fixed Wikipedia scraping for stand-up comedians, Fixed Trakt authentication

### Checklist:

Additional information :


script.tag-generator/resources/lib/trakt/auth/__init__.py
```diff
@@ -0,0 +1,84 @@
...
+    return DeviceAuthAdapter(client=api(), config=config).authenticate()
+
+
+def get_client_info(app_id: bool, config: AuthConfig):
+    """Helper function to poll the user for Client ID and Client Secret
+    strings
+
+    :return: A 2-tuple of client_id, client_secret
+    """
+    print('If you do not have a client ID and secret. Please visit the '
```
romanvm, line 35: Please do not use direct `print()` function. Use `xbmc.log()` instead.

script.tag-generator/main.py
romanvm, line 1: I strongly recommend moving all the code into an importable module leaving only a small entrypoint code here. It will speed-up addon start on lower-end systems like Raspberry Pi or various Android TV boxes.

romanvm (changes requested):

romanvm:
Please check my comments.

script.tag-generator/resources/lib/trakt/auth/__init__.py
wellspokenman: Any error relevant for the user is communicated via xbmc.log, these print statements are from the trakt library that I require for the oauth flow. I can comment them out if you like?

script.tag-generator/main.py
wellspokenman: i had a go, but this is a major rewrite. All the variables will need to be moved then I have to test it all again.

wellspokenman:
> Please check my comments.

I've looked into moving the code as you suggest, but this is more work than I have time for right now. This is an old addon that ran ok on rasperry pi a decade ago - if it was good enough for Kodi Frodo I don't understand why it isn't sufficiently performant for Omega. The trakt library is only ever accessed within a try catch that has Kodi logs and gui errors.

romanvm:
I think it's good as it is now. Regarding the addon entrypoint it is just a recommendation. Regarding print() we usually do not allow it but if the code is a copy-paste from a generic Python library I think exceptions can be made. In Kodi `sys.stdout` is hardwired to `xbmc.log` anyway.

romanvm (approved):

PR #2813, opened 2026-03-24
[script.easymovie] 1.0.1

Rouzax:
## Update: EasyMovie v1.0.1

**Addon ID:** script.easymovie
**Version:** 1.0.1
**Previous version:** 1.0.0

### Changes since 1.0.0

- Fixed Resume from the "Resume Movie?" dialog. The previous implementation
  called `Player.Open` and issued `xbmc.Player().seekTime()` 500 ms later;
  the player was rarely ready by then, so the seek was silently dropped and
  playback started from zero. The fix now passes Kodi's native single-item
  `Player.Open` options.resume with a `Player.Position.Time` object, so the
  start offset is set before the player initialises. A 10-second rewind is
  applied for context, matching EasyTV's behavior.

### Internal

- `build_play_movie_query()` in `data/queries.py` accepts an optional
  `resume_seconds` argument and emits `options.resume = {hours, minutes,
  seconds}` when set.
- Added `RESUME_REWIND_SECONDS` constant (10 seconds).
- Added regression tests covering the resume timecode split, including the
  exact position from the original report (3266s = 0h54m26s).
- Verified live against Kodi 21: requested timecode is honoured, no
  post-play seek required.

### Technical details

- Python 3.8+ compatible (Kodi 21 Omega and newer)
- Depends on xbmc.python 3.0.1
- Passes pyright, pyflakes, and kodi-addon-checker with no errors
- Sibling addon to [script.easytv](https://github.com/Rouzax/script.easytv)

### Links

- **Forum thread:** https://forum.kodi.tv/showthread.php?tid=383903
- **Source code:** https://github.com/Rouzax/script.easymovie
- **Full changelog:** https://github.com/Rouzax/script.easymovie/blob/main/changelog.txt

---

*Note: this PR's title and body were briefly overwritten by an unrelated EasyTV submission and have been restored. The branch HEAD is back to the EasyMovie 1.0.1 commit `50ff8fb`.*


script.easymovie/resources/clone.py
romanvm, line 1: After discussing this with other team members, I would ask you to remove this functionality. Addons should be installed only via proper channels, that is, either from an online repo or from a .zip file using the respective Kodi UI functionality. "Drop in" installs are not allowed and not supported.

romanvm (changes requested):

script.easymovie/resources/clone.py
Rouzax: I fully understand the security concerns of addons cloning themselves, the functionality here is the same as in EasyTV (and LazyTV as it predecessor). 

If you really want me to remove it I'll have to think about it as I don't think I have any other way of offering the same functionality without it.

basrieter: Would it not be possible to add some _profile_ feature? So it no longer needs cloning?

Rouzax: @basrieter, Could you elaborate on this? 
The power is that the clone of the addon uses it own settings so you can configure it like you want to, and it will work as an independent end point for a menu entry to launch it.

basrieter: It is what I assumed. But you could create "profiles" within your add-on, so you can run one instance of the add-on with multiple setting-profiles.

Rouzax: Yeah, I understand.
It would not offer the same level of flexibility. 

Please let me know the final judgement, and also what that means for EasyTV.

basrieter: How many instances would you need? What is the use case?

Rouzax: For EasTV, I run currently 3 clones. 1 for season premières, 1 for show premières and on to continue watching in progress.

Easymovie riight now with 3 clones, 1 for familie night, 1 for horrors and 1 for action movies.
But have just started using EasyMovie so that might not be the final tally.
Also I don't  know how users use it.

Rouzax:
Hi @basrieter, @romanvm,

Just circling back on this. To address the profiles suggestion and explain why cloning offers something a profile system can't fully replace:

**The core value of clones is independent entry points.** Each clone is its own addon ID, which means:
- Each can be assigned to a different menu item/widget in the skin
- Each has its own settings that users configure once and forget
- Kodi treats them as separate endpoints for `RunScript()`, so skins, favorites, and shortcuts all work natively
- No extra UI needed to pick "which profile" at launch - you just launch the one you want

A profile system would require either a selection step on every launch ("which profile?"), or a way to pass a profile parameter through skin shortcuts - which most skins don't support natively. It adds complexity for the user where clones keep it simple.

**Use cases in practice:**
- EasyMovie: "Family Night" / "Horror Night" / "Action Night" - each with different genre filters, ratings, and viewing modes, each on its own home screen shortcut
- EasyTV: "Season Premieres" / "Show Premieres" / "Continue Watching" - three very different workflows, each a single click from the home screen

The number of clones is up to the user. Some might want one, some might want five. That flexibility is hard to replicate with profiles without making the UX worse.

**Regarding EasyTV:** The clone mechanism in EasyMovie is identical to what EasyTV already uses in the official repo. If the decision is to disallow this pattern, I'd like to understand what that means for EasyTV as well, so I can plan accordingly for both addons.

I'm happy to discuss alternative approaches, but I wanted to lay out why I believe clones are the cleanest solution from a user experience perspective.

romanvm:
Let us discuss this internally and reach some conclusion. My _personal_ opinion is split: On one side, you've presented a legit use case, on the other side, we strongly discourage "drop-in" addon installs as a principle because they are often used to create "fully loaded Kodi boxes" that have given us bad reputation. And we cannot guarantee that your cloning method will work in the future.

Rouzax:
I fully understand.

romanvm:
I discussed this with the team and unfortunately we cannot allow this addnon cloning feature. In addition to the reasons explained above it also breaks addon auto-update mechanism, meaning that cloned addons, having their own addon IDs, wont' be auto-updated along with the primary addon. 

As for this feature being present in your other addon, It seems as an honest mistake from one our our reviewers. We are all unpaid volunteers who review addon submissions in their free time and things may fall through the cracks.

As an alternative you can consider developing some kind of profile support as @basrieter suggested. Maybe existing Kodi profiles feature will do, but I never used it so I cannot tell you more.

Rouzax:
@romanvm Thank you for the clear answer. I understand the team's position and I respect that.

One thing I do want to correct: the claim that cloned addons won't be auto-updated is not accurate. I specifically built an update mechanism to handle this. When the main EasyMovie addon gets updated through the repo, every clone detects the version mismatch on its next launch and triggers a mandatory update from the parent addon's files. The version check logic is in [`_check_clone_version()`](https://github.com/Rouzax/script.easymovie/blob/main/resources/lib/ui/main.py#L68) and the actual update process in [`resources/update_clone.py`](https://github.com/Rouzax/script.easymovie/blob/main/resources/update_clone.py). EasyTV has the same update mechanism in place.

That said, if this is the final verdict, I'll need to think about how to proceed. The clone system is a core feature that I've invested significant time in, building it far more robust than what LazyTV ever had. A profile system, while possible, would not offer the same flexibility - profiles can't each have their own icon, name, or appear as independent entries in Kodi's addon list and skin shortcuts. Most likely I would stop releasing into the official repo and distribute through my own repository instead, so I can keep offering the full feature set to users.

I'd also like to understand what this means for EasyTV, which has had the same clone functionality in the official repo for a while now.

basrieter:
Unfortunately, this is the final verdict. I still think that you should be able to create profiles with their own icons within your add-on and then create links/favorites to those profiles to access the directly. The only think that might not work out of the box is putting it as seperate add-ons in a skin. 

We will discuss the LazyTV add-on internally and see how to proceed with that. 

Please don't misunderstand us: We do appreciate the effort that developers put into their contributions very much. And we hope to see you work out an alternative (such as profiles) that would allow us to accept this add-on into our repo.

Rouzax:
@basrieter Thank you for the clarification. I will wait with my decision until you have discussed internally what this means for EasyTV, as that will factor into how I proceed.

Rouzax:
Updating to v1.0.1 — bug fix release.

### Fixed in 1.0.1

- Resume from the in-progress dialog now correctly starts at the saved position. The previous code called `Player.Open` and tried to seek after a hardcoded 500 ms wait, which was rarely enough time for the player to be ready, so the seek was silently dropped and playback started from zero. Now uses `Player.Open` `options.resume` with a `Player.Position.Time` object to set the start offset before the player initialises. A 10-second rewind is applied for context.

[Release notes](https://github.com/Rouzax/script.easymovie/releases/tag/v1.0.1) - [Full changelog](https://github.com/Rouzax/script.easymovie/blob/main/changelog.txt)

Rouzax:
@romanvm @basrieter

Just to be clear about this update: I fully understand that this PR will most likely not be merged given the team's verdict on the cloning feature. I pushed v1.0.1 here primarily so the latest code (with the resume bug fix) is on record alongside the discussion, not as a renewed merge request. I'll continue distributing through my own repository as I mentioned earlier.

### Re-iterating the EasyTV question

I'm still waiting to hear back on what the team's decision means for [`script.easytv`](https://github.com/xbmc/repo-scripts/tree/omega/script.easytv), which has been in the official repo for a while and uses the exact same clone mechanism. @basrieter mentioned the team would discuss this internally, has there been any conclusion?

Knowing the answer here matters because it affects how I plan future EasyTV releases (continue submitting updates to the official repo, vs. moving EasyTV to my own repository alongside EasyMovie). I'd appreciate a heads-up either way so I can communicate clearly with users.

Thanks for the time you've all put into reviewing this.

romanvm:
First, I appologize that it have taken too long to respond. 
The team's decision is that we cannot allow this addon cloning mechanism. Regarding easytv addon with the same machanism, it was an oversight from our side. It's easy to miss something when reviewing a big PR. So I see 2 options: you can remove this cloning mechanism from both this addon and easytv, our you can continue to distribute them from your own repo. In that case we will have to remove easytv from this repo.

Rouzax:
> First, I appologize that it have taken too long to respond. The team's decision is that we cannot allow this addon cloning mechanism. Regarding easytv addon with the same machanism, it was an oversight from our side. It's easy to miss something when reviewing a big PR. So I see 2 options: you can remove this cloning mechanism from both this addon and easytv, our you can continue to distribute them from your own repo. In that case we will have to remove easytv from this repo.

Whatever I chose, I can keep posting to the forum about it? 

edit: I accidentally pressed re-review. :😅

Rouzax:
@romanvm Thank you for the follow-up, and for discussing this internally.

Cloning is a core feature of EasyMovie that I'm not willing to remove, so I'll continue distributing through my own repository. Feel free to close this PR.

I appreciate the thorough review and the honest discussion around the clone mechanism. Thanks to everyone who engaged with it.

romanvm:
> Whatever I chose, I can keep posting to the forum about it?

Personally, I don't see any problem. On the forum we don't allow promoting piracy and other illegal stuff but technical decisions is another matter even if they don't agree with our strict "better safe than sorry" addon admission policy.

romanvm:
Since the conclusion was reached, I'm closing this PR. I will leave the other PR open for now as a reminder for myself to delete the respective addon. You are free to distribute the addons from your own distribution repo and discuss it on the forum.

PR #2814, opened 2026-03-25
[weather.ozweather@matrix] 2.1.8

bossanova808:
### Add-on details:

- **General**
  - Add-on name: OzWeather
  - Add-on ID: weather.ozweather
  - Version number: 2.1.8

- **Code location**
  - URL: https://github.com/bossanova808/weather.ozweather
  
Weather forecasting and radar images for Australia using Bureau of Meteorology data.  For full features (animated radars & ABC weather videos) - make sure you install the replacement skin files - see information at the addon wiki (https://kodi.wiki/index.php?title=Add-on:Oz_Weather).

### Description of changes:

v2.1.8 
- Add RadarRange label
    	

### Checklist:

### DIFF
--- a/weather.ozweather/addon.xml
+++ b/weather.ozweather/addon.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<addon id="weather.ozweather" name="Oz Weather" version="2.1.7" provider-name="Bossanova808">
+<addon id="weather.ozweather" name="OzWeather" version="2.1.8" provider-name="Bossanova808">
 	<requires>
 		<import addon="xbmc.python" version="3.0.0"/>
     	<import addon="script.module.requests" version="2.22.0+matrix.1"/>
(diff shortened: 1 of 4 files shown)


bossanova808:
@kodiai The line you mention has a method call with parantheses - you are incorrect:

` set_property(WEATHER_WINDOW, 'RadarRange', Store.RADAR_RANGE(radar))`

weather.ozweather/addon.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<addon id="weather.ozweather" name="Oz Weather" version="2.1.7" provider-name="Bossanova808">
+<addon id="weather.ozweather" name="OzWeather" version="2.1.8" provider-name="Bossanova808">
```
basrieter, line 2: Was this space removal intended?

basrieter (commented):
See my comment about the removal of the space-char?

weather.ozweather/addon.xml
bossanova808: It was intended - it's never been called 'Oz Weather' always 'OzWeather'... apparently I made a typo somewhere back before...well, before Github was even a thing!  If it will create problems I can change it back, but it's just niggled at me for like 15 years and suddenly I felt like fixing it.  I figured if the addon ID was the same it would just be a cosmetic thing to correct the name, but maybe that is wrong?

romanvm: Only addon id should matter, so I think it's ok.

romanvm (approved):

romanvm:
@basrieter I don't see any issues so I'm approving and merging this.

PR #2824, opened 2026-05-05
[script.easytv] 1.5.2

Rouzax:
## Update: EasyTV v1.5.2

**Addon ID:** script.easytv
**Version:** 1.5.2
**Previous version:** 1.5.1

### Changes since 1.5.1

This update consolidates six `1.5.2~alpha` development releases plus a small set of stable-polish fixes. The changes break down as follows:

**Multi-instance sync hardening (1.5.2~alpha1 through alpha4):**
- Cross-instance updates (new episodes, fully-watched shows, resume progress) now appear in Browse mode and random playlists without requiring a local library scan
- Clone addons now correctly load `script.module.pymysql` and connect read-only to the shared sync database advertised by the main service. Existing clones will prompt for a mandatory update on next launch.
- Clones can no longer accidentally clear the main service's advertised database configuration when their connection attempt fails
- Fixed duplicate primary-key error during show ID validation at service startup when stale entries shared title + year with valid rows
- "Skip premieres" filter no longer hides in-progress E01 episodes (e.g. resume point synced from another instance), consistent with how smart playlists already handle partially-watched premieres

**Browse-view art reliability (1.5.2~alpha5):**
- Show posters and fanart no longer disappear permanently when the first per-Kodi-session art fetch returns an empty or partial JSON-RPC result. The session art-fetch flag now self-heals on each addon launch and is only set when at least one show actually had a non-empty poster. Reported in https://github.com/Rouzax/script.easytv/issues/2.

**Dialog UX redesign (1.5.2~alpha6):**
- Next-episode prompt rebuilt with two-block copy ("You just watched [show SxxEyy]" / "Up next [SxxEyy title]"), bold-formatted titles, "Watch Next" / "Skip" buttons, and a slightly taller 800x490 dialog so the multi-line message and subtitle fit without crowding. Mirrors the EasyMovie sibling addon's continuation pattern.
- Missed-episode warning rephrased and reformatted: episode identifier now uses Kodi-standard SxxEyy format instead of the previous `(06x02)` bracketed form, and the dialog text gained an "instead?" framing so the swap-to-library-version semantics is explicit.

**Stable polish (1.5.2):**
- "Now Playing" notification no longer truncates show titles that contain commas (e.g. "What We Do, In the Shadows"). Switched from `xbmc.executebuiltin('Notification(...)')` (comma-separated argument parsing) to the typed `xbmcgui.Dialog().notification` API.
- Prompt timeout (seconds) help text corrected: now says "Set to 0 for no automatic dismissal" instead of the inaccurate "Maximum (60) waits indefinitely". Code behaviour was always 0 = indefinite.
- Documentation site rebuilt page by page (excluded from this submission; lives at https://rouzax.github.io/script.easytv/docs/).

### Internal

- Service clears advertised DB config on shutdown, DB failure, and when multi-instance sync is disabled.
- Diagnostic logging added for premiere filter (before/after counts), storage backend selection in clones, and shared DB sync skip reasons.
- New tests exercise the real `import pymysql` code path via `sys.modules` manipulation. Previous clone-fallback tests mocked `SharedDatabase.is_available()` directly, which hid two clone bugs (above) by never executing the failing import or the state-clearing side effect.

### Technical details

- Python 3.8+ compatible (Kodi 21 Omega and newer)
- Depends on xbmc.python 3.0.1 and script.module.pymysql 1.1.0
- Passes pyright, pyflakes, and kodi-addon-checker with no errors
- Sibling addon to [script.easymovie](https://github.com/Rouzax/script.easymovie)

### Links

- **Forum thread:** https://forum.kodi.tv/showthread.php?tid=383902
- **Source code:** https://github.com/Rouzax/script.easytv
- **Full changelog:** https://github.com/Rouzax/script.easytv/blob/main/changelog.txt


romanvm:
As I said in my other comment, we cannot allow addon cloning mechanism in the official repo.

Rouzax:
@romanvm Thank you for the update, and for taking the time to discuss this internally and provide a clear answer.

Cloning is a core feature of EasyTV that I'm not willing to remove, so I'll continue distributing through my own repository. Please go ahead with removing EasyTV from the official repo whenever you're ready.

I'd still like to confirm: am I free to continue using this forum thread to support users and post updates?

Thanks again to everyone who reviewed the submissions and engaged with the discussion.

Rouzax:
@romanvm Thank you for the update, and for providing a clear answer.

Cloning is a core feature of EasyTV that I'm not willing to remove, so I'll continue distributing through my own repository. Please go ahead with removing EasyTV from the official repo whenever you're ready.

I'd still like to confirm: am I free to continue using this forum thread to support users and post updates?

Thanks again to everyone who reviewed the submissions and engaged with the discussion.

romanvm:
Closing ths PR

PR #2841, opened 2026-05-30
[script.easytv] Mark addon as broken

romanvm:
## Summary
Removes all files from the script.easytv addon except addon.xml. The addon violates addon submission rules and is marked as broken to prevent further use while preserving the addon entry in the repository index.

## Changes
- Deleted all addon implementation files (Python scripts, resources, etc.)
- Retained only addon.xml for documentation purposes


garbear:
Here's something to consider. You can still install broken add-ons from the add-on manager. We should neuter addon.xml of all add-on types.

script.easytv/addon.xml
basrieter: We should remove this as @garbear mentioned.

basrieter: We should remove this as @garbear mentioned.

romanvm:
done

script.easytv/addon.xml
```diff
@@ -36,5 +27,6 @@
             <icon>icon.png</icon>
             <fanart>fanart.jpg</fanart>
```
garbear, line 28: I would leave these files. Some way to indicate what this add-on used to look like is better than missing textures IMO.

script.easytv/addon.xml
```diff
@@ -36,5 +27,6 @@
...
         </assets>
+        <lifecyclestate type="broken" lang="en_GB">Addon has been removed as violating addon submission rules.</lifecyclestate>
```
garbear, line 30: Technically if we want this to ever be translated we should keep strings.po, but only the strings present in this file.

garbear: Basically these 3 strings for every language: https://github.com/kodi-game/game.libretro.parallel_n64/blob/master/game.libretro.parallel_n64/resources/language/resource.language.en_gb/strings.po#L19-L29

You could use an agent to remove other strings but these from all the languages.

garbear: Probably not worth it IMO.

garbear (approved):

romanvm:
If no objections I'll merge it by the end of this day.

basrieter (approved):

PR #2858, opened 2026-07-17
[script.audiooffsetmanager@nexus] 2.0.0

matthane:
### Add-on details:

- **General**
  - Add-on name: Audio Offset Manager
  - Add-on ID: script.audiooffsetmanager
  - Version number: 2.0.0

- **Code location**
  - URL: https://github.com/matthane/script.audiooffsetmanager
  
This add-on intelligently adjusts the audio offset based on the detected HDR type, audio format, and FPS type according to user-configured settings, along with other helpful features

### Description of changes:

v2.0.0 (2026-07-16)
- More reliable offset application: stream changes are detected and verified before offsets are applied, eliminating misapplied offsets during playback startup and format switches
- Manual offset adjustments are now detected from any input method (keymaps, remotes, JSON-RPC), not just the Kodi audio settings dialog
- Smarter notifications: they wait for the stream to settle, and no longer get cut short when two notifications arrive close together
- Improved coexistence with PM4K (Plexmod for Kodi): seek-backs are announced via the script.audiooffsetmanager.seeking window property
- New Advanced settings category with a debug logging toggle
- Smaller package: the onboarding test video is now a lightweight dummy clip
        

### Checklist:


matthane:
Some context on this file `resources/media/test-video.mp4` since the bot flagged it (its final decision pass approved).

This addon has always shipped a small test video. It's a functional asset: on a fresh install the addon plays it once to query the device's playback capabilities (HDR support modes, etc.) and trims the visible settings to match, so users only see offset settings their platform can actually use. Kodi only exposes this information during active playback, which is why a bundled clip is needed instead of a static probe.

The currently published version of this addon (1.5.0, in this same directory) already ships this file at about 811 KB. This update replaces it with an 18 KB dummy clip, a 45x reduction. The PR shrinks the addon's only media binary rather than introducing one.

If the team would rather see the file gone entirely, I'm happy to rework the onboarding flow in a follow-up. Since it's precedented here and now trivial in size, I've kept the behavior consistent for upgrading users.

matthane:
Will open new PR, accidentally amended incorrectly.

PR #2864, opened 2026-07-21
[script.audiooffsetmanager.evolved@nexus] 1.1.0

matthane:
**A note for reviewers, since I also maintain `script.audiooffsetmanager`: this is a new, separate addon, not an update or a fork of that one.**

The original Audio Offset Manager is built around a fixed list of audio formats, each configured through its own block of settings in settings.xml. Audio Offset Manager: Evolved is a ground-up redesign around the opposite model: it learns an offset for whatever stream profile Kodi reports (HDR type, audio format, and optional finer splits) from the user's own adjustments during playback, and keeps everything it learns in its own JSON data store rather than in settings. Any format Kodi can report works, including future ones, with nothing to configure up front.

Migrating the original addon to this model was not possible without destroying every existing user's configuration, since the two store their data in incompatible ways. So the original remains in the repo for its existing users, and Evolved is submitted under its own id as the successor. I envision the original's 2.1.0 (https://github.com/xbmc/repo-scripts/pull/2862) as its last update, with development moving to Evolved from here on. The two are developed independently: 2.1.0 backported some stability improvements from this codebase, but there is no ongoing shared code, and they share no settings or data. Evolved detects the original at startup and warns the user not to run both, since they would both react to the same playback.

### Add-on details:

- **General**
  - Add-on name: Audio Offset Manager: Evolved
  - Add-on ID: script.audiooffsetmanager.evolved
  - Version number: 1.1.0

- **Code location**
  - URL: https://github.com/matthane/script.audiooffsetmanager.evolved
  
Adjust Kodi's audio offset during playback the way you always have. Audio Offset Manager: Evolved remembers the value for that stream's HDR type and audio format, with optional finer splits by frame rate, spatial audio format, and channel count, then re-applies it automatically on every matching playback. Includes seek-back after adjustments, playback notifications, and a management view of everything it has learned.

### Description of changes:

v1.1.0 (2026-07-21)
- Distinct spatial audio offsets: optional toggle that keeps spatial formats such as Dolby Atmos and DTS:X as their own profiles with their own offsets. Off by default: a spatial format shares its base codec's offset
- Distinct channel count offsets: optional toggle that also keys offsets by the source's channel count, for example 5.1 and stereo versions of one format. Off by default
- Per-frame-rate offsets is now Distinct frame-rate offsets; the three granularity toggles are grouped together in the settings
- The management view marks the entry that is currently playing
- Stored offsets upgrade to a new file format on first save; older versions read the upgraded file without changing it
- Offsets saved for spatial formats by 1.0.0 are kept but not applied until the spatial toggle is enabled

v1.0.0 (2026-07-19)
First release. Supersedes the original Audio Offset Manager: profiles now cover any format Kodi reports instead of a very limited fixed list, and there is nothing to configure up front.
- Adjust Kodi's audio offset during playback the way you always have; the value is remembered for that stream's HDR type, audio format, and frame rate
- Saved offsets are applied automatically on every matching playback
- Optional seek backs to help realign audio and video after key events
- Notifications when an offset is saved or applied
- Management view to browse, delete, or clear everything learned
- Backup export and import of saved offsets
        

### Checklist:


matthane:
Responding to the automated review's six warnings; all are intentional design, and no arbitrary paths are ever accepted:

1. script_router.py:192 (reads special://logpath): this is the "Export addon log" support feature. It runs only when the user presses the export button in the addon settings and picks a destination folder in a dialog. The export contains only this addon's own log entries, with paths and usernames redacted before writing.

2. script_router.py:249 (os.path.expanduser): the user's home prefix is used only to build redaction pairs for that same export: any occurrence of the real home path in exported lines is replaced with "~/" so usernames do not leave the machine. It is never used for file I/O.

3-5. offset_store.py read_profiles / read_import_document / discard_import: these are pure helpers whose paths are supplied by two fixed callers. read_profiles only ever receives the addon's own data file (special://profile/addon_data/<id>/offsets.json, translated). The import path is a backup file the user explicitly selects in Kodi's file browser; it is copied to a fixed staging file inside the addon profile, the service validates it again, and discard_import removes only that fixed staging file. No user-facing surface accepts a typed or arbitrary path.

6. offset_store.py:536: the store module is deliberately pure (no xbmc imports) with its path injected; both construction sites translate first, e.g. OffsetStore(xbmcvfs.translatePath(STORE_PATH)) in the service runtime, with STORE_PATH under special://profile/addon_data/. All writes stay inside the addon profile.

PR #2865, opened 2026-07-29
[script.service.playbackresumer@matrix] 2.0.10

bossanova808:
### Add-on details:

- **General**
  - Add-on name: Kodi Playback Resumer
  - Add-on ID: script.service.playbackresumer
  - Version number: 2.0.10

- **Code location**
  - URL: https://github.com/bossanova808/script.service.playbackresumer
  

        Runs as a service and will periodically update the resume point while videos are playing, so you can re-start from where you were in the event of a crash.  It can also automatically resume a video if Kodi was shutdown while playing it. See setting to configure how often the resume point is set, and whether to automatically resume.
    

### Description of changes:

v2.0.10
- Fixed the startup resume seek silently failing (video would auto-play from 0:00 despite the "Resuming at X:XX" notification): the addon now waits for playback to actually be ready before seeking, then verifies the seek landed and retries if not
- Fixed a rare crash on startup during a resume, which in turn caused the *next* startup's resume to silently fail too (no video was playing at Kodi's last shutdown)
- Fix tested and reported as working by user here: https://forum.kodi.tv/showthread.php?tid=355383&pid=3299515#pid3299515
    

### Checklist:


bossanova808:
@kodai review

pkscout (approved):

PR #2866, opened 2026-07-29
[script.litebox] v1.0.4

yorah:
### Description

Bugfix update: `script.litebox` is broken on Pillow 10+, which means it is broken on any reasonably current Python environment.

I'm submitting this as a third party. The add-on's declared `<source>` (https://github.com/BADMS/script.litebox) has had no commits since 2021-07-07, and its unreleased 1.0.3 still contains both bugs. Happy to adjust or withdraw if you'd rather this went through the original authors.

**This affects users on current Kodi.** The build in this branch is what the official repo serves to Omega users today. I downloaded `https://mirrors.kodi.tv/addons/omega/script.litebox/script.litebox-1.0.0.zip` and compared it to `matrix`:

```
d6f9fa8df07a5402e691b7cd2bf897aa  resources/lib/utils.py
b6050a4b11a13c0458f71766fe201867  resources/lib/imageoperations.py
22f801c41000753c262ef197ec925ff6  addon.xml
```

Byte-identical, hence submitting here rather than to a newer branch.

#### Bug 1: `Image.ANTIALIAS` (removed in Pillow 10.0)

Deprecated in Pillow 9.1, removed in Pillow 10.0 (July 2023). Two `resize()` calls in `resources/lib/utils.py` still used it, so every image operation raised:

```
module 'PIL.Image' has no attribute 'ANTIALIAS'
```

Replaced with `Image.LANCZOS`. `ANTIALIAS` was always an alias for it, and `LANCZOS` has existed since Pillow 2.7, so this stays compatible with old Pillow rather than requiring `Image.Resampling` (9.1+).

#### Bug 2: `image.gaussian_blur()` core signature

`MyGaussianBlur` called the C core directly as `image.gaussian_blur(self.radius)`. Newer Pillow requires a 2-item `(xradius, yradius)` sequence, raising:

```
argument 1 must be 2-item sequence, not int
```

Rather than hardcode either signature, `MyGaussianBlur` now inherits `ImageFilter.GaussianBlur`, so Pillow itself makes the core call and adapts to whatever version is installed. The `radius` default of 10 is preserved. One behavioural note: `GaussianBlur` is a `MultibandFilter`, so the blur is applied to all bands at once instead of per-band. The result is identical for a Gaussian blur.

#### Verification

Exercised both patched paths (`MyGaussianBlur(radius=30)` and `resize(..., LANCZOS)`) against real Pillow releases:

| Pillow | original | patched |
|---|---|---|
| 8.4.0 (py3.9) | OK | **OK** |
| 9.5.0 (py3.11) | OK | **OK** |
| 10.0.0 | blur fails | **OK** |
| 10.4.0 | blur fails | **OK** |
| 11.3.0 | `argument 1 must be 2-item sequence, not int` | **OK** |

So the fix repairs Pillow 10+ without regressing the Pillow 8/9 versions current when this branch was cut.

Also confirmed on a live install (CoreELEC 21.3, Kodi 21, Python 3.13, Pillow 11.0.0), where the unpatched add-on logged 1985 `ANTIALIAS` failures in a single session and never produced an output image. Note that `script.module.pil` there is a stub containing only `addon.xml` and `icon.png`, so `from PIL import Image` resolves to the OS Pillow. The version is set by the distro, entirely outside the add-on system.

`kodi-addon-checker --branch=matrix --PR script.litebox` reports **0 problems**. The 3 warnings are pre-existing (`<forum>` URL redirect/403, `default.py` entry-point length) and untouched by this change.

### Checklist:

### DIFF
--- a/script.litebox/README.md
+++ b/script.litebox/README.md
@@ -1,1 +1,1 @@
-see skin.adonic or AeonMQ6 for current usage!
+see skin.ambex or AeonMQ6 for current usage!
--- a/script.litebox/addon.xml
+++ b/script.litebox/addon.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<addon id="script.litebox" name="litebox Script" version="1.0.0" provider-name="BADMS, Angelinas, drinfernoo">
+<addon id="script.litebox" name="litebox Script" version="1.0.4" provider-name="yorah, BADMS, Angelinas, drinfernoo">
 	<requires>
 		<import addon="xbmc.python" version="3.0.0"/>
 		<import addon="script.module.simplejson" version="2.0.10"/>
@@ -13,9 +13,9 @@
 		<description lang="en_GB">for skinners</description>
 		<platform>all</platform>
 		<license>GNU GENERAL PUBLIC LICENSE Version 2, June 1991</license>
-		<forum>http://forum.kodi.tv/</forum>
-		<source>https://github.com/BADMS/script.litebox</source>
-		<email>philipkdick@gmail.com</email>
+		<forum>https://forum.kodi.tv/showthread.php?tid=363433</forum>
+		<source>https://github.com/yorah/script.litebox</source>
+		<email>yoram.harmelin@gmail.com</email>
 		<assets>
 			<icon>resources/icon.png</icon>
 			<fanart>resources/fanart.jpg</fanart>
--- a/script.litebox/changelog.txt
+++ b/script.litebox/changelog.txt
@@ -1,2 +1,28 @@
+v1.0.4
+- Fix Pillow 10+ compatibility: Image.ANTIALIAS was removed in Pillow 10.0
+- Fix MyGaussianBlur: the PIL core gaussian_blur signature changed, inherit
+  ImageFilter.GaussianBlur so Pillow supplies the correct call
+- Fix hls_to_rgb: ONE_THIRD, ONE_SIXTH and TWO_THIRD were rounded to one
+  decimal, which skewed every hue and pushed channels past 255
+- Clamp in RGB_to_hex so an out-of-range channel can no longer produce a
+  nine character colour string
+- fhls and fhsv: restore the '-' form, and force the hue channel rather
+  than flooring it, since hue is circular and max() is meaningless on it
+- Drop a leftover debug window property from rgb_to_hsv
+- Drop the commented-out RNG daemon scaffolding from default.py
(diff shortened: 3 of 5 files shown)


pkscout (approved):

pkscout:
I approved the review and am checking with the team to see about merging a third-party update.

pkscout:
@BADMS are you OK with this third party merge, or would you rather have them submit the update to you directly and then do an addon update?

pkscout:
@yorah We can consider this addon abandoned at this point.  If you want to take over maintenance, we ask that you fork the original repo BADMS had and then update the addon.xml to add you as a provider (you can put your info first and leave the others there) as well as the source and contact email.

yorah:
@pkscout Done. Forked to https://github.com/yorah/script.litebox and updated addon.xml: I'm first in provider-name with the originals kept, and source and email point at me.

Version is 1.0.4 rather than 1.0.1. This repo has been on 1.0.0, but BADMS put out 1.0.1, 1.0.2 and 1.0.3 in their own repo and never submitted them here. Reusing 1.0.1 would mean two different code states under the same number, so I pulled their changes in and bumped past them. 1.0.2 and 1.0.3 only move the daemon poll interval from 0.2 to 0.1 and back, so nothing from them shows up in the diff.

The diff is bigger than before because the colour helpers in 1.0.0 are broken:

* `fhsv` always raises TypeError. A paren is misplaced, so `check_mod` gets one argument instead of two.
* `hsv` always returns hue 0. The bundled `rgb_to_hsv` is colorsys with `/` replaced by `//`.
* `ONE_THIRD` and friends are `round(1/3, 1)`, so `hls_to_rgb` runs with 0.3, 0.2 and 0.7. Hues come out skewed and channels can go past 255. `RGB_to_hex` doesn't clamp, so those turn into nine character strings like `FF124ff00`.

BADMS fixed the first two in 1.0.1 by moving to colorsys. The constants and the clamp are mine. Over 5616 test cases 1.0.0 crashes 1296 times and emits 270 bad colour strings. This branch does neither.

Worth knowing for skinners: `hls`, `hsv` and `fhls` produce different colours than 1.0.0 now. `bump` is unchanged.

One deliberate difference from BADMS. Their 1.0.1 made `fhls` and `fhsv` numbers mean `max(requested, current)` instead of "set to this". That works for lightness and saturation, but hue is a wheel, so max() just drops the request about half the time. I kept the floor for lightness and saturation and force hue. I also restored the `-` form, which 1.0.1 broke.

romanvm:
@pkscout is it Ok to merge?

PR #2870, opened 2026-08-06
[script.module.bossanova808@matrix] 1.0.3

bossanova808:
### Add-on details:

- **General**
  - Add-on name: Bossanova808 Dependencies
  - Add-on ID: script.module.bossanova808
  - Version number: 1.0.3

- **Code location**
  - URL: https://github.com/bossanova808/script.module.bossanova808
  
Common code needed by Bossanova808 addons

### Description of changes:

v1.0.3
- Defensive coding pass: guard against malformed/missing settings and RPC responses instead of crashing
- Log OS, module version, and Kodi's actual debug logging status at addon startup
- Redact sensitive-looking local variables when logging an unhandled exception
- Always prefix log lines with addon name/version, even for non-string messages
- This update also adds a new shared Playback class (`resources/lib/bossanova808/playback.py`). 

This 'new' (really it's old & tested) Playback class centralises the logic for identifying what's currently playing in Kodi - library/PVR/addon/file source, media type, artwork, episode/season info, etc - so multiple addons can share one tested implementation instead of each maintaining their own copy.  It's being extracted out of `plugin.switchback` (which will be updated to use it in a follow-up submission), and `script.service.playbackresumer` will adopt it as well. No behavioural change to this module's existing consumers from this addition.

### Checklist:

### DIFF
--- a/script.module.bossanova808/addon.xml
+++ b/script.module.bossanova808/addon.xml
@@ -1,7 +1,8 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<addon id="script.module.bossanova808" name="Bossanova808 Dependencies" version="1.0.2" provider-name="bossanova808">
+<addon id="script.module.bossanova808" name="Bossanova808 Dependencies" version="1.0.3" provider-name="bossanova808">
     <requires>
         <import addon="xbmc.python" version="3.0.0"/>
+        <import addon="script.module.infotagger" version="0.0.7"/>
     </requires>
     <extension point="xbmc.python.module" library="resources/lib" />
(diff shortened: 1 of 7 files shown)


bossanova808:
This update also adds a new shared `Playback` class (`resources/lib/bossanova808/playback.py`). It centralises the logic for identifying what's currently playing in Kodi - library/PVR/addon/file source, media type, artwork, episode/season info, etc - so multiple addons can share one tested implementation instead of each maintaining their own copy.

It's being extracted out of `plugin.switchback` (which will be updated to use it in a follow-up submission), and `script.service.playbackresumer` will adopt it as well. No behavioural change to this module's existing consumers from this addition.

PR #2873, opened 2026-08-06
[script.audiooffsetmanager.evolved@nexus] 1.3.0

matthane:
### Add-on details:

- **General**
  - Add-on name: Audio Offset Manager: Evolved
  - Add-on ID: script.audiooffsetmanager.evolved
  - Version number: 1.3.0

- **Code location**
  - URL: https://github.com/matthane/script.audiooffsetmanager.evolved
  
Adjust Kodi's audio offset during playback the way you always have. Audio Offset Manager: Evolved remembers the value for that stream's HDR type and audio format, with optional finer splits by frame rate, spatial audio format, channel count, and audio output device, then re-applies it automatically on every matching playback. Includes seek-back after adjustments, playback notifications, and a management view of everything it has learned.

### Description of changes:

v1.3.0 (2026-08-06)
- Stored offsets are backed up automatically and restored if the file is ever damaged
- The offset saved notification shows the value it replaced
- Stored offsets can be copied to the current audio output from Manage stored offsets
- Settings help text now says where an option depends on what your Kodi build reports
- Fixed: an adjustment made just as the stream changed could be saved to the wrong profile
- Fixed: a codec switch could briefly follow the wrong output device when PCM and passthrough use different ones
- Fixed: re-teaching a value deleted moments earlier did not trigger the seek-back replay

v1.2.0 (2026-08-03)
- Distinct output device offsets: optional toggle that keys offsets by Kodi's configured audio output device, so a soundbar and a pair of headphones can each keep their own value. Off by default
- An output device change during playback is picked up within a second and the matching offset is applied
- The management view groups by output device, then HDR type, and skips a level when it holds only one entry
- Device offsets belong to the machine that learned them and do not transfer to another one; every other offset still does
- Stored offsets upgrade to a new file format on first save. Version 1.1.0 and earlier can still read the file but will not add to it
        

### Checklist:


matthane:
Hi Roman + team, both warnings point at the "Export addon log" feature, which these same checks flagged on the 1.1.0 submission (#2864) and which was explained there. The feature is unchanged since then: a documentation pass rewrote its docstrings, which is what the review picked up here, and the export code itself is identical.

PR #2890, opened 2026-09-01
[script.skinshortcuts] v3.0.3

MikeSiLVO:
### Description
Skin Shortcuts 3.0.3.

3.0.3
- Fix: a widget picked from a library node takes its type from the node, so add-on nodes are no longer mistyped

3.0.2
- Add: skins can restyle the DialogSelect pickers with a skinshortcuts-picker property and ListItem metadata
- Add: skins can define the management dialog context menu, the buttons it lists and where it opens
- Add: skins can retire a widget, background or property in an overrides block and point saved menus at its replacement
- Add: icon overrides name a source folder, so a skin's own icons replace the Kodi defaults, packed or loose
- Add: a rename attribute on a widget picker button prompts for a custom widget label
- Add: a global submenuPath element, plus a submenuPath property on menu items for their widget submenu
- Add: show a source as a library view instead of files, with a sort order
- Fix: program add-ons that expose plugin content are browsable in the content picker
- Fix: property conditions combine + and | as Kodi does, and an expression reference expands as a single term
- Fix: the includes file is written with unix line endings on every platform
- Fix: the skin reloads only when a rebuild changes the includes
- Fix: the picker keeps its place when it reopens instead of jumping back to the current selection
- Fix: backgrounds that open a browse or playlist dialog get the same folder arrow as groups

### Checklist:


MikeSiLVO:
@greptile-apps ignore the language folders and strings.po files. That will bring it under the 100 file limit.

PR #2892, opened 2026-09-12
[script.livetennisscores] 1.0.0

bensynapse:
**Disclosure first:** this add-on reads the Live Tennis API, which I operate. I am submitting it as both the add-on author and the data provider, so please weigh it accordingly. The add-on ships no key, hardcodes no credentials, and does nothing at all until the user pastes their own key into its settings. The provider's free plan needs no card and is enough for everything the add-on does; paid plans are not required and are not advertised in the UI. AI tooling was used while writing it; the code and the checks below are mine.

## Description

A new service add-on: live tennis score notifications plus a scoreboard of the matches in play. It notifies when a match goes on court, when a set is won, when a match finishes, and, only if switched on, on break points. Opening it shows the matches in play with set scores, the current game's points, a marker against the player serving and `BP` on a break point. ATP, WTA, Challenger and ITF.

There is no tennis add-on anywhere in this repository today, and the checker agrees, reporting "This is a new addon". The closest precedent, and the model I followed, is `script.nhlscores` on the matrix branch, a single-sport live-score service add-on declaring the same service extension point. Submitted to nexus, which per `targets.cfg` also reaches the omega and piers repos.

## Notes for review

- **No new dependencies.** `xbmc.python` 3.0.1 only. HTTP is stdlib `urllib.request`, which also keeps the `--PR` dependency check happy.
- **Quota-honest by design.** The provider's free plan allows 100 requests a day, so the add-on polls every 20 minutes by default and refuses to poll faster than every 15, which is 96 a day. That floor is clamped in code, not only in the settings slider. Opening the scoreboard normally costs no request, because it reuses the service's last poll and tells you how old the scores are when it does. An unparseable setting falls back to the 20-minute default rather than to the floor, so a corrupt config cannot become the most aggressive option.
- **Honest about what it cannot do.** The live endpoint only returns matches in progress, so a finish is inferred from a match being absent from three consecutive checks, and no winner is claimed unless the feed named one. A feed outage longer than that can report a match as finished while it is still being played. Break points last a few points, so a 15-minute cadence catches only some of them. The settings help says so, and so does the changelog.
- **Abort handling.** The wait between polls is `Monitor.waitForAbort` in short slices with its result always propagated, never `time.sleep`, and never with a timeout of zero or less, which Kodi reads as "wait forever". The HTTP timeout is 4 seconds so a stalled request cannot outlive Kodi's 5-second shutdown grace.
- Logs at debug level only, per the add-on rules, with no `print()`. The API key is sent as a header, never in a URL, and never logged. With no key the service idles silently and the window explains what to do.
- **Icon and fanart are my own programmatically generated artwork**, meeting the checker's dimension rules. Happy to replace them if you would prefer something else. No screenshots are included, rather than fabricated ones, since I cannot run Kodi to capture any.

## Verification

```
kodi-addon-checker --branch=nexus --PR script.livetennisscores
  We found no problems and no warnings, please enjoy your day.
```

Also run: `py_compile` over all twelve modules, XML parse plus schema validation against the four XSDs the checker selects for nexus, a logical-line count on both entry points well inside the limit, a cross-check that every string id resolves and every settings key read is declared, and a 172-check logic harness with Kodi's modules stubbed. That harness covers break-point derivation including the tiebreak guard, player-major set-score formatting, the null-points and empty-games shapes a completed match returns, the cadence arithmetic, and the abort path returning promptly rather than sleeping out the interval.

## What I could not verify

Kodi is not installed on the machine I built this on, so the add-on has never run. Nothing is confirmed about the settings dialog, the notification appearance, the text-viewer layout or service startup. Every Kodi API claim comes from the nexus branch source and the wiki, and every Kodi call is exercised only against stubs.

I also made no authenticated API call, so field shapes come from our published OpenAPI document rather than an observed payload. Two consequences are unmeasured: whether the combined live list ever exceeds a single page at peak, though pagination is now followed rather than truncating, and how often a feed gap produces a false finish.

## Checklist

- [x] Each add-on submission should be a single commit using the required style


bensynapse:
All four findings were real. Fixed and force-pushed; the checker still reports no problems and no warnings.

**1. Pagination broke the quota budget.** This one invalidated a claim in my own description, so thank you for catching it. The 15-minute floor bounds *polls*, not requests: a day holds 96 polls, and a paginated poll costs one request per page up to a cap of three, so the cadence alone permitted between 96 and 288 requests a day against a free allowance of 100.

Requests are now bounded separately by a persisted daily budget. Every outbound request consumes one of 100; the count lives in the add-on's own profile directory so it survives a Kodi restart and resets at midnight UTC; a poll that cannot afford its next page stops there and reports the list as incomplete; a poll that cannot afford its first page is not made at all, so no socket is opened. Three requests, one whole scoreboard read, are held back from the polling service so opening the scoreboard by hand still works on the last poll of a day. A simulated worst-case day of 96 polls each wanting three pages sends 97 requests from the service plus 3 for a manual open: exactly 100, never more. A setting can raise the limit for a paid plan but is clamped so it can only ever raise it.

**2. Partial lists looked complete.** The client now reports whether the list it returned is complete, and it is judged strictly against the provider's own schema: the body must be an object, the data an array, the metadata an object, and the has-more flag an actual boolean. It is marked incomplete when a page after the first failed, the page cap was hit with more to come, the budget ran out mid-pagination, or a page is off-contract. On an incomplete list the absence counters are carried forward untouched and no finish is announced, so an incomplete list can never retire a match. Matches actually present in a partial list are still diffed, since that is observation rather than inference. Two related bugs fell out of this: the first-poll flag advanced on an incomplete list, which would have turned matches it never read into a burst of false "now on court" notifications, and an empty incomplete list claimed nothing was in play.

**3. A new key did not wake the service.** It now compares what the last poll attempt saw and returns from the wait early on absent-to-present, once, after which the floor applies again from that poll. The woken poll still pays for its requests out of the same budget, so this cannot be used to bypass either the floor or the allowance.

**4. Cache writes were not atomic.** Writes now go to a temporary file in the same directory, are flushed and fsynced, then replaced into place, with the descriptor closed and the temporary file removed on any failure. Abandoned temporary files older than an hour are swept. The reader treats missing, corrupt, truncated or wrong-shaped data as absent rather than raising.

Verification: the add-on checker is clean; every module compiles; the localisation and settings ids cross-check; and the logic harness is now 350 checks, up from 172, covering the budget refusing a request without opening a socket, the budget surviving a simulated restart with the in-memory value explicitly cleared so the file is what is proved, a UTC day rollover and a backwards clock jump, an incomplete list failing to retire a match across ten consecutive polls, the key wake firing exactly once, and a reader at the instant of the atomic replace seeing the whole old file rather than a partial one.

Three things I am leaving documented rather than engineered around, and would change if you disagree. Cross-process budget races between the service and a hand-opened scoreboard are unlocked: they are bounded by the page cap, only reachable at a day's ceiling, and the provider's rate-limit response is the backstop; a lock file would need breaking safely after a crash. A clock correction across a date boundary on hardware without a real-time clock grants a fresh allowance, because comparing dates for ordering instead would lock the add-on out after a backwards jump. And the per-minute limit has no enforcement in code, since caching the scoreboard's own fetch removes the realistic path to reaching it.

Still unverified on my side, as before: Kodi is not installed on my build machine, so the add-on has never run and every Kodi call is exercised against stubs. Only Python 3.14 is available here, so I could not execute against Nexus's 3.8, though a static scan found nothing newer than 3.5 in use.

